PR #8, opened 2026-05-13
fix(cli): self-heal plugin enabled after OpenClaw upgrade + complete groupId/content descriptions

### LINKED ISSUE #5: CLI: self-heal plugins.entries.enabled after OpenClaw major upgrade
### Bug Description

After upgrading OpenClaw from 4.x to 5.x, third-party plugins' `plugins.entries.<id>.enabled` flag is reset to `false`. The DMWork plugin remains installed but inactive — bots silently stop replying until the user manually runs `openclaw plugins enable openclaw-channel-dmwork`.

### Steps to Reproduce

1. Run OpenClaw 4.x with `openclaw-channel-dmwork@0.6.x` installed; verify the bot replies normally.
2. Upgrade OpenClaw to 5.x (`npm install -g openclaw@latest`).
3. Observe in `~/.openclaw/openclaw.json` that `plugins.entries.openclaw-channel-dmwork.enabled` is now `false`.
4. Bot stops responding to messages.

### Expected Behavior

Re-running `npx openclaw-channel-dmwork install` should self-heal the config back to `enabled: true`, so the user does not have to manually invoke `openclaw plugins enable openclaw-channel-dmwork`.

### Proposed Fix

1. Add an `ensurePluginEnabled()` helper in `cli/openclaw-cli.ts` (matching the existing `ensurePluginsAllow()` best-effort/idempotent pattern).
2. Invoke it at the end of `runInstall()` in `cli/install.ts`, including the already-at-target path. The two early `return`s in the update scenario must be converted to `break` so the self-heal runs even when no install/restart is needed.
3. In `cli/doctor.ts`, downgrade the "Plugin enabled" check from `FAIL` to `WARN` (without `--fix`) and include an actionable command hint pointing at either `doctor --fix` or `openclaw plugins enable openclaw-channel-dmwork`.

### Origin

Already implemented and merged upstream in dmwork-org/dmwork-adapters#240. This issue tracks mirroring the fix to `Mininglamp-OSS/octo-adapters`.

### LINKED ISSUE #6: agent-tools: list all 16 actions requiring groupId in dmwork_management schema
### Bug Description

The `dmwork_management` tool's `groupId` parameter description in `src/agent-tools.ts` only lists 4 group-level actions, but the runtime requires `groupId` for 16 actions in total (7 group-level + 9 thread-level). LLM agents read the schema, omit `groupId` on the first call to thread/group-management actions, receive a runtime error, and only succeed on retry — wasting one tool call per operation.

The same issue affects the `content` parameter, which omits `thread-md-update`.

### Steps to Reproduce

1. Configure a DMWork bot.
2. Have an LLM agent invoke `dmwork_management` with `action: "delete-thread"` and `shortId` only (no `groupId`).
3. The first tool call returns `{"error": "groupId is required for delete-thread"}`.
4. The agent retries with `groupId` included; second call succeeds.

### Expected Behavior

LLM agents should pick the correct argument set from the schema description on the first call.

### Proposed Fix

Update `groupId` description to enumerate all 16 actions (group-level + thread) plus an explicit "Not required for" negative list, and add `thread-md-update` to the `content` description.

### Origin

Original report: dmwork-org/dmwork-adapters#237. Implemented and merged upstream in dmwork-org/dmwork-adapters#240. This issue tracks mirroring the fix to `Mininglamp-OSS/octo-adapters`.

caster-Q:
## Summary

Mirror two fixes from upstream `dmwork-org/dmwork-adapters` to `octo-adapters`:

1. CLI self-heal of `plugins.entries.<id>.enabled` after OpenClaw major upgrades.
2. Complete enumeration of all 16 actions requiring `groupId` in the `dmwork_management` schema.

## Related Issue

Closes #5
Closes #6

## Changes

- `cli/openclaw-cli.ts`: add `ensurePluginEnabled()` helper (idempotent / best-effort, mirrors the existing `ensurePluginsAllow()` pattern).
- `cli/install.ts`: invoke `ensurePluginsAllow()` + `ensurePluginEnabled()` at the end of `runInstall()`, including the already-at-target path. Convert the two early `return`s in the update scenario to `break` so the self-heal step always runs even when no install/restart is needed.
- `cli/doctor.ts`: downgrade the "Plugin enabled" check from `FAIL` to `WARN` (without `--fix`) and include an actionable command hint pointing at either `doctor --fix` or `openclaw plugins enable openclaw-channel-dmwork`. The `--fix` path is unchanged.
- `cli/install.test.ts`: add a regression test covering the exact reported scenario — plugin already at target version, `entries.enabled=false` in config; install must write `enabled=true` without re-installing or restarting the gateway.
- `src/agent-tools.ts`: enumerate all 16 actions (7 group-level + 9 thread) in the `groupId` description plus an explicit "Not required for" list; add `thread-md-update` to the `content` description.

## Testing

Local verification:

- `npx tsc --noEmit` — passes (0 errors)
- `npm test` — 695/695 tests pass across 22 files, including the new `install.test.ts` regression
- `npm run build` — passes

Note: octo-adapters does not yet have a CI workflow for type-check/test/build, so the above checks were run locally before pushing.

- [x] Manually verified

## Checklist

- [x] Added tests for my changes (1 new regression test in `cli/install.test.ts`)
- [ ] Updated documentation — N/A (changes are schema description text + an internal CLI helper; no user-facing docs require update)

### DIFF
--- a/openclaw-channel-dmwork/cli/doctor.ts
+++ b/openclaw-channel-dmwork/cli/doctor.ts
@@ -223,6 +223,10 @@ export async function runDoctorChecks(params: {
   }
 
   // 2. Plugin enabled
+  // After OpenClaw major upgrades, plugins.entries.<id>.enabled may be reset on
+  // third-party plugins. Doctor surfaces this as a WARN (not FAIL) with a one-liner
+  // fix command, so users don't have to dig through config to figure out why the bot
+  // went silent.
   if (!params.inProcess) {
     const enabled = reader.get(
       "plugins.entries.openclaw-channel-dmwork.enabled",
@@ -234,10 +238,21 @@ export async function runDoctorChecks(params: {
         configSet("plugins.entries.openclaw-channel-dmwork.enabled", "true");
         checks.push({ name: "Plugin enabled", status: "FIXED", detail: "Enabled" });
       } catch {
-        checks.push({ name: "Plugin enabled", status: "FAIL", detail: "No (auto-fix failed)" });
+        checks.push({
+          name: "Plugin enabled",
+          status: "FAIL",
+          detail:
+            "No (auto-fix failed; run `openclaw plugins enable openclaw-channel-dmwork`)",
+        });
       }
     } else {
-      checks.push({ name: "Plugin enabled", status: "FAIL", detail: "No" });
+      checks.push({
+        name: "Plugin enabled",
+        status: "WARN",
+        detail:
+          "No — installed but disabled. Run `openclaw-channel-dmwork doctor --fix` " +
+          "or `openclaw plugins enable openclaw-channel-dmwork`.",
+      });
     }
   }
 
--- a/openclaw-channel-dmwork/cli/install.test.ts
+++ b/openclaw-channel-dmwork/cli/install.test.ts
@@ -175,4 +175,51 @@ describe("runInstall — update scenario", () => {
     expect(pluginsInstallSpec(calls)).toBe("openclaw-channel-dmwork");
     expect(didCallGatewayRestart(calls)).toBe(true);
   });
+
+  it("already target version + entries.enabled=false: self-heals enabled, no install, no restart", async () => {
+    // Regression: install used to early-return on already-at-target, bypassing
+    // the self-heal that re-enables the plugin after OpenClaw major upgrades
+    // reset entries.<id>.enabled.
+    const fs = await import("node:fs");
+    const mockReadFileSync = vi.mocked(fs.readFileSync);
+    const mockWriteFileSync = vi.mocked(fs.writeFileSync);
+
+    mockReadFileSync.mockImplementation((path) => {
+      if (String(path).endsWith("openclaw.json")) {
+        return JSON.stringify({
+          plugins: {
+            entries: { "openclaw-channel-dmwork": { enabled: false } },
+            installs: {
+              "openclaw-channel-dmwork": { source: "npm", version: "0.6.0" },
+            },
+          },
+        });
+      }
+      return "{}";
+    });
+
+    const { runInstall } = await loadInstall();
+
+    mockExecFileSync.mockImplementation((_cmd, args) => {
+      const a = args as string[];
+      if (a[0] === "config" && a[1] === "file") return "/home/user/.openclaw/openclaw.json";
+      if (a[0] === "--version") return "OpenClaw 2026.4.15\n";
+      if (a[0] === "plugins" && a[1] === "inspect") {
+        return JSON.stringify({ plugin: { id: "openclaw-channel-dmwork", version: "0.6.0", enabled: false } });
+      }
+      if (a[0] === "view") return "0.6.0\n";
+      return "";
+    });
+
+    await runInstall({ force: false, dev: false });
+
+    const calls = getCalledArgs();
+    expect(didCallPluginsInstall(calls)).toBe(false);
+    expect(didCallGatewayRestart(calls)).toBe(false);
+
+    // Self-heal must have written a config with enabled: true
+    const writes = mockWriteFileSync.mock.calls.map((c) => String(c[1]));
+    const enabledWrite = writes.find((w) => w.includes('"enabled": true'));
+    expect(enabledWrite).toBeDefined();
+  });
 });
--- a/openclaw-channel-dmwork/cli/install.ts
+++ b/openclaw-channel-dmwork/cli/install.ts
@@ -16,6 +16,7 @@ import {
   cleanupBrokenInstall,
   deleteLegacyBackup,
   detectScenario,
+  ensurePluginEnabled,
   ensurePluginsAllow,
   gatewayRestart,
   getConfigFilePathSafe,
@@ -87,15 +88,15 @@ export async function runInstall(opts: InstallOptions): Promise<void> {
       const targetVersion = getLatestNpmVersion(tag);
 
       if (!targetVersion) {
-        // Cannot determine target — skip install and restart
+        // Cannot determine target — skip install and restart, but still self-heal config below
         console.log(`Cannot reach npm registry to check ${tag} version.`);
         console.log(`Current version: v${currentVersion}`);
-        return;
+        break;
       }
 
       if (currentVersion === targetVersion) {
         console.log(`DMWork plugin v${currentVersion} is already the target version${opts.dev ? " (dev)" : ""}. No update needed.`);
-        return; // Nothing changed — skip gateway restart
+        break; // No install/restart needed, but self-heal still runs (didChange stays false)
       }
 
       console.log(`Updating DMWork plugin: v${currentVersion} → v${targetVersion}${opts.dev ? " (dev)" : ""}...`);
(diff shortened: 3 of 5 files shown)


Jerry-Xin (approved):
Summary: The PR is focused, aligns with existing CLI config-repair patterns, and I found no blocking bug, security, performance, or architecture issues.

🔴 Blocking

None.

💬 Non-blocking

🔵 Suggestion: Consider adding direct unit coverage for `ensurePluginEnabled()` in `openclaw-channel-dmwork/cli/openclaw-cli.ts:944`, especially idempotent true/no-write behavior and malformed config handling. The install regression test covers the main user-facing path, so this is optional.

✅ Highlights

- `openclaw-channel-dmwork/cli/install.ts:90` and `openclaw-channel-dmwork/cli/install.ts:97` correctly replace early returns with `break`, allowing config self-heal to run without changing install/restart behavior.
- `openclaw-channel-dmwork/cli/openclaw-cli.ts:944` keeps the new helper idempotent and best-effort, matching nearby config helper style.
- `openclaw-channel-dmwork/cli/doctor.ts:248` appropriately downgrades the disabled-plugin state to `WARN` while preserving `--fix` behavior.
- `openclaw-channel-dmwork/src/agent-tools.ts:125` now matches the actual action handler requirements for group and thread actions, including `thread-md-update` content at `openclaw-channel-dmwork/src/agent-tools.ts:135`.

Verification note: I attempted to run the focused tests from `openclaw-channel-dmwork`, but this checkout does not have dev dependencies installed, so `vitest` was unavailable.

lml2468 (approved):
Clean, well-motivated fix. Independent review — no blocking issues found.

**Summary**: Two `return` → `break` changes in `install.ts` fix a regression where already-at-target installs would skip config self-healing. The new `ensurePluginEnabled()` helper is idempotent, atomically writes, and correctly fails open (returns false on I/O error). doctor.ts severity downgrade from FAIL to WARN with actionable fix commands is appropriate. agent-tools.ts description improvements are accurate.

🔴 Blocking

None.

💬 Non-blocking

🔵 **Minor redundancy**: `ensurePluginsAllow()` is now called both inside `runLegacyMigration`/`runDeadlockRepair` AND after the switch in `runInstall` (line 135). Harmless since it is idempotent, but could consolidate to the single post-switch call and remove the duplicates in the sub-functions for clarity.

🔵 **Agree with Jerry-Xin**: Direct unit tests for `ensurePluginEnabled()` (idempotent no-write, null config, missing intermediate keys) would strengthen coverage, but the regression test at line 179 covers the critical user-facing path.

✅ Highlights

- `return` → `break` is the minimal correct fix — `didChange` stays false so no unnecessary gateway restart.
- `writeConfigAtomic` (tmp + rename) prevents half-written config reads. Good pattern already established, correctly reused here.
- Regression test precisely targets the bug scenario (already-at-target + enabled=false) and verifies both the negative (no install/restart) and positive (config written with enabled: true).

yujiawei (approved):
# Code Review — PR #8 (octo-adapters)

## Summary

Two independent fixes mirrored from `dmwork-adapters`:

1. CLI self-heal of `plugins.entries.openclaw-channel-dmwork.enabled` after OpenClaw major upgrades, plus a doctor UX downgrade `FAIL → WARN` with an actionable hint.
2. Completion of the `groupId` / `content` parameter descriptions in the `dmwork_management` agent tool so LLMs supply the right arg set on the first call.

Both fixes are small, focused, idempotent, and well-scoped. The `return → break` conversion in `runInstall` is the load-bearing change and is correct. Approving with a few non-blocking notes.

## Verification

| Item | Result | Evidence |
| --- | --- | --- |
| Self-heal runs on already-at-target path | ✅ | `cli/install.ts:99` (`break` not `return`), then unconditional `ensurePluginsAllow(); ensurePluginEnabled()` at `cli/install.ts:135-136`, before the `if (!didChange) return;` gate at `cli/install.ts:138` |
| Self-heal runs when npm registry unreachable | ✅ | `cli/install.ts:94` (`break` not `return`); same fall-through |
| Gateway is NOT restarted when nothing actually changed | ✅ | `didChange` stays `false` on the early-`break` paths; restart is gated by `if (!didChange) return;` (`cli/install.ts:138`) |
| `ensurePluginEnabled` is idempotent and best-effort | ✅ | `cli/openclaw-cli.ts:944-958` — early-returns when already enabled, all errors swallowed, returns `boolean` |
| Atomic write semantics | ✅ | `writeConfigAtomic` uses tmp + rename (`cli/openclaw-cli.ts:872-877`) |
| `groupId` description enumerates exactly the actions that branch on it | ✅ | Cross-checked against `src/agent-tools.ts:259-460` switch cases — 7 group-level + 9 thread = 16 actions, matches the description verbatim |
| `content` description includes `thread-md-update` | ✅ | `src/agent-tools.ts:444-460` requires `content`; description at `src/agent-tools.ts:135` updated |
| Regression test exercises the exact reported scenario | ✅ | `cli/install.test.ts:179-224` — already-at-target + `entries.enabled=false` → no install, no restart, write with `"enabled": true` |
| Doctor `--fix` path unchanged | ✅ | `cli/doctor.ts:236-247` still calls `configSet(...)`; only the no-fix branch was downgraded to WARN |

`npm run build` / `npx tsc --noEmit` / `npm test` all reported green by the author; the diff is small enough that re-running locally is unlikely to surface anything beyond what the new test already covers.

## Findings

### P2 — Minor

**1. `ensurePluginsAllow()` is called twice on the legacy / deadlock paths.**

In `runLegacyMigration` (`cli/install.ts:214`) and `runDeadlockRepair` (`cli/install.ts:264`) the helper already runs after a successful install. The new unconditional call at `cli/install.ts:135` then runs it a second time. It's idempotent (returns early when the entry is already in `plugins.allow` — see `cli/openclaw-cli.ts:920-928`), so this is harmless, just a redundant file read. Could be cleaned up in a follow-up by removing the in-scenario calls now that the post-switch block always runs.

**2. Doctor `--fix` and install self-heal use different write paths.**

- `runDoctorChecks` with `--fix` writes via `configSet` → external `openclaw config set` (`cli/doctor.ts:238`).
- `runInstall` writes via `ensurePluginEnabled` → direct `writeConfigAtomic` (`cli/openclaw-cli.ts:944-958`).

Both end up flipping the same key, but if OpenClaw ever adds validation/normalization on `config set`, the two paths diverge silently. The doc-comment on `ensurePluginEnabled` already calls this out, so this is just a forward-looking note — consider unifying both callers on `ensurePluginEnabled` later, or at least documenting the canonical path in a single place.

**3. Regression test assertion is a substring match.**

`cli/install.test.ts:222` does `writes.find((w) => w.includes('"enabled": true'))`. Today the only thing in the written JSON that could contain `"enabled": true` is the plugin entry, but if the config grows other `enabled` flags this could match the wrong write. Asserting against the parsed JSON (`JSON.parse(write).plugins.entries["openclaw-channel-dmwork"].enabled === true`) would be more precise and self-documenting. Non-blocking.

**4. Edge cases not covered by the new test.**

The regression test only exercises the `entries["openclaw-channel-dmwork"].enabled === false` shape. Branches in `ensurePluginEnabled` not exercised by any test:

- `entries["openclaw-channel-dmwork"]` missing entirely (uses `??=` to create it).
- `entries["openclaw-channel-dmwork"].enabled` `undefined` (e.g. partially-initialized entry).
- `readConfigFromFile()` returns `null` (returns `false` cleanly).

The branches are all simple one-liners and the helper is best-effort, so test gaps here are low-risk. Fine to defer.

### Nits

- `cli/install.ts:91-94`: the comment "skip install and restart, but still self-heal config below" reads slightly awkward — at this point the function may not actually self-heal (e.g. config file unreadable). Minor wording.
- `src/agent-tools.ts:130`: "voice-context-*" is friendly but slightly less greppable than listing the three actions; readers may also wonder whether `voice-context-update` requires `groupId` for any reason. Acceptable as-is given the description is already long.

### Out of scope (informational)

- The two commit messages combine well-articulated reproduction details with the fix; nice to have on the record.
- No CI workflow exists in this repo yet; per the PR body, checks were run locally. Worth a follow-up to add a basic GitHub Actions job for `tsc --noEmit` + `vitest run` so future PRs don't rely solely on local verification.

## Verdict

The fixes are correct, minimal, and well-tested for the reported scenarios. The notes above are all P2 / nits and don't need to block merge.

PR #9, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -1,0 +1,20 @@
+name: Octo Issue Feed
+
+on:
+  issues:
+    types: [opened, closed, reopened, labeled]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      issue_number: ${{ github.event.issue.number }}
+      issue_title: ${{ github.event.issue.title }}
+      issue_url: ${{ github.event.issue.html_url }}
+      issue_author: ${{ github.event.issue.user.login }}
+      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
+      event_action: ${{ github.event.action }}
+      project_group_id: 0262cefe4463468bb6f2accaeda623cf
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,0 +1,23 @@
+name: Octo PR Feed
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, ready_for_review, review_requested]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      pr_number: ${{ github.event.pull_request.number }}
+      pr_title: ${{ github.event.pull_request.title }}
+      pr_url: ${{ github.event.pull_request.html_url }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      pr_merged: ${{ github.event.pull_request.merged }}
+      pr_additions: ${{ github.event.pull_request.additions }}
+      pr_deletions: ${{ github.event.pull_request.deletions }}
+      pr_changed_files: ${{ github.event.pull_request.changed_files }}
+      event_action: ${{ github.event.action }}
+      project_group_id: 0262cefe4463468bb6f2accaeda623cf
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}


yujiawei (commented):
# Code Review — PR #9 (octo-adapters)

## 1. Verification

| Item | Status | Evidence |
|------|--------|----------|
| Caller workflows reference an existing reusable workflow | ✅ | `Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml` and `octo-pr-feed.yml` both exist on `main`. |
| Inputs match the reusable workflow signatures | ✅ | All 8 issue inputs and 10 PR inputs (incl. `OCTO_BOT_TOKEN` secret) are present and correctly typed. |
| User-controlled fields can't be turned into a shell-injection vector | ✅ | The reusable workflow binds `inputs.*` to `env:` and consumes them via `os.environ[...]` in Python (not via `${{ }}` interpolation inside `run:`). Safe. |
| `pr_merged` correctly sourced for the closed/merged distinction | ✅ | `.github/workflows/octo-pr-feed.yml:18` passes `github.event.pull_request.merged`, which the reusable workflow uses to pick 🟣 vs 🔴. |
| Trigger sets are reasonable | ✅ | Issues: `opened, closed, reopened, labeled` — `.github/workflows/octo-issue-feed.yml:5`. PRs: `opened, closed, reopened, ready_for_review, review_requested` — `.github/workflows/octo-pr-feed.yml:5`. Sensible defaults; avoids `synchronize`/`edited` noise. |

Net: the PR does what it advertises. The integration is clean and the data flow is safe end-to-end.

## 2. Findings

### P2 — Reusable workflow is pinned to `@main`, not a SHA or tag

`.github/workflows/octo-issue-feed.yml:9` and `.github/workflows/octo-pr-feed.yml:9`:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
```

GitHub's hardening guide recommends pinning third-party (and even sister-org) reusable workflows to an immutable commit SHA, so a regression or compromise of the `.github` repo's `main` cannot silently propagate to every caller. Same-org `.github` repos are conventionally trusted, so this is advisory rather than blocking, but pinning to a tag (e.g. `@v1`) and bumping intentionally is a low-cost upgrade.

### P2 — `project_group_id` hardcoded in two places

`.github/workflows/octo-issue-feed.yml:18` and `.github/workflows/octo-pr-feed.yml:21` both literally embed `0262cefe4463468bb6f2accaeda623cf`. If the project channel is ever rotated, both files have to be edited in lockstep. A repo-level variable (`vars.OCTO_PROJECT_GROUP_ID`) would centralize this and make the caller workflows copy-pasteable across repos with only the variable changing.

### P2 — No explicit `permissions:` block on the caller

Neither caller declares `permissions:`, so the called job inherits the repo default `GITHUB_TOKEN` permissions. The notify job calls an external IM API and does not need any GitHub API scope; declaring `permissions: {}` (or `contents: read`) at the caller level enforces least privilege defense-in-depth, especially relevant since both triggers fire on `pull_request` / `issues` events that can originate from forks.

Note: for `pull_request` from forks the token is already read-only by default, so this is hardening, not a vulnerability.

### P2 — No `concurrency` group

A burst of label changes (e.g. labeling an issue with 5 labels in a row) will queue 5 notifier jobs and produce 5 IM messages. A `concurrency` group keyed on `${{ github.event.issue.number }}` with `cancel-in-progress: true` would collapse the burst. Optional — depends on tolerance for IM noise.

### P3 — `unlabeled` not in the issue trigger list

`.github/workflows/octo-issue-feed.yml:5` includes `labeled` but not `unlabeled`. If "label changed" is meant to be observable, the asymmetry is surprising; if labels are only useful at first add, the current set is fine. Worth a one-line comment in the workflow if this is intentional.

## 3. Recommendations

1. Switch `@main` → a tagged release of `Mininglamp-OSS/.github` (or a SHA) and adopt a Dependabot rule for `github-actions` so upgrades are explicit.
2. Move `project_group_id` to a repo/org variable; reference as `${{ vars.OCTO_PROJECT_GROUP_ID }}`.
3. Add `permissions: {}` to each caller job (or at workflow level) to enforce least privilege.
4. (Optional) Add `concurrency:` to suppress duplicate notifications during rapid label/state changes.
5. Decide whether `unlabeled` belongs in the trigger set, and either add it or document its omission.

## 4. Additional observations

- The reusable workflow logs only the first 8 chars of each group ID — good (avoids leaking project IDs in public Action logs).
- The IM API endpoint `https://im.deepminer.com.cn/api/v1/bot/sendMessage` returns no error handling on non-2xx responses; failures are visible only as an HTTP status print. Out of scope for this PR but worth raising upstream in `Mininglamp-OSS/.github`.
- The two workflows are nearly identical scaffolding; a future cleanup could collapse them into a single matrix workflow, though the current shape mirrors the two reusable workflows 1:1, which is easier to read.

Overall: small, well-targeted change. None of the above blocks the merge that already happened; treat as forward-looking polish.

PR #10, opened 2026-05-13
fix: use pull_request_target for fork secret access

lml2468:
### Problem
Workflows triggered by `pull_request` from **forked repositories** do not receive repository secrets. This means `OCTO_BOT_TOKEN` is unavailable, and every external contributor's PR triggers a failed workflow run.

### Fix
Changed trigger from `pull_request` → `pull_request_target`.

This is safe because:
- The workflow only reads event payload metadata (PR title, author, URL)
- No PR code is checked out
- `pull_request_target` runs in the context of the base branch, giving access to secrets

Applied to all 8 repos.

### DIFF
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -4,9 +4,11 @@ on:
   pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
 
+permissions: {}  # GITHUB_TOKEN not used; lock down to minimum
+
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,7 +1,7 @@
 name: Octo PR Feed
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
 
 jobs:


Jerry-Xin (dismissed):
The PR is relevant to `Mininglamp-OSS/octo-adapters` and the workflow change is narrowly scoped to fixing this repository’s PR notification workflow for forked PRs.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion — `.github/workflows/octo-pr-feed.yml:3`  
Consider adding explicit minimal permissions for this `pull_request_target` workflow, for example `permissions: contents: read` or `permissions: {}` if the reusable workflow does not need `GITHUB_TOKEN`. This reduces risk if repository defaults are broader than needed.

🔵 Suggestion — `.github/workflows/octo-pr-feed.yml:9`  
The reusable workflow is referenced via `@main` while receiving `OCTO_BOT_TOKEN`. For stronger supply-chain control, consider pinning to a reviewed tag or commit SHA, especially now that fork-triggered runs can access this secret.

✅ Highlights  
The workflow does not check out PR code, and it passes only GitHub event metadata into the reusable workflow. Based on the current caller file, switching from `pull_request` to `pull_request_target` is the correct mechanism for allowing fork PR notifications to use repository secrets.

lml2468 (commented):
Independent cross-review by 齐静春 (via lml2468 account — cannot self-approve).

**Verdict: APPROVED** — safe use of `pull_request_target`.

**Summary**: Single-line change `pull_request` → `pull_request_target` to give fork PRs access to `OCTO_BOT_TOKEN` for IM notifications.

**Security analysis** (verified by reading the reusable workflow at `Mininglamp-OSS/.github`):

1. **No code checkout** — workflow never runs `actions/checkout` on the fork, so no attacker-controlled code executes.
2. **Safe input handling** — PR metadata (title, author, etc.) is passed to the reusable workflow as typed `inputs`, then assigned to `env:` variables and read via Python `os.environ[]`. No `${{ inputs.* }}` in shell `run:` blocks. This is the correct pattern that prevents script injection.
3. **Quoted heredoc** — `<< 'PYEOF'` prevents shell expansion inside the Python block.
4. **Minimal secret surface** — only `OCTO_BOT_TOKEN` (IM notification bot), not a PAT or deployment credential.

🔴 Blocking

None.

💬 Non-blocking

🔵 Agree with Allen's suggestion: add explicit `permissions:` block. This workflow needs zero `GITHUB_TOKEN` scopes (the reusable workflow uses its own bot token, not `GITHUB_TOKEN`), so `permissions: {}` is the tightest option:

```yaml
on:
  pull_request_target:
    types: [opened, closed, reopened, ready_for_review, review_requested]

permissions: {}
```

This would prevent privilege escalation if the reusable workflow is later modified to include `actions/checkout` or other steps that use `GITHUB_TOKEN`.

✅ Highlights

- Correct and minimal fix for the fork secret access problem.
- The reusable workflow pattern (caller → org-level shared workflow) keeps the notification logic centralized.

yujiawei (commented):
# Code Review — PR #10 (octo-adapters)

## Summary

The diagnosis is correct and the fix is appropriate. Fork-originated `pull_request` events run with an empty secret context, so `OCTO_BOT_TOKEN` is unavailable and the notification job fails on every external contribution. Switching the caller to `pull_request_target` is the canonical workaround when (and only when) the workflow does not check out PR head code.

I reviewed both the caller (`.github/workflows/octo-pr-feed.yml` in this repo) and the reusable workflow it invokes (`Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main`). The standard `pull_request_target` footguns are all avoided here.

## 1. Verification

| Check | Result | Evidence |
|-------|--------|----------|
| No checkout of PR head code | ✅ | Neither caller nor reusable workflow uses `actions/checkout`; the reusable workflow only reads event metadata and posts to an IM API. |
| Untrusted inputs not shell-interpolated | ✅ | `pr_title`, `pr_author`, `pr_url`, `repo_name` flow into the reusable workflow's `env:` block and are read via `os.environ[...]` in Python. No `${{ inputs.* }}` expansion inside `run:` shell strings. |
| Reusable workflow source is trusted | ✅ | Same org (`Mininglamp-OSS/.github`). |
| Trigger types exclude `synchronize` | ✅ | Types are `opened, closed, reopened, ready_for_review, review_requested` — won't re-fire on every fork push. |
| Token scope appropriate | ✅ | `OCTO_BOT_TOKEN` is an IM bot token; blast radius is limited to the notification channel. |

## 2. Findings

### P2 — Add explicit `permissions:` block to the caller

`pull_request_target` runs with the repository's default `GITHUB_TOKEN` permission set, which on many older repos is `read/write` to the entire repo contents. This workflow does not use `GITHUB_TOKEN` at all (it only sends an outbound HTTPS request with `OCTO_BOT_TOKEN`), so the default token should be revoked at the workflow level.

Suggested change in `.github/workflows/octo-pr-feed.yml`:

```yaml
on:
  pull_request_target:
    types: [opened, closed, reopened, ready_for_review, review_requested]

permissions: {}
```

This is defense in depth — if a future change ever introduces a checkout or a script that uses `GITHUB_TOKEN`, an empty permissions block prevents the unintended write access from leaking into a `pull_request_target` context.

### P2 — Consider pinning the reusable workflow to a commit SHA

The caller references `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main`. Anyone with push access to `main` of `Mininglamp-OSS/.github` can change what runs with `OCTO_BOT_TOKEN`. Pinning to a SHA (and bumping it via Dependabot) tightens the supply chain:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@<commit-sha>
```

Not blocking — but worth doing since the same pattern is being rolled out across 8 repos.

## 3. Additional Observations

- **PR scope vs. description.** The PR body says "Applied to all 8 repos", but this PR only touches `octo-adapters`. I verified the 7 sibling PRs exist (`octo-web#18`, `octo-smart-summary#5`, `octo-lib#4`, `octo-deployment#4`, `octo-matter#5`, `octo-server#15`, `octo-admin#6`). Recommend merging them together (or close in time) so the rollout is consistent and the IM feed doesn't go silent on a subset of repos that already failed under `pull_request`.
- **First-time contributor approval.** `pull_request_target` does not require maintainer approval before the first run from a new contributor (unlike `pull_request` with the default org setting). This is acceptable here because the workflow is read-only on event payload, but worth noting if the reusable workflow ever grows.

## 4. Verdict

The change is correct, minimal, and safe. The two hardening items above are recommendations, not blockers. Approving the intent; flagging as comment so the `permissions: {}` and SHA-pin suggestions can be considered before this pattern is locked in across all 8 repos.

Jerry-Xin (approved):
The PR is relevant to this repository and safely fixes fork PR notification failures by moving the PR feed workflow to `pull_request_target`.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
- `.github/workflows/octo-pr-feed.yml:4` uses `pull_request_target` for a metadata-only notification workflow, which is appropriate because the job does not check out or execute PR code.
- `.github/workflows/octo-pr-feed.yml:7` explicitly removes `GITHUB_TOKEN` permissions, reducing token exposure.
- `.github/workflows/octo-pr-feed.yml:11` pins the reusable workflow to a commit SHA instead of `@main`, which is important for `pull_request_target` safety.
- The pinned reusable workflow only reads environment variables from the PR event payload and sends a notification; I did not find PR-code execution or checkout behavior. No additional tests are required for this workflow-only change.

yujiawei (approved):
# Code Review — PR #10 (octo-adapters)

## Summary

The fix is correct, minimal, and now fully hardened. The two non-blocking suggestions from the previous round (explicit `permissions:` block and SHA-pinning the reusable workflow) have both been addressed in `0b89520`. I re-verified the caller and the pinned reusable workflow at `Mininglamp-OSS/.github@c6c51348fccac99161166dce1b52dec1063aeea2` — the `pull_request_target` usage is safe.

## 1. Verification on HEAD `0b89520`

| Check | Result | Evidence |
|-------|--------|----------|
| `pull_request_target` is appropriate (fork PRs need `OCTO_BOT_TOKEN`) | ✅ | `.github/workflows/octo-pr-feed.yml:4` |
| No checkout of PR head code | ✅ | Neither caller nor reusable workflow uses `actions/checkout`. |
| Untrusted PR metadata not shell-interpolated | ✅ | Reusable workflow assigns `inputs.pr_title`, `inputs.pr_author`, etc. to `env:` and reads them via `os.environ[...]` inside a single-quoted heredoc (`<< 'PYEOF'`). No `${{ inputs.* }}` inside `run:` strings. |
| `GITHUB_TOKEN` locked down | ✅ | `permissions: {}` at `.github/workflows/octo-pr-feed.yml:7`. The reusable workflow uses only `OCTO_BOT_TOKEN` for an outbound HTTPS call, so zero scopes is correct. |
| Reusable workflow pinned to immutable ref | ✅ | `uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2` (`.github/workflows/octo-pr-feed.yml:11`). |
| Trigger types exclude `synchronize` | ✅ | `[opened, closed, reopened, ready_for_review, review_requested]` — won't fire on every fork push. |
| Secret blast radius bounded | ✅ | `OCTO_BOT_TOKEN` is an IM-bot token, not a PAT or deploy credential. |

## 2. Findings

### P0 / P1 — None

### P2 — None for this PR

Both previously raised P2 items (explicit minimal `permissions:`, SHA-pinning the reusable workflow) have been resolved.

## 3. Additional Observations (non-blocking, future-facing)

- **Document the `pull_request_target` invariant.** A one-line comment near the trigger explaining *why* `pull_request_target` is safe here ("metadata-only — DO NOT add `actions/checkout` of PR head") would help future maintainers avoid silently breaking the safety property when extending this workflow. Optional.
- **PR description vs. PR scope.** The body says "Applied to all 8 repos", but this PR only touches `octo-adapters`. Sibling PRs are open against the other repos. Recommend landing them together (or close in time) so the IM feed isn't silent on a subset of repos that were already failing under `pull_request`.
- **Reusable workflow update path.** Since `@main` was replaced by a SHA, the caller will not auto-pick up future fixes to `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml`. Consider Dependabot's `package-ecosystem: github-actions` to bump the SHA, or a documented manual cadence. Tradeoff is intentional and acceptable; just calling it out.

## 4. Verdict

Approving. The diagnosis was right, the fix is the canonical workaround, and the hardening is complete (no checkout, no shell-injection vector, `permissions: {}`, SHA-pinned reusable workflow).

PR #12, opened 2026-05-13
fix(inbound): accumulate deliver chunks instead of overwriting them

### LINKED ISSUE #11: bug: deliver-buffer overwrites blocks/finals, causing assistant replies to be silently dropped
### Description

`openclaw-channel-dmwork@0.6.4` silently drops assistant replies in two distinct ways inside the `dispatchReplyWithBufferedBlockDispatcher` `deliver` callback in `src/inbound.ts:1685-1719`:

1. **LIFO overwrite** — multiple `block`/`final` payloads overwrite a single `lastText` slot (`deliverBuffer.lastText = content`), so when the assistant emits multiple text blocks **only the last one is delivered to the user**.
2. **onError nukes the buffer** — any non-fatal `onError` resets `lastText = null; textSent = true` and replaces the entire reply with a generic error string, throwing away anything already buffered.

Confirmed root cause aligns with the OpenClaw SDK semantics (verified against `openclaw@2026.5.7`):

- `kind: "block"` is a **streaming chunk** (SDK accumulates internally via `accumulatedBlockText += payload.text`), so chunks must be **concatenated**, not overwritten.
- `kind: "final"` is an **independent reply payload** distinct from `block`, so finals also must not overwrite each other.

### Steps to Reproduce

1. Install `openclaw-channel-dmwork@0.6.4` against OpenClaw `2026.5.7`.
2. Configure a DMWork bot, send a private-message conversation to the bot.
3. Ask the assistant to perform an action that emits a tool call followed by a multi-block text reply (e.g. "Create a group called X with users A, B, C and tell me when done"). The assistant typically generates several `block` payloads + a final summary.
4. Observe: only the last block (or nothing) reaches the user. The OpenClaw dashboard shows the full assistant output, but DMWork client receives at most one fragment.

Direct API verification (bypassing the plugin) confirms the DMWork server, network, and bot credentials are healthy:

```bash
curl -X POST https://im.deepminer.com.cn/api/v1/bot/sendMessage \
  -H "Authorization: Bearer <bot_token>" \
  -H "Content-Type: application/json" \
  -d '{"channel_id":"<uid>","channel_type":1,"payload":{"type":1,"content":"test"}}'
# → 200 OK, message delivered instantly
```

### Expected Behavior

All assistant-generated `block` and `final` payloads should reach the user as a single coherent reply (or sequentially) with the correct ordering. `block` chunks should be concatenated without inserting fake separators (they are streaming fragments). `final` payloads should be delimited from each other and from `block` content.

### Actual Behavior

| Scenario | Outcome |
|---|---|
| Multi-block reply | Only the **last** chunk arrives — earlier chunks are silently dropped |
| Reply during which `onError` fires | The whole reply is **discarded** and replaced with a generic "处理您的消息时遇到了问题" error, even when most blocks succeeded |
| `deliver` never fires (or all `content === ""`) | `finally` does nothing, **entire turn is lost without any log line** |

Captured production log lines (gateway.log, 2026-05-13):

```
17:29  recv message  → no sendMessage / no fallback (entire turn lost)
17:31  recv message  → fallback text sent (64 chars)   # actual reply was multi-paragraph
17:35  recv message  → fallback text sent (785 chars)  # actual reply was much longer
17:51  recv message  → no fallback (entire turn lost again)
18:02  recv message  → fallback text sent (683 chars)
```

### Adapter Version

0.6.4

### Affected Component

OpenClaw Channel Plugin

### Severity

P0 - Service Down / Data Loss

### Reproducibility

- [x] Consistently reproducible (100% in current session for any reply with tool call + multi-block text)

### Environment

- OpenClaw: `2026.5.7 (eeef486)`
- Plugin: `openclaw-channel-dmwork@0.6.4`
- Node.js: `v25.9.0`
- OS: `macOS Darwin 25.3.0 (arm64)`
- DMWork Server: `https://im.deepminer.com.cn/api`
- WebSocket: `wss://im.deepminer.com.cn/ws`
- Plugin install path: `~/.openclaw/npm/node_modules/openclaw-channel-dmwork/`
- `dmScope`: `per-channel-peer`
- Affected bot: `27nOoWsprYA7f77c333_bot` (Q420)

### Logs / Screenshots

The `deliver` callback never logs the `kind` or `content.length` it receives, so when an entire turn is dropped there is **no log line at all** to indicate what happened.

### Proposed Fix

Replace the LIFO `lastText` slot with two accumulating arrays that respect the SDK semantics:

```ts
const deliverBuffer = {
  blockChunks: [] as string[],   // streaming fragments → join("")
  finals:      [] as string[],   // independent payloads → join("\n\n")
  textSent:    false,
};
```

- `kind === "block"` → `blockChunks.push(content)` (concatenate raw, no separator)
- `kind === "final"` (or unknown kinds) → `finals.push(content)`
- `kind === "tool"` → continue to send immediately as today
- `finally` → join blocks with `""`, finals with `"\n\n"`, send once
- `onError` → **first attempt to flush already-buffered content**, then send the error notice

Add a `log.debug` at `deliver` entry with `kind` + `content.length` so that "entire turn lost" cases are at least visible in the log.

Will follow up with a PR.

caster-Q:
## Summary

Fix a P0 silent message-loss bug in the `deliver` callback used by
`dispatchReplyWithBufferedBlockDispatcher`. Multi-block assistant
replies were being overwritten and only the last chunk reached the
user; non-fatal `onError` events nuked any buffered content and
replaced the entire reply with a generic notice.

## Related Issue

Closes #11

## Changes

- `src/inbound.ts`:
  - Replace the single `lastText` slot with two accumulator arrays:
    `blockChunks` (joined with `""` — streaming fragments) and `finals`
    (joined with `"\n\n"` — independent payloads).
  - Block chunks now keep raw `payload.text` (no `trim()`) so spaces
    between streamed fragments are preserved by `join("")`. Final/tool
    payloads still trim because they are complete independent messages.
  - Extract a single `flushBufferedText(reason)` helper that combines
    blocks and finals, sends once, and is idempotent via `textSent`.
  - `onError` now calls `flushBufferedText("on-error")` BEFORE sending
    the generic error notice, so partial replies are preserved on
    non-fatal failures.
  - Add a `deliver` entry `log.debug` (`kind`/`contentLen`/`hasMedia`)
    plus per-kind buffer-append logs, so "entire turn lost" cases leave
    a trace.

- `src/__tests__/deliver-buffer.test.ts`:
  - Mirror the new helpers (`createDeliverBuffer`, `makeDeliver`,
    `makeFlushBufferedText`, `makeOnError`).
  - Replace overwrite-semantics expectations with accumulator
    expectations: multi-block concat, multi-final blank-line join,
    blocks-then-final ordering, tool + block + final mix.
  - Add regression tests for `onError` flushing buffered content first
    (with and without buffered content), idempotent flush, and the
    "deliver never called" no-op path.

## Testing

Local verification:

- `npx tsc --noEmit` — passes (0 errors)
- `npm test` — **699/699 tests pass** across 22 files (added 4 new
  regression tests, updated 10 existing cases to match new semantics)
- `npm run build` — passes

- [x] Manually verified via test cases mirroring the production failure
      modes in #11

## Checklist

- [x] I have read CONTRIBUTING.md
- [ ] Updated documentation — N/A (internal pipeline behavior)

### DIFF
--- a/openclaw-channel-dmwork/src/__tests__/deliver-buffer.test.ts
+++ b/openclaw-channel-dmwork/src/__tests__/deliver-buffer.test.ts
@@ -3,22 +3,43 @@ import { describe, it, expect, vi } from "vitest";
 /**
  * Unit tests for the deliver buffer pattern used in handleInboundMessage.
  *
- * The deliver callback uses info.kind to decide behavior:
- * - "tool"  → send immediately via resolveAndSendText (verbose output)
- * - "block" / "final" / other → buffer text (overwrite), send once in finally
+ * The deliver callback dispatches by `info.kind`:
+ *   - "tool"  → send immediately via sendTextFn (verbose tool output)
+ *   - "block" → push into blockChunks (streaming fragment, joined with "")
+ *   - "final" / unknown → push into finals (independent payload, joined with "\n\n")
+ *   - isReasoning payloads are skipped entirely
+ *   - media payloads are sent immediately with dedup
  *
- * - isReasoning payloads are skipped entirely
- * - Media payloads are always sent immediately with dedup
- * - The finally block sends the last buffered text after dispatcher finishes
- * - onError clears the buffer to prevent stale text
+ * After the dispatcher finishes, `flushBufferedText` joins blockChunks (no separator,
+ * stream semantics) and finals (blank-line separator, payload semantics), and sends
+ * everything once. `onError` flushes whatever was buffered first, then sends a
+ * generic error notice.
+ *
+ * This mirrors the production logic in inbound.ts; if you change one, change the other.
  */
 
 // ---- helpers that mirror the production logic in inbound.ts ----
 
 function createDeliverBuffer() {
   return {
-    lastText: null as string | null,
+    blockChunks: [] as string[],
+    finals: [] as string[],
     textSent: false,
+  };
+}
+
+function makeFlushBufferedText(
+  deliverBuffer: ReturnType<typeof createDeliverBuffer>,
+  sendTextFn: (text: string) => Promise<void>,
+) {
+  return async (_reason: "finally" | "on-error"): Promise<void> => {
+    if (deliverBuffer.textSent) return;
+    const blockText = deliverBuffer.blockChunks.join("");
+    const finalText = deliverBuffer.finals.join("\n\n");
+    const parts = [blockText, finalText].filter((s) => s.length > 0);
+    if (parts.length === 0) return;
+    deliverBuffer.textSent = true;
+    await sendTextFn(parts.join("\n\n"));
   };
 }
 
@@ -37,12 +58,10 @@ function makeDeliver(
     },
     info?: { kind?: string },
   ) => {
-    // Skip reasoning blocks
     if (payload.isReasoning) return;
 
     const kind = info?.kind ?? "final";
 
-    // Media: send immediately with dedup
     const outboundMediaUrls = [
       ...(payload.mediaUrls ?? []),
       ...(payload.mediaUrl ? [payload.mediaUrl] : []),
@@ -58,87 +77,123 @@ function makeDeliver(
       }
     }
 
-    // Text handling based on kind
-    const content = payload.text?.trim() ?? "";
+    const content = (payload.text ?? "").trim();
+    const rawText = payload.text ?? "";
     if (!content && outboundMediaUrls.length > 0) return;
     if (!content) return;
 
     if (kind === "tool") {
-      // Verbose tool call output: send immediately
       await sendTextFn(content);
       return;
     }
 
-    // kind === "block" / "final" / anything else: buffer, send once after dispatcher finishes
-    deliverBuffer.lastText = content;
+    if (kind === "block") {
+      // Preserve raw text (including spaces) so join("") recovers the original stream.
+      deliverBuffer.blockChunks.push(rawText);
+      return;
+    }
+
+    // "final" or unknown kinds — trim, independent payloads
+    deliverBuffer.finals.push(content);
   };
 }
 
 function makeOnError(
   deliverBuffer: ReturnType<typeof createDeliverBuffer>,
+  flushBufferedText: (reason: "finally" | "on-error") => Promise<void>,
   sendErrorFn: () => Promise<void>,
 ) {
   return async (_err: unknown) => {
-    deliverBuffer.lastText = null;
+    // Try to flush already-buffered partial output first; do NOT discard.
+    await flushBufferedText("on-error");
     deliverBuffer.textSent = true;
     await sendErrorFn();
   };
 }
 
-async function runFinally(
-  deliverBuffer: ReturnType<typeof createDeliverBuffer>,
-  sendTextFn: (text: string) => Promise<void>,
-) {
-  if (deliverBuffer.lastText && !deliverBuffer.textSent) {
-    deliverBuffer.textSent = true;
-    await sendTextFn(deliverBuffer.lastText);
-  }
-}
-
 // ---- tests ----
 
 describe("deliver buffer pattern", () => {
-  it("block kind: buffers text without sending", async () => {
+  it("multiple blocks: all chunks accumulate, finally joins them with no separator", async () => {
+    // Regression: 0.6.4 only kept the LAST block, dropping every prior chunk.
     const deliverBuffer = createDeliverBuffer();
     const sentMediaUrls = new Set<string>();
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, sentMediaUrls, sendMedia, sendText);
+    const flush = makeFlushBufferedText(deliverBuffer, sendText);
 
-    await deliver({ text: "Hello" }, { kind: "block" });
-    await deliver({ text: "Hello, how are you" }, { kind: "block" });
-    await deliver({ text: "Hello, how are you? I'm here to help." }, { kind: "block" });
+    // Streaming fragments — each block is a NEW slice of text, not a snapshot.
+    await deliver({ text: "Hello, " }, { kind: "block" });
+    await deliver({ text: "world! " }, { kind: "block" });
+    await deliver({ text: "This is a test." }, { kind: "block" });
 
-    // Text should NOT have been sent
+    // Nothing sent yet
     expect(sendText).not.toHaveBeenCalled();
-    // Buffer should have the latest text
-    expect(deliverBuffer.lastText).toBe("Hello, how are you? I'm here to help.");
+    expect(deliverBuffer.blockChunks).toHaveLength(3);
     expect(deliverBuffer.textSent).toBe(false);
 
-    // finally block sends the buffered text
-    await runFinally(deliverBuffer, sendText);
+    // finally → joined with "" (stream semantics)
+    await flush("finally");
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Hello, how are you? I'm here to help.");
+    expect(sendText).toHaveBeenCalledWith("Hello, world! This is a test.");
     expect(deliverBuffer.textSent).toBe(true);
   });
 
-  it("final kind: buffers text, sends via finally", async () => {
+  it("single final: buffered, sent via finally", async () => {
     const deliverBuffer = createDeliverBuffer();
     const sentMediaUrls = new Set<string>();
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, sentMediaUrls, sendMedia, sendText);
+    const flush = makeFlushBufferedText(deliverBuffer, sendText);
 
     await deliver({ text: "Final answer" }, { kind: "final" });
 
-    // final is buffered, NOT sent immediately
     expect(sendText).not.toHaveBeenCalled();
-    expect(deliverBuffer.lastText).toBe("Final answer");
+    expect(deliverBuffer.finals).toEqual(["Final answer"]);
 
-    // finally block sends it
-    await runFinally(deliverBuffer, sendText);
+    await flush("finally");
     expect(sendText).toHaveBeenCalledTimes(1);
     expect(sendText).toHaveBeenCalledWith("Final answer");
+  });
+
+  it("multiple finals: each independent payload, joined with blank lines", async () => {
+    // Regression: 0.6.4 only kept the LAST final.
+    const deliverBuffer = createDeliverBuffer();
+    const sentMediaUrls = new Set<string>();
+    const sendMedia = vi.fn().mockResolvedValue(undefined);
+    const sendText = vi.fn().mockResolvedValue(undefined);
+    const deliver = makeDeliver(deliverBuffer, sentMediaUrls, sendMedia, sendText);
+    const flush = makeFlushBufferedText(deliverBuffer, sendText);
+
+    await deliver({ text: "First reply" }, { kind: "final" });
+    await deliver({ text: "Second reply" }, { kind: "final" });
+
+    expect(sendText).not.toHaveBeenCalled();
+    expect(deliverBuffer.finals).toHaveLength(2);
+
+    await flush("finally");
+    expect(sendText).toHaveBeenCalledTimes(1);
+    expect(sendText).toHaveBeenCalledWith("First reply\n\nSecond reply");
+  });
+
+  it("blocks then final: stream concatenated then final separated by blank line", async () => {
+    const deliverBuffer = createDeliverBuffer();
+    const sentMediaUrls = new Set<string>();
+    const sendMedia = vi.fn().mockResolvedValue(undefined);
+    const sendText = vi.fn().mockResolvedValue(undefined);
+    const deliver = makeDeliver(deliverBuffer, sentMediaUrls, sendMedia, sendText);
+    const flush = makeFlushBufferedText(deliverBuffer, sendText);
+
+    await deliver({ text: "Working on it" }, { kind: "block" });
+    await deliver({ text: "..." }, { kind: "block" });
+    await deliver({ text: "Done!" }, { kind: "final" });
+
+    await flush("finally");
+    expect(sendText).toHaveBeenCalledTimes(1);
+    // block stream → "Working on it..."  ; then final → "Done!"
+    expect(sendText).toHaveBeenCalledWith("Working on it...\n\nDone!");
   });
 
   it("tool kind: sends text immediately, does not affect buffer", async () => {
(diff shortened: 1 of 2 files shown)


Jerry-Xin (changes requested):
The PR is relevant to this repository, but the new block accumulation can corrupt multi-block replies.

🔴 Blocking

🔴 Critical — [openclaw-channel-dmwork/src/inbound.ts](/private/tmp/pr-review-rSXl39/repo/openclaw-channel-dmwork/src/inbound.ts:1533): `deliverBuffer.blockChunks.join("")` assumes every `kind: "block"` payload is a raw streaming delta with its own spacing. That is not a safe contract for the OpenClaw buffered block dispatcher. Its block chunker can emit display-ready chunks split at paragraph/newline/sentence/whitespace boundaries, and those splits may remove the separator from both sides. Joining with `""` can merge paragraphs or words, e.g. `"First paragraph"` + `"Second paragraph"` becomes `"First paragraphSecond paragraph"`.

The tests reinforce the unsafe assumption at [deliver-buffer.test.ts](/private/tmp/pr-review-rSXl39/repo/openclaw-channel-dmwork/src/__tests__/deliver-buffer.test.ts:117) by only covering raw delta-like chunks with embedded spaces. Please either preserve block payload boundaries when flushing, or add a separator strategy that matches dispatcher chunk semantics and covers paragraph/newline/whitespace-split blocks.

💬 Non-blocking

🔵 Suggestion — The test file mirrors production logic instead of exercising the actual `handleInboundMessage` delivery path. That is useful for small logic checks, but this bug is specifically about dispatcher semantics, so at least one integration-style test with realistic block chunks would reduce drift risk.

I could not run the suite in this checkout because dependencies are not installed: `npm test` failed with `vitest: command not found`, and `npx tsc --noEmit` tried to install the unrelated `tsc` package instead of using a local TypeScript install.

✅ Highlights

The relevance gate passes: this PR modifies the Dmwork inbound reply delivery path in the OpenClaw adapter. The intent to stop overwriting buffered replies and preserve partial output on `onError` is correct; the blocker is the exact reconstruction of block text.

yujiawei (commented):
# Code Review — PR #12 (octo-adapters)

## Summary

This PR fixes a real P0 dataloss bug (#11): the previous `lastText` slot in `dispatchReplyWithBufferedBlockDispatcher`'s `deliver` callback was overwritten by every subsequent `block`/`final` payload, so multi-block assistant replies arrived as fragments and `onError` discarded any partial output before sending the generic notice. The fix replaces the slot with two accumulator arrays (`blockChunks`, `finals`), introduces an idempotent `flushBufferedText(reason)` that's reused by both the `finally` block and `onError`, and adds visibility logs. The accumulator semantics — `join("")` for streaming block fragments, `\n\n` between independent finals — match the stated SDK contract (`accumulatedBlockText += payload.text`).

The diff is focused, the test coverage is genuinely strengthened (multi-block, multi-final, blocks+final interleaving, `onError` with/without buffered content, idempotent flush, deliver-never-called), and the change has been verified against the failure modes in #11. **Verdict: COMMENTED** — the fix is correct and ship-quality for the documented bug, but there are a few edge cases worth resolving before/shortly after merge.

## 1. Verification

| Claim | Status | Evidence |
|---|---|---|
| Multi-block accumulation no longer drops earlier chunks | ✅ | `src/inbound.ts:1733-1738`; test `src/__tests__/deliver-buffer.test.ts:107-130` |
| Multi-final accumulation joins with blank lines | ✅ | `src/inbound.ts:1741-1744`; test `:152-171` |
| `onError` flushes buffered partial text before generic notice | ✅ | `src/inbound.ts:1749-1753`; test `:296-322` |
| `onError` with empty buffer sends only the error notice | ✅ | test `:325-339` |
| `flushBufferedText` is idempotent (`textSent` short-circuit) | ✅ | `src/inbound.ts:1532`; test `:341-355` |
| Tool-kind path still sends inline (unchanged) | ✅ | `src/inbound.ts:1722-1727` |
| Media path still sends inline with dedup (unchanged) | ✅ | `src/inbound.ts:1683-1699` |
| `kind` defaulting (`info?.kind ?? "final"`) covers `undefined info` and `undefined kind` | ✅ | `src/inbound.ts:1680`; test `:236-247` |
| Visibility log at deliver entry | ⚠️ | `src/inbound.ts:1712-1714` — present but at `debug` level; see P1 below |

CI `notify` job is green; `code-review` is the gate this review unblocks.

## 2. Findings

### P1 — Whitespace-only block deltas are still silently dropped

`src/inbound.ts:1706-1720`:

```ts
const rawText = payload.text ?? "";
const content = rawText.trim();
...
if (!content && outboundMediaUrls.length > 0) { ... return; }
if (!content) return;            // <-- early return BEFORE the block branch
...
if (kind === "block") {
  deliverBuffer.blockChunks.push(rawText);   // never reached when rawText is " "
  ...
}
```

The PR description explicitly says:

> Block chunks now keep raw `payload.text` (no `trim()`) so spaces between streamed fragments are preserved by `join("")`.

That is true for the *push* itself, but the early `if (!content) return;` runs first and uses the **trimmed** content. If the upstream SDK ever emits a pure-whitespace delta (e.g. a tokenizer that yields `"Hello"`, `" "`, `"world"` as three separate `block` payloads), the middle `" "` block is dropped here, and `join("")` produces `"Helloworld"` — the very class of bug this PR is trying to eliminate.

Whether this triggers in practice depends on the OpenClaw SDK's stream chunking. The current tests don't cover it — every block in the new tests already carries non-whitespace content (`"Hello, "`, `"world! "`, etc.). Two options:

1. (Preferred) For `kind === "block"`, gate on `rawText.length === 0` instead of trimmed content:
   ```ts
   if (kind === "block") {
     if (rawText.length === 0) return;
     deliverBuffer.blockChunks.push(rawText);
     ...
   }
   ```
   Keep the trimmed `if (!content) return;` for `final`/`tool` only.
2. Document the assumption in the comment block (e.g. "SDK never emits pure-whitespace block deltas") and add a test that asserts on `[" "]` if you can verify it from the SDK source.

This is the only finding I'd want resolved either in this PR or as an immediate follow-up; everything below is P2.

### P2 — Interleaved block↔final order is not preserved

`flushBufferedText` concatenates *all* blocks first, then *all* finals:

```ts
const blockText = deliverBuffer.blockChunks.join("");
const finalText = deliverBuffer.finals.join("\n\n");
const parts = [blockText, finalText].filter((s) => s.length > 0);
```

If the dispatcher ever emits in the order `block → final → block` (e.g. an answer that re-streams text after a finalized intermediate payload), the output rearranges to all-blocks-first. The current "blocks-then-final" test (`:175-192`) doesn't cover that case. If the SDK contract guarantees the ordering `block* (final|tool)*`, please make that explicit in the comment at `src/inbound.ts:1513-1519` so future readers don't reach for the array order without checking.

If interleaving *is* possible, you'd need a single ordered list of `{kind, text}` records and join differently, which is a larger refactor.

### P2 — `flushBufferedText` marks `textSent = true` before awaiting the actual send

`src/inbound.ts:1537-1549`:

```ts
deliverBuffer.textSent = true;     // set BEFORE the send
const combined = parts.join("\n\n");
try {
  await resolveAndSendText(combined);
  log?.info?.(...);
} catch (sendErr) {
  log?.error?.(`dmwork: [deliver-buffer] flush failed (reason=${reason}): ${String(sendErr)}`);
}
```

If `resolveAndSendText` throws, the buffer is logged as failed but the `finally` path won't retry, and there is no user-facing fallback. This is functionally identical to the pre-fix behavior, so it's not a regression — just calling it out because the surrounding code now invests heavily in *not* losing partial text. Consider one of:

- A single retry on transient failure, or
- Falling back to sending an error notice to the user when both flush attempts have failed.

Out of scope for this PR; flag it as a follow-up.

### P2 — Visibility log is at `debug` level

Issue #11 specifically called out "the deliver callback never logs the kind or content.length it receives, so when an entire turn is dropped there is no log line at all." The new log at `src/inbound.ts:1712-1714` is at `log?.debug?.`. If the production logger defaults to `info` (which appears to be the case from the surrounding `log?.info?.` calls on the same path), this regression-detection log will still be invisible in prod. Either bump this single line to `info` (cheap, one line per inbound turn), or document that operators must enable `debug` to retain this visibility.

### P2 — Tests duplicate the production logic instead of importing it

`src/__tests__/deliver-buffer.test.ts` defines `createDeliverBuffer`, `makeDeliver`, `makeFlushBufferedText`, and `makeOnError` inline. These are hand-mirrored copies of the production logic in `inbound.ts`, and the test file even acknowledges this: *"This mirrors the production logic in inbound.ts; if you change one, change the other."*

Hand-synchronized test models reliably drift over time — a future change to `inbound.ts` can keep the tests green while breaking the real callback. Not a blocker, but worth a follow-up to extract the deliver-buffer state machine into an exported helper module that both `inbound.ts` and the tests import.

### P2 — Test for `outboundMediaUrls` doesn't use the production helper

The test's `makeDeliver` inlines:
```ts
const outboundMediaUrls = [
  ...(payload.mediaUrls ?? []),
  ...(payload.mediaUrl ? [payload.mediaUrl] : []),
];
```
while production calls `resolveOutboundMediaUrls(payload)` (`src/inbound.ts:39-44`), which also `.filter(Boolean)`s. Behaviorally equivalent today, but the same drift risk as above. Subsumed by the previous P2.

## 3. Recommendations

1. **Resolve P1 before merge** by guarding the early return in the `block` branch on `rawText.length === 0` instead of `!content`. Add one regression test: `[{text: "Hello"}, {text: " "}, {text: "world"}]` as three `block` payloads → flush should send `"Hello world"`.
2. Either bump the deliver-entry log to `info`, or note in the comment that `debug` must be enabled in prod for the "entire turn dropped" diagnostic to work.
3. Capture the SDK's ordering guarantee for `block` vs `final` in the comment block at `src/inbound.ts:1513-1519`. If only `block* (final|tool)*` is guaranteed, that's enough — readers shouldn't have to infer it.
4. Follow-up (separate PR): extract the deliver-buffer state machine into `src/deliver-buffer.ts` and have both `inbound.ts` and the tests import it, replacing the hand-mirrored test helpers.
5. Follow-up (separate PR): consider a retry / user-visible error path when the buffered flush itself fails inside `resolveAndSendText`.

## 4. Additional Observations

- Backwards compatibility: `deliverBuffer` is a private local; no external consumers. ✅
- No new dependencies; `npx tsc --noEmit` and `npm test` pass per the PR body. ✅
- Comment quality is unusually good — the prose at `src/inbound.ts:1513-1550, 1701-1714, 1729-1744` explains the *why* (regression, SDK semantics, idempotency reasoning) rather than restating the code. Keep this style.
- `onError` uses `sendMessage` directly (not `resolveAndSendText`) for the error notice. Correct — the canned error string contains no mentions, so skipping mention-resolution is appropriate.
- The CHANGELOG / docs aren't touched. The PR body marks docs as N/A; given this is a P0 user-visible bug (silent dataloss), a one-line entry in the package's release notes / `CHANGELOG.md` would help downstream operators correlate `0.6.4 → 0.6.5` upgrades with the disappearance of "missing reply" reports. Nit, not blocking.

lml2468 (approved):
Independent cross-review. **APPROVED** — this fixes a real P0 data-loss bug. Detailed SDK source analysis below.

## Summary

The old code overwrote `deliverBuffer.lastText` on every deliver call, silently dropping all but the last block/final. The fix accumulates into two arrays (`blockChunks`, `finals`) and flushes once via an idempotent `flushBufferedText`. The `onError` path now flushes partial output before sending the error notice, which is strictly better than the old discard-everything behavior.

## 🔴 Blocking

None.

## Allen's `join("")` concern — investigated against SDK source

Allen flagged `blockChunks.join("")` as unsafe because the `EmbeddedBlockChunker` can split at paragraph/sentence/newline boundaries, stripping separators from both sides. I traced the full dispatch chain to evaluate this:

**SDK internals** (`dispatch-8E8vi2HV.js:904-905`, `dispatch-acp-HFOhX_U-.js:600-601`):
```js
if (accumulatedBlockText.length > 0) accumulatedBlockText += "\n";
accumulatedBlockText += rawBlockText;
```
The SDK's internal accumulator uses `"\n"` between block deliveries.

**Block reply coalescer** (`block-reply-pipeline-DEYxmc4h.js`):
```js
const joiner = config.joiner ?? "";
const nextText = bufferText ? `${bufferText}${joiner}${text}` : text;
```
The coalescer joins consecutive chunks with `""` by default **before** each delivery reaches the channel plugin.

**EmbeddedBlockChunker** (`pi-embedded-block-chunker-B4G7Ez5P.js:84-87`):
```js
const chunk = `${reopenPrefix}${source.slice(start, paragraphBreak.index)}`;
if (chunk.trim().length > 0) emit(chunk);
start = skipLeadingNewlines(source, paragraphBreak.index + paragraphBreak.length);
```
Paragraph breaks ARE stripped from chunk boundaries.

**My conclusion**: Allen's concern is technically valid for the ACP path where `EmbeddedBlockChunker` is active — paragraph breaks between chunks can be lost. However:

1. **For non-ACP streaming** (raw LLM deltas, the most common DMWork path): each block delivery is a token-level delta with embedded spacing. `join("")` is correct.
2. **The SDK's own coalescer** already uses `""` joiner internally, so this same "issue" exists within the SDK pipeline before the channel even sees the data.
3. **The separator loss is cosmetic** (missing `\n` between paragraphs), not data loss. Going from "only the last chunk survives" to "all chunks with potentially missing newlines" is a massive net improvement.
4. Using `join("\n")` would match the SDK's `accumulatedBlockText` behavior and is a safe refinement, but should not block this P0 fix.

## 💬 Non-blocking

🔵 **Refinement**: Consider `blockChunks.join("\n")` to match the SDK's internal accumulation separator. This would handle the ACP chunker edge case while being harmless for raw streaming (deltas already contain their own newlines). Low urgency — can be a follow-up.

🔵 **yujiawei's P1 (whitespace-only block dropped)**: Valid in theory — the `if (!content) return;` early exit uses trimmed text, so a pure-space delta `" "` would be dropped before reaching the block push. However, both the `EmbeddedBlockChunker` (`if (rawChunk.trim().length === 0) return`) and the coalescer filter whitespace-only payloads upstream, so this never triggers in practice. Still, for defensive completeness, could guard the early return with `if (kind !== "block" && !content) return;`. Low urgency.

🔵 **Agree with yujiawei P2**: `flushBufferedText` sets `textSent = true` before `await resolveAndSendText()`. If the send throws, the finally path won't retry. Not a regression (old code had the same gap), but worth a follow-up.

🔵 **Agree with yujiawei P2**: The deliver-entry log at `debug` level won't be visible in production where the default is likely `info`. Bump to `info` for the "entire turn dropped" diagnostic to work.

## ✅ Highlights

- The fix correctly identifies and resolves the root cause: single-slot overwrite → array accumulation.
- `flushBufferedText` is idempotent and shared between `finally` and `onError` — clean design.
- `rawText` (untrimmed) for block push preserves inter-fragment spacing in the streaming path.
- Comprehensive regression tests with explicit annotations ("`Regression: 0.6.4 only kept the LAST block`").
- Comment quality is excellent — explains the SDK contract, why raw vs trimmed, and the idempotency reasoning.

PR #20, opened 2026-05-15
chore: remove pr-merged-done caller (superseded by Project built-in automation)

lml2468:
## What

Remove `.github/workflows/pr-merged-done.yml` caller stub.

## Why

The org-level reusable workflow (`Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml`) is being retired. GitHub Projects v2 already has a native **'Pull request merged'** automation enabled on Octo Board that handles this automatically — no custom workflow needed.

## Related

- Org-level removal: https://github.com/Mininglamp-OSS/.github/pull/13

PR #21, opened 2026-05-15
chore(workflows): pin reusable workflow refs to @v1

lml2468:
## Summary

Pin all reusable workflow references to `@v1` for stable, version-locked behavior.

**Before:** Mixed `@main` and hardcoded commit SHAs
**After:** All callers use `Mininglamp-OSS/.github/...@v1`

## Why

The `.github` repo has been tagged `v1` as the stable baseline. Using a version tag
ensures callers get a predictable, reviewed version of the shared workflows.

No caller-side logic changes — this is a reference update only.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -11,6 +11,6 @@ permissions: {}
 
 jobs:
   add-to-project:
-    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@239d0b47cfe1947d5e6de826e795feac4ed90562
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@v1
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -8,4 +8,4 @@ jobs:
   welcome:
     permissions:
       issues: write
-    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@v1
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       issue_number: ${{ github.event.issue.number }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -8,7 +8,7 @@ permissions: {}  # GITHUB_TOKEN not used; lock down to minimum
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}


yujiawei (approved):
# Code Review — PR #21 (octo-adapters)

## Summary

Pure reference update across four caller workflows: replace mixed `@main` / hardcoded-SHA refs with the new `@v1` annotated tag on `Mininglamp-OSS/.github`. No caller-side logic changes. Recommendation: **approve**.

## Verification

| File | Before | After | Notes |
|---|---|---|---|
| `.github/workflows/auto-add-to-project.yml` | `@239d0b47cfe1947d5e6de826e795feac4ed90562` (SHA) | `@v1` | Workflow blob at the old SHA (`e9365d5`) is byte-identical to the blob at `v1`. No behavioral change. |
| `.github/workflows/issue-welcome.yml` | `@main` (floating) | `@v1` | Floating → version tag. Net improvement (pinned, reviewable). |
| `.github/workflows/octo-issue-feed.yml` | `@main` (floating) | `@v1` | Same as above. |
| `.github/workflows/octo-pr-feed.yml` | `@c6c51348fccac99161166dce1b52dec1063aeea2` (SHA) | `@v1` | Workflow content **does** differ at v1 vs the old SHA — the reusable workflow was refactored to externalize its Python notification logic into a sparse-checked-out script. Functionally equivalent (same Octo notification behavior); v1 also adds `permissions: contents: read` and `timeout-minutes: 5`. Worth noting but acceptable: this is exactly the kind of upgrade `@v1` is meant to deliver. |

Confirmed `refs/tags/v1` resolves to commit `1aab7b304de30a61f1b1e950c66c72bd558a97c7` on `Mininglamp-OSS/.github` and is an **annotated** tag (good — preserves authorship metadata and is harder to silently mutate than a lightweight tag). All four target workflow files exist at that ref.

## Findings

### P2 — Tag mutability vs SHA pinning (informational)

`@v1` is mutable: it can be moved with `git tag -f && git push -f --tags`. The previous mix had two workflows pinned to immutable commit SHAs (`auto-add-to-project.yml`, `octo-pr-feed.yml`); after this PR they ride the `v1` tag instead. This is a deliberate tradeoff (readability + ability to deliver patch-level fixes vs strict immutability) and is fine for org-owned reusable workflows. It would be wrong for third-party actions, but `Mininglamp-OSS/.github` is in-house. No change requested — just calling it out so future reviewers don't re-litigate.

If the org wants strict immutability later, the standard pattern is to keep `@v1` as the public surface but ALSO pin to the commit SHA in branch-protection-critical paths, or to enforce signed annotated tags only.

### P2 — `code-review` workflow uses `@main` (out of scope)

`.github/workflows/code-review.yml` (already on `main`, not touched by this PR) still references the reusable code-review workflow at `@main`. If the goal is "all reusable workflow refs pinned to `@v1`," this one was missed. Likely intentional (the code-review workflow may not be tagged into v1 yet, or may be evolving fast enough that pinning hurts), but worth confirming. **Not a blocker for this PR.**

## Unrelated CI noise (not introduced by this PR)

The `add-to-project / add-to-project` job is failing on this PR with `Input required and not supplied: github-token`. This same job failed on the immediately preceding PR (`chore/remove-pr-merged-done`, headSha `42cf5a6`) at 2026-05-15T11:15:50Z, **before** this PR existed. The failure is a pre-existing token-passing problem in the reusable `auto-add-to-project.yml` workflow itself, not something this PR introduced. Pinning the ref from a SHA whose blob is identical to v1 cannot have changed runtime behavior. Suggest filing a separate issue against `Mininglamp-OSS/.github` for that.

The `notify` and `Build & Type-check` checks pass.

## Verdict

Reference-only update, every replacement verified against the actual blob/ref at `Mininglamp-OSS/.github`, the new `@v1` tag exists, is annotated, and contains all four target workflows. The single content delta (`octo-pr-feed.yml`) is a refactor toward better hygiene, not a behavior change. Approve.

Jerry-Xin (changes requested):
Summary: Relevance gate passed; this is in-scope repository automation maintenance, but one workflow permission mismatch can break the PR feed after moving to `@v1`.

🔴 Blocking

🟡 Warning: [.github/workflows/octo-pr-feed.yml](/private/tmp/pr-review-VARiE8/repo/.github/workflows/octo-pr-feed.yml:7) and [.github/workflows/octo-pr-feed.yml](/private/tmp/pr-review-VARiE8/repo/.github/workflows/octo-pr-feed.yml:11) — the caller keeps `permissions: {}` while switching the reusable workflow to `@v1`. The `@v1` `octo-pr-feed` workflow now checks out `Mininglamp-OSS/.github` scripts and declares `contents: read`, but reusable workflows cannot elevate the `GITHUB_TOKEN` permissions passed by the caller; GitHub documents that those permissions can only be downgraded. This means the checkout step in the called workflow may run without the required contents permission and fail. Add `permissions: contents: read` to the `notify` job in this caller, or otherwise pass the minimum required token permission. ([docs.github.com](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations))

💬 Non-blocking

None.

✅ Highlights

The other reusable workflow references resolve to the existing `v1` tag, and the caller inputs/secrets match the called workflow contracts for `auto-add-to-project`, `issue-welcome`, and `octo-issue-feed`.

lml2468 (commented):
# Code Review — PR #21 (齐静春 📖)

Reviewed at head `2f3081f`. Four workflow files, ref-only changes `@main`/SHA → `@v1`.

## Intended Verdict: CHANGES_REQUESTED — one confirmed P1.

_(Cannot submit APPROVED/CHANGES_REQUESTED — author is lml2468, same as gh CLI auth.)_

---

## Confirmed blocker: `octo-pr-feed.yml` permissions mismatch (agrees with Jerry-Xin)

I independently verified this by diffing the old and new reusable workflows:

**Old version** (SHA `c6c5134`): inline Python script, no `actions/checkout`, no permissions needed → `permissions: {}` was correct.

**New version** (`@v1`): refactored to `actions/checkout` of `Mininglamp-OSS/.github` + external `scripts/octo_pr_feed_notify.py`. Declares `permissions: contents: read`.

The caller's `permissions: {}` restricts GITHUB_TOKEN to zero permissions. While `Mininglamp-OSS/.github` is public (so unauthenticated clone might work), relying on this is fragile — `actions/checkout` v4 uses the token by default, and the comment "GITHUB_TOKEN not used" is now misleading since v1 does use checkout.

**Fix**: add `contents: read` at job level (matching `octo-issue-feed.yml` which has no restrictions and works fine):

```yaml
permissions: {}  # ← delete or change to:
# permissions:
#   contents: read  # needed by v1 reusable for script checkout

jobs:
  notify:
    permissions:
      contents: read   # ← OR grant here at job level
```

Alternatively, update the comment to explain why `permissions: {}` is intentionally kept (if public-repo unauthenticated checkout is the deliberate design choice).

## Other callers — verified ✅

| Caller | permissions | Reusable @v1 | Match? |
|---|---|---|---|
| `auto-add-to-project.yml` | `{}` | `{}` (explicit "do NOT checkout") | ✅ |
| `issue-welcome.yml` | job: `issues: write` | job: `issues: write` | ✅ |
| `octo-issue-feed.yml` | (none → defaults) | `contents: read` + checkout | ✅ defaults include `contents: read` |
| `octo-pr-feed.yml` | `{}` | `contents: read` + checkout | ❌ mismatch |

## General observations

- ✅ Moving from mixed `@main`/SHA to uniform `@v1` is the right call — version tags are predictable and the org repo now has a stable baseline.
- ✅ `add-to-project` CI failure is pre-existing (same as PR #20), unrelated to this change.
- ✅ `Build & Type-check` green — no code changes, just workflow refs.

One-line fix → ship it.

PR #22, opened 2026-05-15
fix(workflows): grant contents:read to octo-pr-feed caller

lml2468:
## Summary

Follow-up fix for PR #21 review finding: the `octo-pr-feed.yml` caller
workflow had `permissions: {}` which conflicts with the `@v1` reusable
workflow's `actions/checkout` step.

## Root Cause

The `@v1` reusable workflow was refactored from inline Python (no checkout
needed) to `actions/checkout` + external `scripts/octo_pr_feed_notify.py`.
The caller's `permissions: {}` was correct for the old version but starves
the new version of `contents:read` needed for the checkout.

## Fix

Replace `permissions: {}` with `permissions: { contents: read }` and update
the comment to explain why.

## Related

- PR #21 (merged with this gap)
- Review findings: 齐静春 + Jerry-Xin independently identified the mismatch

### DIFF
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -4,7 +4,12 @@ on:
   pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
 
-permissions: {}  # GITHUB_TOKEN not used; lock down to minimum
+# Minimum permissions: the @v1 reusable workflow checks out
+# Mininglamp-OSS/.github scripts via actions/checkout, which needs
+# contents:read. (The prior SHA-pinned version used inline Python
+# and needed no token at all — see PR #21 review.)
+permissions:
+  contents: read
 
 jobs:
   notify:


Jerry-Xin (approved):
APPROVED — verified against the reusable workflow.

The @v1 reusable (`octo-pr-feed.yml`) declares `permissions: contents: read` and uses `actions/checkout` to pull `.github` scripts. Caller must grant at least the same scope for token inheritance to work reliably. Fix is minimal and correctly scoped.

Comment explaining the history (inline Python → checkout + external script) is a nice touch for future readers.

`add-to-project` CI failure is the known pre-existing issue (unrelated). 🚀

Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-adapters`: it fixes a repository GitHub Actions caller workflow used by the project.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

🔵 Suggestion / Validation: The change in `.github/workflows/octo-pr-feed.yml:11` correctly grants only `contents: read`, which is the minimum repository permission needed for `actions/checkout` in the reusable workflow while keeping the `pull_request_target` token constrained.

🔵 Suggestion / Validation: The workflow still passes only event metadata and `OCTO_BOT_TOKEN` to the pinned reusable workflow at `.github/workflows/octo-pr-feed.yml:16`, so this change does not expand write permissions or introduce an obvious new PR-code execution path.

yujiawei (approved):
# Code Review — PR #22 (octo-adapters)

## Verdict: APPROVE

This is a correct, minimal follow-up fix for the gap surfaced in PR #21 review. Granting `contents: read` to the caller workflow is required for the `@v1` reusable workflow to function, and the change does not weaken the security posture of `pull_request_target`.

## Why the fix is correct

The reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@v1` (currently at `0bc479c4ebf6a1c28646b00cf23fb263356e7933`) checks out scripts via `actions/checkout`:

```yaml
- name: Checkout reusable-workflow scripts
  uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
  with:
    repository: Mininglamp-OSS/.github
    ref: ${{ job.workflow_sha }}
    persist-credentials: false
```

For reusable workflows invoked via `uses:`, the effective `GITHUB_TOKEN` permissions are the **intersection** of the caller's `permissions` block and the called workflow's `permissions` block. The reusable workflow declares `permissions: contents: read`, but with the previous caller block of `permissions: {}` the intersection was empty, leaving `actions/checkout` without `contents:read` and starving the job of the token scope it needs to fetch `Mininglamp-OSS/.github`.

After this PR the caller declares `permissions: contents: read`, so the intersection is `contents: read` — the exact minimum the reusable workflow asks for. No other scopes are granted.

## Security review

- `pull_request_target` runs in the base-ref context with access to secrets, so over-permissioning here would be dangerous. This PR keeps the token strictly read-only and limited to `contents` — no `pull-requests`, no `issues`, no `actions`, no `id-token`. Good hygiene.
- The PR does not introduce a checkout of untrusted PR head code. Only the trusted `Mininglamp-OSS/.github` repo at `job.workflow_sha` is checked out, and `persist-credentials: false` prevents the token from being written to `.git/config` for the subsequent script execution.
- `OCTO_BOT_TOKEN` is forwarded as a secret to the reusable workflow, unchanged. The reusable workflow uses it only as an env var for the notify script (not via git auth), so the new `contents:read` grant on `GITHUB_TOKEN` does not increase the blast radius of `OCTO_BOT_TOKEN`.
- The comment block explaining the why (and pointing at PR #21) is helpful for future maintainers and matches the actual cause.

## Notes (not blocking, not introduced by this PR)

- The reusable workflow is referenced by mutable tag `@v1` rather than a commit SHA. Anyone with write access to `Mininglamp-OSS/.github` can retarget `v1`, and `pull_request_target` makes any compromise of that tag impactful (secrets + write-capable token environment from the perspective of any future scope expansion). This predates PR #22 and is out of scope here, but worth tracking — consider SHA-pinning `@v1` once it stabilises, or restricting who can move org-level workflow tags.
- `actions/checkout` inside the reusable workflow uses `ref: ${{ job.workflow_sha }}`, which keeps script-and-workflow lockstep — that part is solid; the `@v1` mutability concern is purely about the initial resolution of the reusable workflow itself.

## Diff scope sanity

- `.github/workflows/octo-pr-feed.yml` — only file touched (+6 / −1).
- No changes to triggers, jobs, secrets, or inputs beyond the `permissions` block and its preceding comment.
- Caller workflow body is otherwise unchanged from PR #21's merged state.

LGTM — safe to merge.

PR #17, opened 2026-05-15
feat(octo): rename openclaw-channel-dmwork → openclaw-channel-octo with auto-migration

caster-Q:
## Summary

Product rebrand of the npm package `openclaw-channel-dmwork` to
`openclaw-channel-octo`, including OpenClaw plugin id, channel id, and
all user-visible labels. Existing dmwork users get a one-command
automatic migration; existing scripts using the old package name keep
working through a thin shim package.

Delivered in two phases (squash-merging is fine — both commits are
self-contained):

- **Phase A** (`8dca5e4`): copy `openclaw-channel-dmwork/` →
  `openclaw-channel-octo/` and rename everything user-visible. The old
  package is left untouched on disk to keep the dmwork v0.6.x branch
  shippable.
- **Phase B** (`a1733ae`): automatic rebrand migration on `install`,
  shim package `openclaw-channel-dmwork` (forwarder), and CI.

## Related Issue

Internal product rebrand initiative (no public issue).

### Phase A — new package, manual coexistence

- New directory `openclaw-channel-octo/` with package name + plugin id +
  channel id renamed; `bin/octo.js`, `skills/octo-bot-api/`, etc.
- `src/constants.ts`: centralised constants (`PLUGIN_ID`, `CHANNEL_ID`,
  `LEGACY_*`) + helpers (`stripChannelPrefix`, `getChannelConfig`,
  `ensureChannelConfigObject`).
- Backwards compatibility (one release cycle): `/dmwork_*` slash command
  aliases, `dmwork_management` agent tool alias, dual-prefix parsing
  (`octo:` / `dmwork:`) in target strings, sessionKeys, and
  `GROUP.md` regex.
- All user-visible labels, log prefixes, and runtime sessionKey prefixes
  switched to `octo`.

### Phase B — automatic migration + shim + CI

- **Migration** (`cli/install.ts`): `runMigration(legacyPluginId)` with
  variants `runRebrandMigration` (openclaw-channel-dmwork → octo) and
  `runLegacyToOctoMigration` (very-legacy `dmwork` plugin id → octo).
  Command-driven (`pluginsDisable` / `pluginsInstall` / `pluginsEnable`
  / `pluginsUninstall`) with 4.20 fallback to direct
  `cfg.plugins.entries` edits where commands don't exist.
  Transactional: backs up `openclaw.json`; rolls back on
  install/verify failure (cfg restore + re-enable legacy if it was
  enabled). Verifies `isHealthyInstall(octo)`, `channels.octo` presence
[...]

PR #1, opened 2026-05-11
chore: repo cleanup — security email, remove legacy dirs, update configs

Jerry-Xin:
## Summary

Repo-wide cleanup addressing 6 identified issues:

### Changes

1. **SECURITY.md** — Email updated to `security@mininglamp.com`, supported version updated to `v0.6.3`
2. **Remove legacy directories** — `claude-code-dmwork-ws/` and `hermes-channel-dmwork/` fully removed
3. **FUNDING.yml** — GitHub org updated to `Mininglamp-OSS`
4. **TASK.md** — Deleted
5. **.claude/** — Directory removed, added to `.gitignore`
6. **dependabot.yml** — Fixed ecosystem from `gomod` to `npm`, directory to `/openclaw-channel-dmwork`

### Additional fixes (from review feedback)

- **ISSUE_TEMPLATE/config.yml** — Security link updated to this repo with `security@mininglamp.com`
- **CONTRIBUTING.md** — Added Security section with correct email
- **README.md** — Removed stale `claude-code-dmwork-ws` reference
- **.gitignore** — Added `.claude/` and `TASK.md` ignore rules

### Verification

- Zero grep hits for deleted directory names in the tree
- All cross-references verified clean
- No functional code changes to `openclaw-channel-dmwork`

PR #24, opened 2026-05-15
fix(ci): extend coverage to both packages, add CI status, unify Node 22, pin SHAs

lml2468:
### 🟠 H-3: Extend CI to cover both packages
Previously only `openclaw-channel-dmwork` was tested in CI. Switch to a `matrix` strategy covering both `openclaw-channel-dmwork` and `openclaw-channel-octo`.

### 🟠 H-4: Replace `npm install` with `npm ci`
`npm install` may modify the lockfile; `npm ci` is deterministic and matches how publish workflows install dependencies.

### 🟠 H-5: Add octo-ci-status.yml
Notify Octo IM `📦 octo-adapters` channel when CI status changes on main.

### 🟡 M-7: Unify Node.js version to 22
CI was using Node 20 while publish workflows use Node 22. Align to avoid environment drift.

### 🟢 L-5: Add type-check to publish_shim job
Even a thin shim deserves a type-check before publishing.

### 🟢 L-1: Pin action SHAs
`actions/checkout@v4` → `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4`
`actions/setup-node@v4` → `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         include:
           - package: openclaw-channel-dmwork
-            install_cmd: npm install
+            install_cmd: npm ci
           - package: openclaw-channel-octo
             install_cmd: npm ci
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,20 +15,23 @@ concurrency:
 
 jobs:
   build:
-    name: Build & Type-check
+    name: Build & Type-check (${{ matrix.package }})
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package: [openclaw-channel-dmwork, openclaw-channel-octo]
     defaults:
       run:
-        working-directory: openclaw-channel-dmwork
+        working-directory: ${{ matrix.package }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: '20'
+          node-version: '22'
 
       - name: Install dependencies
-        run: npm install
+        run: npm ci
 
       - name: Type-check
         run: npm run type-check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        package: [openclaw-channel-dmwork, openclaw-channel-octo]
+        include:
+          - package: openclaw-channel-dmwork
+            install_cmd: npm install
+          - package: openclaw-channel-octo
+            install_cmd: npm ci
     defaults:
       run:
         working-directory: ${{ matrix.package }}
@@ -31,7 +35,7 @@ jobs:
           node-version: '22'
 
       - name: Install dependencies
-        run: npm ci
+        run: ${{ matrix.install_cmd }}
 
       - name: Type-check
         run: npm run type-check
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -6,6 +6,8 @@ on:
     types: [completed]
     branches: [main]
 
+permissions: {}
+
 jobs:
   notify:
     uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -1,0 +1,21 @@
+name: Octo CI Status
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
+    with:
+      repo_name: ${{ github.event.workflow_run.repository.name }}
+      workflow_name: ${{ github.event.workflow_run.name }}
+      conclusion: ${{ github.event.workflow_run.conclusion }}
+      run_id: ${{ github.event.workflow_run.id }}
+      run_url: ${{ github.event.workflow_run.html_url }}
+      project_group_id: "0262cefe4463468bb6f2accaeda623cf"
+      workflow_id: ${{ github.event.workflow_run.workflow_id }}
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -4,6 +4,8 @@ on:
   issues:
     types: [opened, closed, reopened, labeled]
 
+permissions: {}
+
 jobs:
   notify:
     uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
--- a/.github/workflows/publish-octo-dev.yml
+++ b/.github/workflows/publish-octo-dev.yml
@@ -46,8 +46,8 @@ jobs:
       run:
         working-directory: ${{ env.PKG_DIR }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
       - run: npm ci --ignore-scripts
@@ -63,8 +63,8 @@ jobs:
       run:
         working-directory: ${{ env.PKG_DIR }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
           registry-url: "https://registry.npmjs.org"
--- a/.github/workflows/publish-octo.yml
+++ b/.github/workflows/publish-octo.yml
@@ -59,8 +59,8 @@ jobs:
       run:
         working-directory: ${{ env.PKG_DIR }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
       - name: npm ci
@@ -80,8 +80,8 @@ jobs:
       run:
         working-directory: ${{ env.PKG_DIR }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
           registry-url: "https://registry.npmjs.org"
@@ -133,8 +133,8 @@ jobs:
     needs: publish_octo
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
           registry-url: "https://registry.npmjs.org"
@@ -160,6 +160,9 @@ jobs:
             echo "DRY_RUN=$DRY"
           } >> "$GITHUB_OUTPUT"
           echo "[publish_shim] version=$VERSION dist-tag=$DIST_TAG dry_run='$DRY'"
+      - name: type-check
+        working-directory: ${{ env.PKG_DIR }}
+        run: npm ci --ignore-scripts && npm run type-check
       - name: build shim
         env:
           CI_COMMIT_TAG: v${{ steps.version.outputs.VERSION }}
--- a/.github/workflows/publish-octo.yml
+++ b/.github/workflows/publish-octo.yml
@@ -161,7 +161,7 @@ jobs:
           } >> "$GITHUB_OUTPUT"
           echo "[publish_shim] version=$VERSION dist-tag=$DIST_TAG dry_run='$DRY'"
       - name: type-check
-        working-directory: ${{ env.PKG_DIR }}
+        working-directory: openclaw-channel-dmwork
         run: npm ci --ignore-scripts && npm run type-check
       - name: build shim
         env:
--- a/.github/workflows/publish-octo.yml
+++ b/.github/workflows/publish-octo.yml
@@ -160,9 +160,6 @@ jobs:
             echo "DRY_RUN=$DRY"
           } >> "$GITHUB_OUTPUT"
           echo "[publish_shim] version=$VERSION dist-tag=$DIST_TAG dry_run='$DRY'"
-      - name: type-check
-        working-directory: openclaw-channel-dmwork
-        run: npm ci --ignore-scripts && npm run type-check
       - name: build shim
         env:
           CI_COMMIT_TAG: v${{ steps.version.outputs.VERSION }}
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ node_modules/
 package-lock.json
 # Track lockfile for openclaw-channel-octo so publish workflows' `npm ci` works.
 !openclaw-channel-octo/package-lock.json
+# Track lockfile for openclaw-channel-dmwork so publish workflows' `npm ci` works.
+!openclaw-channel-dmwork/package-lock.json
 dist/
 src/**/*.js
 src/**/*.js.map
--- a/openclaw-channel-dmwork/package-lock.json
+++ b/openclaw-channel-dmwork/package-lock.json
@@ -1,0 +1,9121 @@
+{
+  "name": "openclaw-channel-dmwork",
+  "version": "0.6.3",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "openclaw-channel-dmwork",
+      "version": "0.6.3",
+      "bundleDependencies": [
+        "cos-nodejs-sdk-v5",
+        "crypto-js",
+        "curve25519-js",
+        "md5-typescript"
+      ],
+      "dependencies": {
+        "axios": "^1.7.0",
+        "commander": "^12.1.0",
+        "cos-nodejs-sdk-v5": "^2.15.4",
+        "crypto-js": "^4.2.0",
+        "curve25519-js": "^0.0.4",
+        "md5-typescript": "^1.0.5",
+        "ws": "^8.16.0"
+      },
+      "bin": {
+        "openclaw-channel-dmwork": "bin/dmwork.js"
+      },
+      "devDependencies": {
+        "@types/crypto-js": "^4.2.0",
+        "@types/ws": "^8.5.10",
+        "openclaw": "^2026.5.4",
+        "typescript": "^5.9.3",
+        "vitest": "^3.0.0"
+      },
+      "peerDependencies": {
+        "openclaw": ">=2026.5.4"
+      }
+    },
+    "node_modules/@agentclientprotocol/sdk": {
+      "version": "0.21.0",
+      "resolved": "https://registry.npmjs.org/@agentclientprotocol/sdk/-/sdk-0.21.0.tgz",
+      "integrity": "sha512-ONj+Q8qOdNQp5XbH5jnMwzT9IKZJsSN0p0lkceS4GtUtNOPVLpNzSS8gqQdGMKfBvA0ESbkL8BTaSN1Rc9miEw==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "peerDependencies": {
+        "zod": "^3.25.0 || ^4.0.0"
+      }
+    },
+    "node_modules/@anthropic-ai/sdk": {
+      "version": "0.91.1",
+      "resolved": "https://registry.npmjs.org/@anthropic-ai/sdk/-/sdk-0.91.1.tgz",
+      "integrity": "sha512-LAmu761tSN9r66ixvmciswUj/ZC+1Q4iAfpedTfSVLeswRwnY3n2Nb6Tsk+cLPP28aLOPWeMgIuTuCcMC6W/iw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "json-schema-to-ts": "^3.1.1"
+      },
+      "bin": {
+        "anthropic-ai-sdk": "bin/cli"
+      },
+      "peerDependencies": {
+        "zod": "^3.25.0 || ^4.0.0"
+      },
+      "peerDependenciesMeta": {
+        "zod": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@aws-crypto/crc32": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-5.2.0.tgz",
+      "integrity": "sha512-nLbCWqQNgUiwwtFsen1AdzAtvuLRsQS8rYgMuxCrdKf9kOssamGLuPwyTY9wyYblNr9+1XM8v6zoDTPPSIeANg==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=16.0.0"
+      }
+    },
+    "node_modules/@aws-crypto/sha256-browser": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-5.2.0.tgz",
+      "integrity": "sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-crypto/sha256-js": "^5.2.0",
+        "@aws-crypto/supports-web-crypto": "^5.2.0",
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "@aws-sdk/util-locate-window": "^3.0.0",
+        "@smithy/util-utf8": "^2.0.0",
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@aws-crypto/sha256-js": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-js/-/sha256-js-5.2.0.tgz",
+      "integrity": "sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=16.0.0"
+      }
+    },
+    "node_modules/@aws-crypto/supports-web-crypto": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/supports-web-crypto/-/supports-web-crypto-5.2.0.tgz",
+      "integrity": "sha512-iAvUotm021kM33eCdNfwIN//F77/IADDSs58i+MDaOqFrVjZo9bAal0NK7HurRuWLLpF1iLX7gbWrjHjeo+YFg==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@aws-crypto/util": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/util/-/util-5.2.0.tgz",
+      "integrity": "sha512-4RkU9EsI6ZpBve5fseQlGNUWKMa1RLPQ1dnjnQoe07ldfIzcsGb5hC5W0Dm7u423KWzawlrpbjXBrXCEv9zazQ==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-sdk/types": "^3.222.0",
+        "@smithy/util-utf8": "^2.0.0",
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@aws-sdk/client-bedrock-runtime": {
+      "version": "3.1048.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/client-bedrock-runtime/-/client-bedrock-runtime-3.1048.0.tgz",
+      "integrity": "sha512-u+NT61JZEkRFtpL0CAw1N1dwxnaLgwVXQl/zjJxTGgLyS/jTIdg2SdoEoCTHxgDyCnqa1HEi9QOoE9/pYRNpOQ==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-crypto/sha256-browser": "5.2.0",
+        "@aws-crypto/sha256-js": "5.2.0",
+        "@aws-sdk/core": "^3.974.11",
+        "@aws-sdk/credential-provider-node": "^3.972.42",
+        "@aws-sdk/eventstream-handler-node": "^3.972.16",
+        "@aws-sdk/middleware-eventstream": "^3.972.12",
+        "@aws-sdk/middleware-websocket": "^3.972.19",
+        "@aws-sdk/token-providers": "3.1048.0",
+        "@aws-sdk/types": "^3.973.8",
+        "@smithy/core": "^3.24.2",
+        "@smithy/fetch-http-handler": "^5.4.2",
+        "@smithy/node-http-handler": "^4.7.2",
+        "@smithy/types": "^4.14.1",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=20.0.0"
+      }
+    },
+    "node_modules/@aws-sdk/core": {
+      "version": "3.974.11",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/core/-/core-3.974.11.tgz",
+      "integrity": "sha512-QpnINq5FZH6EOaDEkmHdT7eUunbvD27pDNQypaWjFyYz7Zl1q3UCMQErBZxpmfGfI7MvI2TlK8KTkgNpv8b1ug==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-sdk/types": "^3.973.8",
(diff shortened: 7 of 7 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `octo-adapters`, but the new CI workflow currently breaks one matrix leg.

🔴 Blocking

- 🟡 Warning: `.github/workflows/ci.yml:33`-`.github/workflows/ci.yml:34` switches every matrix package to `npm ci`, but `openclaw-channel-dmwork` has no `package-lock.json` or `npm-shrinkwrap.json`. I verified locally that `npm ci` in `openclaw-channel-dmwork` exits with `EUSAGE: The npm ci command can only install with an existing package-lock.json`. This means the new `openclaw-channel-dmwork` CI leg will fail before type-check/build/test. Add a lockfile for `openclaw-channel-dmwork`, or keep a package-specific install command until that package is lockfile-backed.

- 🟡 Warning: `.github/workflows/octo-ci-status.yml:11` introduces a reusable workflow dependency pinned to mutable `@v1` while passing `OCTO_BOT_TOKEN` at `.github/workflows/octo-ci-status.yml:20`-`.github/workflows/octo-ci-status.yml:21`. Since this PR’s stated hardening goal includes SHA pinning, this new secret-consuming workflow should also be pinned to an immutable commit SHA, or the exception should be explicitly justified.

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/octo-ci-status.yml` does not declare explicit `permissions`. If the reusable workflow does not need repository token privileges from the caller, set the minimum permissions explicitly for least privilege.

✅ Highlights

- The CI matrix direction is correct for this repo: both `openclaw-channel-dmwork` and `openclaw-channel-octo` are first-class packages.
- Aligning CI and publish workflows on Node 22 reduces environment drift.
- Adding a type-check before publishing the shim is a useful guard.

yujiawei (dismissed):
# Code Review — PR #24 (octo-adapters)

Thanks for the cleanup — the matrix, SHA pinning, and Node-version unification all move CI in the right direction. However, one change is a hard blocker as written, and a couple of smaller items are worth tightening before merge.

## Verdict summary

| Item | Status |
| --- | --- |
| H-3 Matrix covers both packages | ⚠️ Breaks dmwork run (see P0) |
| H-4 `npm install` → `npm ci` | ❌ Fails for `openclaw-channel-dmwork` (no lockfile) |
| H-5 `octo-ci-status.yml` | ✅ Reusable workflow exists, triggers wired correctly |
| M-7 Unify Node 22 | ✅ Aligned with publish workflows |
| L-5 `type-check` in `publish_shim` | ⚠️ Redundant + does not actually type-check the shim |
| L-1 Pin action SHAs | ✅ SHAs match upstream `v4` tags |

---

## P0 — `npm ci` will fail for `openclaw-channel-dmwork`

**File:** `.github/workflows/ci.yml` (lines 19-34)

The matrix runs the same install step against both packages:

```yaml
strategy:
  matrix:
    package: [openclaw-channel-dmwork, openclaw-channel-octo]
defaults:
  run:
    working-directory: ${{ matrix.package }}
...
- name: Install dependencies
  run: npm ci
```

But `openclaw-channel-dmwork/` has **no `package-lock.json`** — neither at the PR head (`6e0dc26`) nor on `main`. Verified:

```
$ git ls-files openclaw-channel-dmwork | grep -i lock
(no output)
$ ls openclaw-channel-dmwork/  # only package.json, no lockfile
```

`npm ci` requires an existing `package-lock.json` (or `npm-shrinkwrap.json`) and will exit with `EUSAGE The npm ci command can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync...`. The dmwork matrix leg will fail on the first run and CI will go red on `main`.

The PR description frames H-4 as "matches how publish workflows install dependencies" — that is true for `openclaw-channel-octo` (which has a lockfile and which the publish workflow already uses `npm ci --ignore-scripts` against), but it is not currently true for the dmwork package.

**Recommended fixes (pick one):**

1. **Preferred:** generate and commit `openclaw-channel-dmwork/package-lock.json` in this PR (`cd openclaw-channel-dmwork && npm install`). This is consistent with the H-4 rationale and unblocks reproducible installs.
2. Keep `npm ci` for octo and use `npm install --no-audit --no-fund` for dmwork until a lockfile lands — e.g. by splitting the matrix step or adding `if: matrix.package == 'openclaw-channel-octo'` / `else` branches. This reintroduces non-determinism on the dmwork side, so option 1 is better.
3. Drop dmwork from the matrix in this PR and re-add it in a follow-up that also commits the lockfile. Loses H-3 but keeps the rest of the PR landable.

Whichever path you take, please verify the workflow with a green run on the PR before merging — the current PR's CI will mask this because it runs the new workflow against itself, so a red signal on the dmwork leg is the expected pre-merge state, not a flake.

---

## P2 — `type-check` step in `publish_shim` does not check the shim

**File:** `.github/workflows/publish-octo.yml` (lines 162-165)

```yaml
- name: type-check
  working-directory: ${{ env.PKG_DIR }}     # PKG_DIR = openclaw-channel-octo
  run: npm ci --ignore-scripts && npm run type-check
```

Two issues:

1. **It type-checks `openclaw-channel-octo`, not the shim.** The shim source lives in `openclaw-channel-octo/shim-stub/` and is assembled into `/tmp/shim-build` only by the next step (`build shim`). At the time this step runs, there is nothing shim-specific to type-check that wasn't already type-checked.
2. **It duplicates work already done in the `test` job.** The same workflow's `test` job (lines 64-77) runs `npm ci --ignore-scripts` and `npx tsc --noEmit` against the same directory, and `publish_shim` only runs after `publish_octo` which `needs: test`. So type-check has already passed before this step ever fires.

Net: the step adds ~30s of runtime and a redundant `npm ci` without strengthening the gate. Suggested options:

- Drop the step (the `test` job already protects the publish path).
- Or, if the intent is genuinely to gate the shim, run a `tsc --noEmit` against the *built* shim after `build shim` (i.e. against `/tmp/shim-build`). That requires a tsconfig in the shim output, which doesn't currently exist — so this would be a follow-up.

The PR description ("Even a thin shim deserves a type-check before publishing") suggests the second intent; if so, this step doesn't satisfy it.

---

## P2 — `setup-node` cache hint (nit, optional)

For both `ci.yml` and `publish-octo.yml`, consider adding `cache: 'npm'` to the `setup-node` invocations (with `cache-dependency-path` pointing at the relevant lockfile). With `npm ci`, the node_modules install is the dominant cost; node-cache typically halves it. Not blocking — flag for future.

---

## Verified items

- **Pinned SHAs (L-1):** `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` and `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020` correspond to the published `v4` tags. Comment annotation `# v4` is the right convention for renovate/dependabot to recognize.
- **`octo-ci-status.yml` (H-5):**
  - Trigger `workflows: ["CI"]` matches the existing CI workflow's `name: CI` (line 1 of `ci.yml`). ✅
  - Branch filter `main` is correct.
  - Reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` exists upstream (verified via API).
  - `OCTO_BOT_TOKEN` is forwarded as a secret rather than an input — correct.
- **Concurrency:** the existing `concurrency` group at workflow level is matrix-safe — all matrix legs share one workflow run, so cancellation behavior is unchanged.
- **Node 22 (M-7):** matches the `node-version: 22` already used in every job of `publish-octo.yml`. Drift removed.

---

## Suggested order of operations

1. Land a small precursor PR that adds `openclaw-channel-dmwork/package-lock.json`.
2. Rebase this PR on top; CI matrix will then pass.
3. Optionally drop the redundant `type-check` step in `publish_shim` (or convert it into a real shim type-check in a follow-up).

Happy to re-review once the lockfile is in place.

lml2468 (commented):
# Code Review — PR #24 (齐静春 📖)

Reviewed at head `6e0dc26`. 3 files: ci.yml (matrix + sha pins + node 22 + npm ci), octo-ci-status.yml (new), publish-octo.yml (sha pins + shim type-check).

## Intended Verdict: CHANGES_REQUESTED — one confirmed blocker.

_(Cannot submit CHANGES_REQUESTED — author is lml2468.)_

---

## 🔴 Blocker: `npm ci` on dmwork without lockfile (agrees with Jerry-Xin)

CI confirms: `Build & Type-check (openclaw-channel-dmwork)` → **FAILURE**. `openclaw-channel-dmwork/` has no `package-lock.json`. `npm ci` exits `EUSAGE`.

Three fix options, in my recommended order:

1. **Drop dmwork from matrix** — it's in maintenance mode (0.6.x branch patches only, per PR #17). The canonical package is octo. CI for dmwork patches runs on its own GitLab pipeline.
2. **Conditional install** — `npm ci` for octo, `npm install` for dmwork (ugly but works).
3. **Commit a lockfile** — invests in a legacy package being superseded.

I'd go with (1). The dmwork package is explicitly kept for backwards compat, not active development.

## P2 — `publish-octo-dev.yml` not pinned (inconsistency)

PR description says "Pin action SHAs" and `ci.yml` + `publish-octo.yml` are pinned, but `publish-octo-dev.yml` still uses `@v4` tags:

```
publish-octo-dev.yml:49: actions/checkout@v4
publish-octo-dev.yml:50: actions/setup-node@v4
publish-octo-dev.yml:66: actions/checkout@v4
publish-octo-dev.yml:67: actions/setup-node@v4
```

Either pin all or document why dev is excluded. Inconsistent pinning creates a false sense of supply-chain protection.

## P3 — `publish_shim` type-check is redundant

The new step (`publish-octo.yml:162-164`):
```yaml
- name: type-check
  working-directory: ${{ env.PKG_DIR }}
  run: npm ci --ignore-scripts && npm run type-check
```

This type-checks `openclaw-channel-octo/` (the canonical package), which is already done in the `test` job. The shim itself has no TypeScript — it's `bin/dmwork.js` + `package.json`. The step doesn't validate the shim, it re-validates the canonical package.

Not harmful, but the step name in a `publish_shim` job is misleading. Either:
- Remove it (already covered by `test` job which `publish_octo` depends on)
- Rename to "install canonical package (dependency of shim build)"

## Verified ✅

- **SHA pins correct**: `actions/checkout@34e1148` and `actions/setup-node@49933ea` both match current `v4` tag HEAD.
- **Node 22 unification**: ci.yml `20 → 22`, matches publish workflows. Good.
- **`octo-ci-status.yml`**: caller triggers on `workflow_run` (CI on main), calls `@v1` reusable which needs `actions: read` + `contents: read`. No explicit permissions in caller → inherits defaults which include both. Works correctly. The reusable only fires on state transitions (success→failure / failure→success), not every run — good signal-to-noise.
- **`add-to-project` CI now green** 🎉 (previously failing)

## Summary

Fix the dmwork npm-ci blocker (recommend dropping from matrix) + pin `publish-octo-dev.yml` for consistency → ship.

Jerry-Xin (changes requested):
Summary: The PR is in scope for this repository, but the updated workflows will fail for `openclaw-channel-dmwork`.

🔴 Blocking

🔴 Critical — `.github/workflows/ci.yml:33-34` switches the matrix job to `npm ci`, but `openclaw-channel-dmwork` does not have a `package-lock.json`. I verified locally from `openclaw-channel-dmwork` that `npm ci` exits with `EUSAGE` because no lockfile exists, so the new `openclaw-channel-dmwork` CI matrix entry cannot pass. Add and commit a lockfile for `openclaw-channel-dmwork`, or keep that package on an install command that matches the repository’s current state.

🔴 Critical — `.github/workflows/publish-octo.yml:163-165` adds the same `npm ci --ignore-scripts` requirement to the `publish_shim` job against `openclaw-channel-dmwork`, which also lacks a lockfile. This means release publishing can fail after `publish_octo` succeeds but before publishing/deprecating the shim, leaving the release workflow partially completed.

🟡 Warning — `.github/workflows/ci.yml:23-25` uses `${{ matrix.package }}` inside `jobs.build.defaults.run.working-directory`. GitHub’s documentation says contexts and expressions cannot be used in `defaults.run`; move the working directory expression onto each `run` step or use another supported pattern. ([docs.github.com](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/set-default-values-for-jobs))

💬 Non-blocking

🔵 Suggestion — `.github/workflows/octo-ci-status.yml:11` uses the reusable workflow at `Mininglamp-OSS/.github/...@v1`. Since this job receives `OCTO_BOT_TOKEN`, consider pinning that reusable workflow to an immutable SHA as well, or document why the org-controlled mutable tag is trusted.

✅ Highlights

✅ Extending CI coverage to both `openclaw-channel-dmwork` and `openclaw-channel-octo` is the right direction.

✅ Aligning CI and publish jobs on Node 22 reduces environment drift.

✅ The pinned SHAs for `actions/checkout` and `actions/setup-node` improve supply-chain stability for the standard actions.

lml2468 (commented):
# Code Review — PR #24 (incremental re-review by 齐静春 📖)

Reviewed incremental diff `6e0dc26...46014ef`. Two files changed.

## Intended Verdict: CHANGES_REQUESTED — P0 still open + new issue introduced.

---

## My P2 — resolved ✅

**`publish-octo-dev.yml` SHA pins**: All 4 references now pinned to `34e1148` / `49933ea`, matching ci.yml and publish-octo.yml. Supply-chain posture now consistent.

## My P0 — still open ❌

**dmwork in CI matrix without lockfile**: ci.yml still has `matrix: [openclaw-channel-dmwork, openclaw-channel-octo]` with `npm ci`. CI confirms: dmwork leg still **FAILURE**, octo leg **CANCELLED**. This has not been addressed.

## 🔴 New issue: `publish_shim` type-check now points to wrong package

The `publish-octo.yml` change:
```yaml
- name: type-check
-   working-directory: ${{ env.PKG_DIR }}  # was: openclaw-channel-octo
+   working-directory: openclaw-channel-dmwork  # ← wrong
    run: npm ci --ignore-scripts && npm run type-check
```

This is broken in two ways:

1. **`npm ci` will fail** — `openclaw-channel-dmwork` has no `package-lock.json` (same issue as the CI matrix).
2. **Wrong package** — the shim is built from `shim-stub/` inside `openclaw-channel-octo`, and depends on the canonical octo package. Type-checking the dmwork package is irrelevant to the shim build.

I think my P3 feedback was misinterpreted. The original step was redundant (re-checking octo, already covered by the `test` job). The fix should be to **remove the step entirely**, not redirect it to dmwork:

```yaml
# DELETE this step — canonical package is already type-checked by the test job,
# and the shim has no TypeScript source.
- name: type-check
    working-directory: openclaw-channel-dmwork
    run: npm ci --ignore-scripts && npm run type-check
```

## Summary

Two items to fix:
1. Remove dmwork from CI matrix (or commit a lockfile — but it's legacy)
2. Remove the `publish_shim` type-check step (broken + wrong target)

yujiawei (dismissed):
# Code Review — PR #24 (octo-adapters)

**Verdict: Changes requested.** The PR's headline change — switching `npm install` → `npm ci` and extending the matrix to both packages — currently breaks CI on this very PR, and will also break the release path. Live evidence: workflow run [25927984254](https://github.com/Mininglamp-OSS/octo-adapters/actions/runs/25927984254) shows `Build & Type-check (openclaw-channel-dmwork)` failing at the install step with `npm error EUSAGE — The 'npm ci' command can only install with an existing package-lock.json`.

The intent of each change is good. Three of them ship cleanly; two need work before merge.

---

## P0 — Blocker: `npm ci` will fail in `openclaw-channel-dmwork` (no lockfile by design)

`.gitignore` (root) explicitly excludes lockfiles for every package except octo:

```gitignore
package-lock.json
# Track lockfile for openclaw-channel-octo so publish workflows' `npm ci` works.
!openclaw-channel-octo/package-lock.json
```

So `openclaw-channel-dmwork/package-lock.json` does not exist in the tree, and `npm ci` cannot run there. Two places in this PR are now broken because of that:

1. **`.github/workflows/ci.yml` line 33** — the matrix runs `npm ci` for every `package` value, including `openclaw-channel-dmwork`. Already failing on this PR's head (run 25927984254). Once merged, every push to `main` and every PR will see a red CI.
2. **`.github/workflows/publish-octo.yml` lines 163–165 (new `type-check` step in `publish_shim`)** —
   ```yaml
   - name: type-check
     working-directory: openclaw-channel-dmwork
     run: npm ci --ignore-scripts && npm run type-check
   ```
   This runs at release time. The first `npm version`-driven release that hits this path will fail at `npm ci` for the same reason and **block the shim from publishing**, even though the octo package itself was already published in the prior `publish_octo` job. That is a worse failure mode than the CI one — it splits a release halfway.

**Suggested fixes (pick one):**

- **Preferred:** keep `npm ci` (deterministic installs are the right call) and commit `openclaw-channel-dmwork/package-lock.json`, then update `.gitignore` to whitelist it the same way the octo lockfile is whitelisted (`!openclaw-channel-dmwork/package-lock.json`). The dmwork package has real dependencies (`vitest`, `tsc`, etc.), so a lockfile is the correct posture for both reproducible CI and supply-chain pinning.
- **Alternative (smaller scope):** if the team genuinely wants dmwork to stay lockfile-free, fall back to `npm install --no-package-lock` (or just `npm install`) for that matrix slot only, and similarly for the `publish_shim` type-check step. This preserves the current dmwork policy but loses the determinism `npm ci` was meant to bring — so it's strictly worse than committing the lockfile.

Either way, this must be resolved before merge — the current state is a guaranteed red CI.

---

## P1 — `fail-fast: true` (default) hides the octo result when dmwork fails

`.github/workflows/ci.yml` (matrix block) doesn't set `fail-fast`. With the default `true`, the moment dmwork dies, the octo job is cancelled (see check-runs API: octo conclusion = `cancelled` on this PR). For a CI matrix whose explicit purpose is to give independent visibility into both packages, `fail-fast: false` is the right default:

```yaml
strategy:
  fail-fast: false
  matrix:
    package: [openclaw-channel-dmwork, openclaw-channel-octo]
```

Not a blocker — but once the P0 above is fixed, you'll want this so a regression in one package doesn't mask the state of the other.

---

## P2 — `actions/checkout@v4` and `actions/setup-node@v4` are entering the Node 20 deprecation window

The pinned SHAs check out (verified: `34e114876b…` = checkout v4.3.1, `49933ea528…` = setup-node v4.4.0 — both real release tags, ✅). However, the live job log shows:

> Node.js 20 actions are deprecated. … Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026.

`actions/checkout@v6` and `actions/setup-node@v6` already run on Node 24. Since this PR is explicitly about CI hardening / future-proofing, it would be reasonable to pin to v6 SHAs in the same change:

- `actions/checkout` v6.0.2 → `de0fac2e4500dabe0009e67214ff5f5447ce83dd`
- `actions/setup-node` v6.4.0 → `48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e`

Optional — staying on v4.x is fine for now, but the team will be back here within four months. Worth doing once.

---

## L — `Mininglamp-OSS/.github/.../octo-ci-status.yml@v1` is a mutable tag

`octo-ci-status.yml` line 11:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
```

The PR's stated goal (`L-1: Pin action SHAs`) was applied to public actions but not to the internal reusable workflow. `v1` is a mutable tag (today it points at `0bc479c4ebf6a1c28646b00cf23fb263356e7933`, but anyone with write access to that repo can re-point it). Pinning would be:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@0bc479c4ebf6a1c28646b00cf23fb263356e7933 # v1
```

Lower priority because it's an internal first-party repo, but keeping the policy consistent makes future audits simpler.

---

## What's good

- ✅ Matrix expansion to cover both packages — correct, and the right shape (independent jobs, independent working-directory).
- ✅ Node 20 → 22 alignment with the publish workflows — eliminates the environment-drift class of bugs.
- ✅ SHA pinning for `actions/checkout` and `actions/setup-node` — both SHAs verified against legitimate release tags (v4.3.1 / v4.4.0).
- ✅ New `octo-ci-status.yml` is correctly scoped (`branches: [main]` + `workflow_run` on `CI`), so it stays dormant on PRs and only fires post-merge — exactly the right trigger model for status notifications.
- ✅ Adding a type-check to `publish_shim` is a good defense-in-depth idea (even if its current implementation is broken — see P0).

---

## Required before merge

1. **P0**: fix the missing dmwork lockfile (preferred) or fall back from `npm ci` for that path. CI must be green on this PR's head.
2. **P1**: add `fail-fast: false` to the matrix.

Items P2 and L are nice-to-have and can ship in this PR or a follow-up — your call.

Jerry-Xin (changes requested):
This PR is relevant to `Mininglamp-OSS/octo-adapters`, but it introduces a blocking release workflow failure.

🔴 Blocking

- 🟡 Warning: `.github/workflows/publish-octo.yml:163-165` runs `npm ci --ignore-scripts` in `openclaw-channel-dmwork`, but that package has no `package-lock.json`. `npm ci` fails immediately with `EUSAGE: The npm ci command can only install with an existing package-lock.json`. This breaks the `publish_shim` job before the shim can build or publish. Fix by committing `openclaw-channel-dmwork/package-lock.json` and using `npm ci`, or keep this step on `npm install` until the lockfile exists.

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/ci.yml:23-26` still uses `npm install` for `openclaw-channel-dmwork`, so the PR does not fully satisfy the stated “replace npm install with npm ci” goal. If a lockfile is added for dmwork, update this matrix entry to `npm ci` as well.

✅ Highlights

- CI coverage is expanded to both adapter packages via the matrix in `.github/workflows/ci.yml`.
- Node.js is consistently aligned to version 22 across CI and publish workflows.
- The direct GitHub Actions references are pinned to commit SHAs, improving supply-chain stability.

lml2468 (commented):
# Code Review — PR #24 (incremental re-review #3 by 齐静春 📖)

Reviewed incremental diff `46014ef...6fa1475`.

## Intended Verdict: CHANGES_REQUESTED — one P0 remains.

---

## My P0 #1 (CI matrix) — resolved ✅

CI matrix now uses conditional `install_cmd`: `npm install` for dmwork, `npm ci` for octo. Both legs green. Works.

## My P0 #2 (publish_shim type-check) — still open ❌

`publish-octo.yml` `publish_shim` job still has:

```yaml
- name: type-check
  working-directory: openclaw-channel-dmwork
  run: npm ci --ignore-scripts && npm run type-check
```

This will fail at publish time:
1. `npm ci` — dmwork has no lockfile (same issue just fixed in CI by using `npm install`)
2. Wrong package — the shim is built from octo's `shim-stub/`, not from dmwork

The `test` job (which `publish_octo` depends on, and `publish_shim` depends on `publish_octo`) already type-checks the canonical package. This step adds no value for the shim build. **Delete it.**

Jerry-Xin (changes requested):
Third round. CI matrix fix is solid — `npm install` for dmwork / `npm ci` for octo, both legs green ✅

🔴 Blocking — `publish-octo.yml:163-165` still has:
```yaml
- name: type-check
  working-directory: openclaw-channel-dmwork
  run: npm ci --ignore-scripts && npm run type-check
```
This step will break `publish_shim` on every tag push: dmwork has no lockfile → `npm ci` exits EUSAGE. And it's redundant anyway — CI `build` job already covers type-check for both packages, and the shim itself has no TypeScript to check.

Fix: delete lines 163-165 entirely. That's it.

yujiawei (dismissed):
# Code Review — PR #24 (octo-adapters)

## Summary

The intent of this PR is solid — extending CI to both packages, adding a status notifier, unifying Node 22, and pinning action SHAs are all welcome hygiene improvements. However, one change introduces a **release-blocking regression in `publish-octo.yml`**, and the PR description doesn't fully match the diff. Requesting changes before merge.

---

## P0 — `publish_shim` type-check step will break every release

**File:** `.github/workflows/publish-octo.yml` (lines 163-165 of the PR diff)

```yaml
- name: type-check
  working-directory: openclaw-channel-dmwork
  run: npm ci --ignore-scripts && npm run type-check
```

`openclaw-channel-dmwork/` has **no `package-lock.json`** committed in this branch (verified at HEAD `6fa1475`). `npm ci` requires a lockfile and exits non-zero otherwise:

> The `npm ci` command can only install with an existing package-lock.json or npm-shrinkwrap.json…

Result: the next package-scoped tag push (`openclaw-channel-octo/v*`) will fail at the `publish_shim` job's new type-check step, blocking the npm publish of the shim. This is the same constraint that forced the matrix in `ci.yml` to use `npm install` for dmwork.

Two ways to fix:
1. **Match the CI matrix:** use `npm install --ignore-scripts && npm run type-check` here, OR
2. **Commit `openclaw-channel-dmwork/package-lock.json`** and then use `npm ci` consistently across CI + publish (this is the better long-term answer and removes the special case from `ci.yml`).

There's also a secondary concern even if the install works: the published shim is generated from `openclaw-channel-octo/shim-stub/` by `scripts/build-shim.mjs`, not from `openclaw-channel-dmwork/` source. So this step type-checks a directory whose code is **not** what gets published as the shim. If the goal is "type-check what we ship," this validates the wrong target. If the goal is "keep the legacy dmwork source from rotting," that's fine — but it's worth a comment in the workflow to make that intent explicit.

---

## P1 — PR description disagrees with the diff (H-4)

The body advertises:

> ### 🟠 H-4: Replace `npm install` with `npm ci`
> `npm install` may modify the lockfile; `npm ci` is deterministic …

But the actual `ci.yml` keeps `npm install` for `openclaw-channel-dmwork` and only uses `npm ci` for `openclaw-channel-octo`. The matrix encodes the difference via `install_cmd`. That contradicts the headline of H-4 and may mislead readers (and future maintainers grepping git log) into thinking dmwork is now lockfile-pinned.

Either:
- Update the PR body to state honestly that dmwork stays on `npm install` until a lockfile is added, or
- Commit a dmwork lockfile and actually deliver H-4 across both packages.

---

## P2 — `octo-ci-status.yml` reusable-workflow ref is a mutable tag

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
```

L-1's stated goal in this PR is to pin third-party actions to immutable SHAs. The new `octo-ci-status.yml` ignores that for an internal reusable workflow by using `@v1` (a mutable tag). Internal repos can be retagged as easily as external ones. Recommendation: pin by SHA and keep `# v1` as a trailing comment, matching the rest of the diff.

---

## P2 — `publish_shim` type-check duplicates a full install on every release

The new step runs an unconditional `npm ci --ignore-scripts && npm run type-check` inside `publish_shim` after `publish_octo` has already finished. That's a fresh checkout, fresh node setup, and a full install just to call `tsc --noEmit` on a directory that — per the comment above — isn't even what the shim publishes. Two cheaper alternatives:

- Move the type-check into the existing `test` job (which already does `npm ci` for octo) and rely on a matrix entry for dmwork there, OR
- Skip the type-check in publish entirely if CI on the merged main has already validated dmwork.

This is style, not correctness — but it adds ~30-60s to every release for marginal value.

---

## Minor / Nits

1. **`node-version` quoting inconsistency.** `ci.yml` uses `'22'` (quoted), the publish workflows use `22` (bare). Both work, but it's worth picking one to keep diffs of future bumps tidy.
2. **`octo-ci-status.yml` has no `concurrency:` group.** Two `workflow_run` events for `CI` arriving close together (e.g. force-push to main) will run the notifier in parallel and may post out-of-order Octo IM notifications. Consider:
   ```yaml
   concurrency:
     group: octo-ci-status-${{ github.event.workflow_run.head_branch }}
     cancel-in-progress: false
   ```
3. **`octo-ci-status.yml` only listens to `branches: [main]`.** That matches the H-5 description ("when CI status changes on main"), so this is fine — just confirming it's intentional that PR CI failures don't post to the channel.
4. **L-1 pinning is inconsistent on first principles.** SHA pins are great, but without something like Dependabot configured for `package-ecosystem: github-actions`, these pins will rot and you'll be on stale `actions/checkout` indefinitely. Consider adding a Dependabot config in a follow-up.

---

## What's good

- Matrix approach in `ci.yml` is clean and the per-package install command is the right escape hatch given the lockfile situation.
- SHA pinning across `publish-octo.yml` and `publish-octo-dev.yml` is consistent and the `# v4` trailing comments preserve human readability.
- New `octo-ci-status.yml` is small and focused.
- Node 22 unification removes a real source of CI/publish drift.

---

## Verdict

The CI-side changes are good to merge as-is, but the `publish_shim` type-check step in `publish-octo.yml` will break the next release. Please address P0 before merging — either flip to `npm install --ignore-scripts` or commit a dmwork lockfile. P1 (description vs. diff mismatch) should also be cleaned up since this PR will be referenced in CHANGELOG context. P2 items are nice-to-haves and can be deferred.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `octo-adapters`, but the release workflow now fails before publishing the shim package.

🔴 Blocking

🔴 Critical — `.github/workflows/publish-octo.yml:163-165`: the new `publish_shim` type-check runs `npm ci --ignore-scripts` in `openclaw-channel-dmwork`, but that package has no `package-lock.json` or `npm-shrinkwrap.json`. I confirmed the exact command exits with `EUSAGE: The npm ci command can only install with an existing package-lock.json`. This means every `publish_shim` run will fail after `publish_octo` succeeds, leaving the canonical package published but the shim unpublished/deprecated. Use `npm install --ignore-scripts` here, or add and commit a lockfile for `openclaw-channel-dmwork`.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/ci.yml:23-26`: the matrix correctly uses `npm install` for `openclaw-channel-dmwork` and `npm ci` for `openclaw-channel-octo`, matching the current lockfile reality. Consider adding a short YAML comment explaining why the two install commands differ, since the PR description says CI is switching to `npm ci`.

✅ Highlights

The CI matrix now covers both package directories, Node is aligned to 22, and the pinned `actions/checkout` / `actions/setup-node` SHAs match the current `v4` tag targets.

lml2468 (commented):
Intended verdict: CHANGES_REQUESTED. I cannot submit a formal changes-requested review because GitHub treats this authenticated account as the PR author, but the blockers below should prevent merge.

Previous blocker status: still-open. `.github/workflows/publish-octo.yml:163-165` still runs `npm ci --ignore-scripts && npm run type-check` from `openclaw-channel-dmwork`, which has no committed `package-lock.json` / `npm-shrinkwrap.json`; this will fail with `EUSAGE` in `publish_shim` before the shim can build or publish. The prior recommendation still stands: delete this type-check step unless there is a real shim-specific check to run.

Assessment of new changes: adding explicit minimal permissions is the right direction, but `permissions: {}` on the caller workflows is not safe here. The called reusable workflows declare their required token permissions, but reusable-workflow permissions can only be maintained or reduced from the caller. With the caller set to `{}`, `octo-ci-status` will not receive the `actions: read` / `contents: read` permissions it declares, and `octo-issue-feed` will not receive `contents: read` for its checkout step. Set the caller permissions to the minimum required scopes instead: `actions: read` + `contents: read` for `.github/workflows/octo-ci-status.yml`, and `contents: read` for `.github/workflows/octo-issue-feed.yml`.

Blocking items:
- `.github/workflows/publish-octo.yml:163-165` — `publish_shim` still runs `npm ci` in `openclaw-channel-dmwork`, which has no lockfile.
- `.github/workflows/octo-ci-status.yml:9` — caller passes no `GITHUB_TOKEN` scopes to a reusable workflow that requires `actions: read` and `contents: read`.
- `.github/workflows/octo-issue-feed.yml:7` — caller passes no `GITHUB_TOKEN` scopes to a reusable workflow that requires `contents: read`.

Non-blocking notes: none.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but it does not fully deliver the deterministic CI change it advertises.

🔴 Blocking

- 🟡 Warning: `.github/workflows/ci.yml:23-26` still runs `npm install` for `openclaw-channel-dmwork`. That package has no committed `package-lock.json`, so CI remains non-deterministic for half of the matrix and can start passing or failing based on newly resolved transitive dependency versions. This directly conflicts with the stated `npm ci` goal. Please add a lockfile for `openclaw-channel-dmwork` and use `npm ci` for both matrix entries, or explicitly split this out of the PR.

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/publish-octo.yml:131-166` does not include the “type-check to publish_shim job” described in the PR. The shim is JavaScript rather than TypeScript, so a full type-check may not apply, but consider adding a cheap validation such as building the shim and running `node --check /tmp/shim-build/bin/dmwork.js`, or remove that claim from the PR description.

- 🔵 Suggestion: `.github/workflows/ci.yml:20-26` uses the default matrix `fail-fast: true`. Since the purpose is to cover both packages, `fail-fast: false` would make failures easier to diagnose when both package jobs are relevant.

- 🔵 Suggestion: `.github/workflows/octo-ci-status.yml:13` adds a reusable workflow pinned to `@v1` while the PR is otherwise improving SHA pinning. If the policy is strict pinning for executable workflow dependencies, pin this reusable workflow reference too.

✅ Highlights

- The CI matrix now includes both `openclaw-channel-dmwork` and `openclaw-channel-octo`.
- Node.js is consistently moved to 22 across CI and publish workflows.
- The pinned `actions/checkout` and `actions/setup-node` SHAs match the current `v4` tag targets.

lml2468 (commented):
# Code Review — PR #24 (incremental re-review #5 by 齐静春 📖)

Reviewed incremental diff `4ad103a...c48cceb` (1 commit, 1 file).

## Intended verdict: APPROVED ✅

All previous merge-blocking issues are resolved. Core PR goals (CI coverage, Node 22, SHA pinning) achieved.

---

## Previous blocker resolution

### P0 publish_shim type-check — ✅ RESOLVED
The broken `npm ci` + wrong-package type-check step (`publish-octo.yml:162-164`) has been deleted. Clean fix.

### P0 octo-ci-status.yml permissions — ⚠️ CORRECTED + DOWNGRADED to P1
I verified the called reusable workflow (`Mininglamp-OSS/.github/.../octo-ci-status.yml@v1`):
- It declares `permissions: { actions: read }` and uses `GITHUB_TOKEN` to fetch workflow run history from `api.github.com/repos/.../actions/runs/...`
- The caller's `permissions: {}` means GITHUB_TOKEN gets zero scopes → the called workflow cannot elevate → API calls will fail with 403
- **Fix**: change caller to `permissions: { actions: read }`

Downgraded to P1 because this only affects the CI status notification (not CI checks or publishing). The notification feature is new in this PR and the fix is trivial.

### P0 octo-issue-feed.yml permissions — ❌ FALSE POSITIVE, RETRACTED
I re-read the called reusable workflow and found it declares `permissions: {}` itself and does **not** use `GITHUB_TOKEN` — it only uses `OCTO_BOT_TOKEN` via inputs. All issue data is passed via `with:` parameters. The caller's `permissions: {}` is correct. My previous flag was wrong.

---

## New changes assessment

The single commit cleanly deletes the broken type-check step. No new issues introduced.

## CI status

Both matrix legs green ✅. Mergeable ✅.

## Non-blocking note

- `octo-ci-status.yml` caller needs `permissions: { actions: read }` for the notification to actually work (P1, can be a follow-up commit)

Jerry-Xin (dismissed):
Incremental re-review at c48cceb. All previous blockers resolved.

✅ P0 publish_shim type-check — deleted, clean fix.
✅ CI matrix — both legs green, dmwork on npm install / octo on npm ci.
✅ SHA pins — consistent across ci.yml, publish-octo.yml, publish-octo-dev.yml.
✅ Node 22 — unified.
✅ octo-issue-feed.yml permissions: {} — correct (uses OCTO_BOT_TOKEN only, no GITHUB_TOKEN).

📝 P1 follow-up (non-blocking):
- octo-ci-status.yml caller `permissions: {}` should be `permissions: { actions: read }`. The called reusable workflow uses GITHUB_TOKEN to fetch workflow run history; `{}` gives zero scopes → 403. Notification feature is new in this PR and won't affect CI/publish, so fine as a follow-up commit.

Downgrading my previous blocker (dmwork npm install non-determinism) — dmwork is maintenance mode and the conditional install_cmd is a pragmatic compromise. Not worth holding the PR.

Ship it 🚢

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-adapters` and the CI/package workflow changes look mergeable.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion: The PR description says the `publish_shim` job gets a type-check, but `.github/workflows/publish-octo.yml:131` through `.github/workflows/publish-octo.yml:167` still only resolves the version, builds the shim, and publishes it. The main `test` job does run `npx tsc --noEmit` at `.github/workflows/publish-octo.yml:68`, so this is not blocking, but the description or workflow should be aligned.

🔵 Suggestion: `npm audit --omit=dev` still reports production transitive vulnerabilities under `cos-nodejs-sdk-v5`, including `fast-xml-parser`, `request`, and `form-data` in `openclaw-channel-dmwork/package-lock.json:3813`, `:4708`, `:7412`, and `:7445`. This appears pre-existing from the direct runtime dependency rather than introduced by this PR, but it is worth tracking separately.

✅ Highlights  
The CI matrix in `.github/workflows/ci.yml:20` now covers both `openclaw-channel-dmwork` and `openclaw-channel-octo`, and both package lockfiles support the move to `npm ci`.

The lockfile exception in `.gitignore:5` correctly allows `openclaw-channel-dmwork/package-lock.json` to be tracked.

Verification performed: full diff reviewed with `gh pr diff`; both packages passed `npm ci --ignore-scripts`, `npm run type-check`, `npm run build`, and `npm test`. `git diff --check main...HEAD` also passed.

PR #37, opened 2026-05-19
feat(openclaw-channel-octo): install via ClawHub (clawhub:octo) with legacy npm cleanup

caster-Q:
## Summary

Reroute `npx openclaw-channel-octo install` to install the ClawHub plugin (plugin id `octo`) instead of the npm package itself. Preserves the existing dmwork→octo rebrand migration path; adds transactional cleanup of any pre-2.0 npm-installed `openclaw-channel-octo` plugin to avoid duplicate `octo` channel registration after install.

Plugin runtime code unchanged in this PR (kept for now — to be deprecated in a later phase that publishes only the CLI shim to npm and consolidates plugin code into `Mininglamp-OSS/openclaw-channel-octo`).

## Related Issue

N/A

## Changes

- **`src/constants.ts`**: `PLUGIN_ID` semantic shift `"openclaw-channel-octo"` → `"octo"` (ClawHub plugin id); add `NPM_PACKAGE_NAME` and `CLAWHUB_INSTALL_SPEC`
- **`cli/install.ts`**:
  - spec defaults to `CLAWHUB_INSTALL_SPEC` (`"clawhub:octo"`)
  - `getLatestClawHubVersion` via `openclaw plugins inspect --json` (no external `clawhub` binary dep), falls back to standalone `clawhub` CLI
  - `--dev` / `--next` deprecated: warn + fall back to latest (ClawHub release model differs from npm dist-tags)
  - Transactional legacy npm cleanup: detect at entry → install ClawHub → verify `isHealthyInstall(PLUGIN_ID)` → uninstall legacy → throw on uninstall failure (loud, avoids silent double-registration)
- **`cli/doctor.ts`**: `--fix` delegates to `runInstall({ force: true })` for full scenario coverage; surfaces `err.message` in failure detail
- **`src/channel.ts`**: runtime version check uses ClawHub registry (`https://clawhub.ai/api/v1/packages/octo`); upgrade prompt uses `openclaw plugins install clawhub:octo --force` or `npx openclaw-channel-octo install`
- **`cli/index.ts`**: `info` shows installed plugin id (`octo`) + npm package name; install/update `--next` help updated to `[v2.0.0+ no-op]`
- **`README.md`**: `--dev` / `--next` flag deprecation notice
- **Tests**: 749/749 pass — `install.test.ts` rewritten for new behavior + transactional cleanup test; `install-migration.test.ts` updated to assert `clawhub:octo` spec and `octo` plugin id

## Testing

- [x] Manually verified

### Automated

- `npm run build` — passes
- `npm test` — 749/749 pass
- codex review (4 rounds) — Ready to ship; 0 BLOCKER / 0 MAJOR remaining

### E2E (local OpenClaw 2026.5.18 + ClawHub octo 1.0.7)

- **Scenario 1: fresh install** — no prior plugin → `npx ... install` → `~/.openclaw/extensions/octo/` populated, `openclaw channels list --all` shows Octo as installable
- **Scenario 2: dmwork → octo rebrand** — `openclaw-channel-dmwork` v0.6.4 + `channels.dmwork.*` + dmwork bindings → `npx ... install` → `channels.dmwork` migrated to `channels.octo`, bindings rewritten to `channel=octo`, dmwork plugin uninstalled, feishu channel unaffected
- **Scenario 3: dmwork → octo + IM end-to-end** — sent a message to the migrated bot in OpenClaw IM, received correct response (proves channel + binding migration is functionally correct)
- **Scenario 4: npm rebrand → ClawHub** — npm `openclaw-channel-octo` v1.0.0 (plugin id `openclaw-channel-octo`) + `channels.octo.*` already configured + binding → `npx ... install` → deadlock-repair path (channels.octo exists but `PLUGIN_ID=octo` not yet installed) → ClawHub octo installed → `channels.octo` + bindings restored intact → legacy npm plugin uninstalled only after `isHealthyInstall` verified (transactional)

## Release plan

Version intentionally not bumped in this PR; merging will publish `@dev` (no impact on stable users). A future formal `v2.0.0` release requires bumping `package.json` + `CHANGELOG.md`, and `npm deprecate openclaw-channel-octo@'<2.0.0'` is recommended to nudge legacy users.

## Checklist

- [x] Updated documentation

### DIFF
--- a/openclaw-channel-octo/README.md
+++ b/openclaw-channel-octo/README.md
@@ -35,7 +35,8 @@ npx -y openclaw-channel-octo bind \
 `install` flags:
 
 - `--force`: reinstall even if already installed
-- `--dev`: install the `@dev` dist-tag instead of `@latest`
+- `--from <spec>`: install from a local tarball or alternate `openclaw plugins install` spec (pre-publish local testing)
+- `--dev` / `--next`: deprecated in v2.0.0+ (ClawHub installs use a single channel; use `--from` for pre-release testing)
 
 ## CLI Commands
 
--- a/openclaw-channel-octo/cli/doctor.ts
+++ b/openclaw-channel-octo/cli/doctor.ts
@@ -183,32 +183,28 @@ export async function runDoctorChecks(params: {
         detail: `${versionLabel}${sourceNote}`,
       });
     } else if (fix) {
-      // Not installed (or broken/partial). Use detectScenario() for precise fix.
+      // Not installed (or broken/partial). Defer to runInstall for full
+      // scenario handling — it covers legacy-to-octo / rebrand / legacy /
+      // deadlock / broken / fresh with the correct CLAWHUB_INSTALL_SPEC
+      // and transactional rollback. Doctor used to inline a partial subset
+      // (only legacy/deadlock/broken) and passed PLUGIN_ID as install spec,
+      // which broke after PLUGIN_ID's semantics changed from npm name to
+      // ClawHub plugin id "octo".
       try {
-        const scenario = detectScenario();
-        if (scenario === "legacy") {
-          const { runLegacyMigrationForUpdate } = await import("./install.js");
-          runLegacyMigrationForUpdate(PLUGIN_ID, true);
-        } else if (scenario === "deadlock") {
-          const { runDeadlockRepairForUpdate } = await import("./install.js");
-          runDeadlockRepairForUpdate(PLUGIN_ID, true);
-        } else if (scenario === "broken") {
-          cleanupBrokenInstall();
-          pluginsInstall(PLUGIN_ID, true);
-        } else {
-          pluginsInstall(PLUGIN_ID, true);
-        }
+        const { runInstall } = await import("./install.js");
+        await runInstall({ force: true });
         pluginState = resolvePluginState(PLUGIN_ID);
         checks.push({
           name: "Plugin installed",
           status: "FIXED",
           detail: `Installed v${pluginState.version ?? "unknown"}`,
         });
-      } catch {
+      } catch (err) {
+        const errMsg = err instanceof Error ? err.message : String(err);
         checks.push({
           name: "Plugin installed",
           status: "FAIL",
-          detail: "Not installed (auto-install failed)",
+          detail: `Auto-install failed: ${errMsg}`,
         });
         return summarize(checks);
       }
--- a/openclaw-channel-octo/cli/doctor.ts
+++ b/openclaw-channel-octo/cli/doctor.ts
@@ -238,7 +238,7 @@ export async function runDoctorChecks(params: {
           name: "Plugin enabled",
           status: "FAIL",
           detail:
-            "No (auto-fix failed; run `openclaw plugins enable ${PLUGIN_ID}`)",
+            `No (auto-fix failed; run \`openclaw plugins enable ${PLUGIN_ID}\`)`,
         });
       }
     } else {
@@ -247,7 +247,7 @@ export async function runDoctorChecks(params: {
         status: "WARN",
         detail:
           "No — installed but disabled. Run `openclaw-channel-octo doctor --fix` " +
-          "or `openclaw plugins enable ${PLUGIN_ID}`.",
+          `or \`openclaw plugins enable ${PLUGIN_ID}\`.`,
       });
     }
   }
--- a/openclaw-channel-octo/cli/index.ts
+++ b/openclaw-channel-octo/cli/index.ts
@@ -14,7 +14,7 @@ import {
 } from "./doctor.js";
 import { runUninstall } from "./uninstall.js";
 import { runRemoveAccount } from "./remove-account.js";
-import { ensureOpenClawCompat, PLUGIN_ID } from "./utils.js";
+import { ensureOpenClawCompat, NPM_PACKAGE_NAME, PLUGIN_ID } from "./utils.js";
 import { getOpenClawVersionStrict, resolvePluginState } from "./openclaw-cli.js";
 import { PLUGIN_VERSION } from "../src/version.js";
 
@@ -55,8 +55,9 @@ program
 
     console.log(`${b}openclaw-channel-octo-cli:${r} ${g}${PLUGIN_VERSION}${r}`);
     console.log(`${b}openclaw:${r} ${g}${openclawVersion}${r}`);
-    console.log(`${b}openclaw-channel-octo:${r} ${g}${installedVersion}${r}`);
-    console.log(`${b}plugin package:${r} ${g}${PLUGIN_ID}${r}`);
+    console.log(`${b}installed plugin id:${r} ${g}${PLUGIN_ID}${r} (ClawHub)`);
+    console.log(`${b}installed plugin version:${r} ${g}${installedVersion}${r}`);
+    console.log(`${b}npm package:${r} ${g}${NPM_PACKAGE_NAME}${r}`);
     console.log();
     console.log(`${b}Environment:${r}`);
     console.log(`${b}OS:${r} ${g}${process.platform} ${process.arch}${r}`);
@@ -69,10 +70,14 @@ program
   .command("install")
   .description("Install or update the Octo plugin")
   .option("--force", "Force reinstall", false)
-  .option("--next", "Install the @next pre-release dist-tag (use during rc/beta cycles)", false)
+  .option(
+    "--next",
+    "[v2.0.0+ no-op] Previously selected npm @next dist-tag. ClawHub installs use a single channel; use --from <tarball> for pre-release testing.",
+    false,
+  )
   .option(
     "--from <spec>",
-    "Override install spec (tarball path or alternate npm spec). Pre-publish local testing.",
+    "Override install spec (tarball path or alternate `openclaw plugins install` spec). Pre-publish local testing.",
   )
   .addOption(new Option("--dev").hideHelp().default(false))
   .action(async (opts) => {
@@ -88,7 +93,11 @@ program
 program
   .command("update")
   .description("Update the Octo plugin (alias for install)")
-  .option("--next", "Update to the @next pre-release dist-tag", false)
+  .option(
+    "--next",
+    "[v2.0.0+ no-op] Previously selected npm @next dist-tag. ClawHub installs use a single channel; use `install --from <tarball>` for pre-release testing.",
+    false,
+  )
   .addOption(new Option("--dev").hideHelp().default(false))
   .action(async (opts) => {
     await runUpdate({ dev: opts.dev, next: opts.next });
--- a/openclaw-channel-octo/cli/install-migration.test.ts
+++ b/openclaw-channel-octo/cli/install-migration.test.ts
@@ -527,7 +527,7 @@ describe("runInstall — rebrand scenario (openclaw-channel-dmwork → octo)", (
         state.cfg.plugins.entries["octo"] = { enabled: true };
         state.cfg.plugins.installs["octo"] = { source: "npm", version: "1.0.0" };
         state.extDirs.add("octo");
-        inspectMap["openclaw-channel-octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
+        inspectMap["octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
         return "";
       }
       if (a[0] === "plugins" && (a[1] === "enable" || a[1] === "disable")) return "";
@@ -702,7 +702,7 @@ describe("runInstall — legacy-to-octo scenario (very-legacy 'dmwork' → octo)
         state.cfg.plugins.entries["octo"] = { enabled: true };
         state.cfg.plugins.installs["octo"] = { source: "npm", version: "1.0.0" };
         state.extDirs.add("octo");
-        inspectMap["openclaw-channel-octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
+        inspectMap["octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
         return "";
       }
       if (a[0] === "plugins" && a[1] === "uninstall") {
@@ -772,7 +772,7 @@ describe("runInstall — legacy-to-octo scenario (very-legacy 'dmwork' → octo)
         state.cfg.plugins.entries["octo"] = { enabled: true };
         state.cfg.plugins.installs["octo"] = { source: "npm", version: "1.0.0" };
         state.extDirs.add("octo");
-        inspectMap["openclaw-channel-octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
+        inspectMap["octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
         return "";
       }
       if (a[0] === "plugins" && a[1] === "uninstall") {
@@ -860,7 +860,7 @@ describe("runInstall — deadlock scenario (channels.octo without plugin)", () =
         state.cfg.plugins.entries["octo"] = { enabled: true };
         state.cfg.plugins.installs["octo"] = { source: "npm", version: "1.0.0" };
         state.extDirs.add("octo");
-        inspectMap["openclaw-channel-octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
+        inspectMap["octo"] = { plugin: { id: "octo", version: "1.0.0", enabled: true } };
         return "";
       }
       if (a[0] === "plugins" && (a[1] === "enable" || a[1] === "disable")) return "";
--- a/openclaw-channel-octo/cli/install-migration.test.ts
+++ b/openclaw-channel-octo/cli/install-migration.test.ts
@@ -2,12 +2,13 @@
  * Phase B migration tests.
  *
  * Two scenarios share the same runMigration() implementation:
- *   - rebrand:        openclaw-channel-dmwork → openclaw-channel-octo
- *   - legacy-to-octo: very-legacy "dmwork" plugin id → openclaw-channel-octo
+ *   - rebrand:        openclaw-channel-dmwork → ClawHub octo (plugin id "octo")
+ *   - legacy-to-octo: very-legacy "dmwork" plugin id → ClawHub octo
  *
- * Both transform channels.dmwork → channels.octo and rewrite bindings'
- * match.channel from "dmwork" to "octo". Tests assert the command sequence
- * (via execFileSync mock) and final config state (via fs writeFileSync mock).
+ * Both install via CLAWHUB_INSTALL_SPEC ("clawhub:octo"), transform
+ * channels.dmwork → channels.octo, and rewrite bindings' match.channel from
+ * "dmwork" to "octo". Tests assert the command sequence (via execFileSync
+ * mock) and final config state (via fs writeFileSync mock).
  */
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
--- a/openclaw-channel-octo/cli/install-migration.test.ts
+++ b/openclaw-channel-octo/cli/install-migration.test.ts
@@ -208,25 +208,25 @@ describe("runInstall — rebrand scenario (openclaw-channel-dmwork → octo)", (
     const origExecImpl = mockExecFileSync.getMockImplementation();
     mockExecFileSync.mockImplementation((cmd: any, args: any) => {
       const a = args as string[];
-      if (a[0] === "plugins" && a[1] === "install" && a[2]?.startsWith("openclaw-channel-octo")) {
+      if (a[0] === "plugins" && a[1] === "install" && a[2] === "clawhub:octo") {
         // Side effect: octo install populates entries/installs/dir
         state.cfg.plugins ??= {};
         state.cfg.plugins.entries ??= {};
-        state.cfg.plugins.entries["openclaw-channel-octo"] = { enabled: true };
+        state.cfg.plugins.entries["octo"] = { enabled: true };
         state.cfg.plugins.installs ??= {};
-        state.cfg.plugins.installs["openclaw-channel-octo"] = { source: "npm", version: "1.0.0" };
+        state.cfg.plugins.installs["octo"] = { source: "npm", version: "1.0.0" };
         state.cfg.plugins.allow ??= [];
-        if (!state.cfg.plugins.allow.includes("openclaw-channel-octo")) {
-          state.cfg.plugins.allow.push("openclaw-channel-octo");
+        if (!state.cfg.plugins.allow.includes("octo")) {
+          state.cfg.plugins.allow.push("octo");
         }
-        state.extDirs.add("openclaw-channel-octo");
+        state.extDirs.add("octo");
         return "";
       }
       // After install, inspect octo should succeed too — re-dispatch via opts
-      if (a[0] === "plugins" && a[1] === "inspect" && a[2] === "openclaw-channel-octo") {
+      if (a[0] === "plugins" && a[1] === "inspect" && a[2] === "octo") {
         if (state.cfg.plugins?.entries?.["openclaw-channel-octo"]) {
           return JSON.stringify({
-            plugin: { id: "openclaw-channel-octo", version: "1.0.0", enabled: true },
+            plugin: { id: "octo", version: "1.0.0", enabled: true },
           });
         }
       }
@@ -239,8 +239,8 @@ describe("runInstall — rebrand scenario (openclaw-channel-dmwork → octo)", (
     const ops = calledOps();
     // Sequence: disable legacy → install octo → enable octo → uninstall legacy
     const disableIdx = ops.findIndex((o) => /^plugins disable openclaw-channel-dmwork$/.test(o));
-    const installIdx = ops.findIndex((o) => /^plugins install openclaw-channel-octo/.test(o));
-    const enableIdx = ops.findIndex((o) => /^plugins enable openclaw-channel-octo$/.test(o));
+    const installIdx = ops.findIndex((o) => /^plugins install clawhub:octo/.test(o));
+    const enableIdx = ops.findIndex((o) => /^plugins enable octo$/.test(o));
     const uninstallIdx = ops.findIndex((o) => /^plugins uninstall openclaw-channel-dmwork/.test(o));
 
     expect(disableIdx).toBeGreaterThanOrEqual(0);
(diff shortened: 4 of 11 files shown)


Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-adapters`, but the `PLUGIN_ID = "octo"` semantic shift leaks into runtime command handlers, and the legacy npm cleanup can remove existing `channels.octo` config.

**🔴 Blocking**

🔴 Critical: Runtime `/octo_install` and `/octo_update` now use the wrong install/update target.
`src/constants.ts:15` changes `PLUGIN_ID` to `"octo"`, but `index.ts` still treats `PLUGIN_ID` as an npm/package install spec. As a result, `handleInstall()` calls `pluginsInstall(PLUGIN_ID, ...)`, which becomes `openclaw plugins install octo` instead of `clawhub:octo`; `handleUpdate()` queries `npm view octo@latest` and installs `octo@latest`, which is unrelated to the new ClawHub flow. See `index.ts:78`, `index.ts:94`, and `index.ts:100`. Runtime slash commands need the same split introduced for the CLI: plugin id `octo`, install spec `clawhub:octo`, and package name `openclaw-channel-octo`.

🔴 Critical: Legacy npm cleanup can wipe user `channels.octo` configuration.
`runInstall()` detects the old npm plugin, verifies ClawHub `octo`, then calls `pluginsUninstall(NPM_PACKAGE_NAME, true)` directly at `cli/install.ts:264-272`. Both the old npm plugin and the new ClawHub plugin own channel id `octo`, and this repo’s own uninstall path documents that plugin uninstall removes channel configs. See `cli/uninstall.ts:2-4` and the direct config-removal workaround in `cli/openclaw-cli.ts:647-650`. If OpenClaw deletes `channels.octo` during `openclaw plugins uninstall openclaw-channel-octo --force`, this cleanup loses bot accounts after the supposedly successful migration. The new test at `cli/install.test.ts:423-437` does not include `channels.octo`, so it cannot catch this. Save and restore `channels.octo` plus `bindings(channel=octo)` around the legacy uninstall, or use an unregister-only/keep-config path if OpenClaw supports one, and add a regression test that simulates uninstall deleting `channels.octo`.

**💬 Non-blocking**

🟡 Warning: `cli/doctor.ts:241` and `cli/doctor.ts:250` use normal string literals containing `${PLUGIN_ID}`, so the user-facing fix commands print the placeholder instead of `octo`. This predates the core flow but is more visible now that plugin id semantics changed.

🟡 Warning: I could not run the targeted tests locally because `vitest` is not installed in this checkout (`sh: vitest: command not found`).

**✅ Highlights**

The PR is clearly relevant to this repository: it updates the Octo adapter CLI, migration tests, docs, and runtime version messaging.

The install migration code has a good rollback shape for dmwork-to-octo and deadlock repair paths, especially the backup/restore steps around channel config and bindings.

yujiawei (approved):
# Code Review — PR #37 (octo-adapters)

## Summary

Reroutes `npx openclaw-channel-octo install` from npm to the ClawHub plugin (`clawhub:octo`), introduces a `PLUGIN_ID` / `NPM_PACKAGE_NAME` / `CLAWHUB_INSTALL_SPEC` split, and adds a transactional cleanup of any pre-2.0 npm-installed `openclaw-channel-octo` so the new ClawHub install does not double-register the `octo` channel. Migration paths (rebrand from `openclaw-channel-dmwork`, legacy-to-octo from very-legacy `dmwork`) are factored through a single `runMigration()` parameterized by legacy plugin id.

Code is well-structured, transactional semantics are clear, and the new test coverage (cleanup path, `--from` spec, `--dev`/`--next` deprecation, 4.20 hardening) reads thoroughly. Build + 749/749 tests pass locally at `933f453`. No blocker issues found.

## Verdict — APPROVED

No P0/P1 issues. One real-but-cosmetic regression and a few P2 robustness suggestions noted below; none should block merge.

## Findings

### P2 — cleanupStaleStageDirectories check is now dead code

`cli/openclaw-cli.ts:912`

```ts
if (pkg.name === PLUGIN_ID) {
  isDmwork = true;
  break;
}
```

After this PR, `PLUGIN_ID === "octo"`, but the staged install's `package.json.name` is `"openclaw-channel-octo"` (the npm package name — this is what `npm pack` / ClawHub stages). The comparison can no longer match, so doctor's stale-stage cleanup silently no-ops. Stage dirs would accumulate but do not break functionality.

Suggested fix:

```ts
if (pkg.name === NPM_PACKAGE_NAME || pkg.name === PLUGIN_ID) {
  isDmwork = true;
  break;
}
```

This file wasn't touched in the diff, but the `PLUGIN_ID` semantic shift introduced here turns the existing comparison into a no-op. Worth either fixing in this PR or filing a follow-up.

### P2 — Double-registration window between ClawHub install and legacy npm uninstall

`cli/install.ts:264-289`

Transactional design defers the legacy npm uninstall until *after* `isHealthyInstall(PLUGIN_ID)` confirms the new install. Between `pluginsInstall("clawhub:octo")` and `pluginsUninstall(NPM_PACKAGE_NAME)`, both plugins are installed + enabled and both register `channel.id = "octo"`. The mitigation (gateway restart deferred until after uninstall, at line 293) works only if the gateway does not auto-reload on config changes during this window. The e2e Scenario 4 in the PR description reports passing end-to-end, so this appears benign in practice — but worth a manual sanity check on an environment where the gateway watches config writes (e.g. inotify-based hot reload), since `ensurePluginsAllow` / `ensurePluginEnabled` writes config between install and uninstall.

If a config-watcher race is plausible in any supported OpenClaw version, consider disabling the legacy npm plugin (`pluginsDisable(NPM_PACKAGE_NAME)`) *before* the new install, then doing the uninstall after — disable is reversible if the new install fails (rollback can re-enable).

### P2 — Deadlock-repair `pluginsInstall` omits `force`

`cli/install.ts:638`

```ts
console.log(`  Installing ${PLUGIN_ID}...`);
pluginsInstall(spec, quiet);
```

`runDeadlockRepair` is itself a recovery scenario (config says the channel exists but no plugin). If a partial `extensions/octo` dir lingers from a previous crashed attempt, OpenClaw could refuse the bare install. `pluginsInstall(spec, quiet, true)` would mirror the other recovery paths (`broken`, `fresh` with `opts.force`, the npm-cleanup path) and is consistent with the deadlock-repair intent. Optional hardening.

### P2 — `--dev` / `--next` warning fires even with `--from`

`cli/install.ts:147-154`

The doc comment for `from` (line 113-114) says "When set, --dev and --next are ignored." The warning at line 147 fires before that check, so a user invoking `install --from ./pkg.tgz --next` gets a warning that does not actually apply (the `spec = opts.from ?? CLAWHUB_INSTALL_SPEC` line silently wins). Minor UX nit; trivial to gate on `!opts.from`.

### P2 — Untrusted upstream string in log (security_sensitive note)

`src/channel.ts:299-305`

```ts
const data = await resp.json() as { package?: { latestVersion?: string } };
const latest = data?.package?.latestVersion;
if (latest && latest !== localVersion) {
  log?.info?.(`octo: new version available: ${latest} (current: ${localVersion}). ...`);
}
```

`latest` is interpolated into a log line with no sanitization. A compromised or hostile `clawhub.ai` response could inject newlines / terminal escape sequences into the log. This matches the previous npm-registry pattern (same risk surface), and the log goes to the gateway's structured logger which likely handles escaping, but for a `security_sensitive`-tagged PR it's worth a human verification that the logger sanitizes control characters and that `latest` is type-validated (e.g. `typeof latest === "string"` and a regex on semver-like strings).

The new endpoint (`https://clawhub.ai/api/v1/packages/octo`) is hardcoded — good. No new shell-injection surface; `pluginsInstall(spec, ...)` ultimately routes through `execFileSync` (argv, not shell), so user-supplied `--from <spec>` is not shell-interpretable.

### Positive observations

- Transactional semantics around the npm→ClawHub swap are clearly documented and tested (`install.test.ts:411-483` covers the cleanup path explicitly).
- `runMigration` consolidation between `rebrand` and `legacy-to-octo` is a nice cleanup — single rollback closure, single verification block.
- Doctor's `--fix` delegation to `runInstall({ force: true })` (replacing the inline scenario subset) is the right call after `PLUGIN_ID` stopped being the install spec; this avoids a latent bug where `pluginsInstall(PLUGIN_ID, true)` would have shipped `"octo"` as the spec.
- `--from` plumbing skips both `--dev`/`--next` semantics and update-scenario version comparison — correct for unconditional tarball installs.
- `info` command now distinguishes installed plugin id vs npm package name, which makes the new dual identity explicit to users.

## Manual-verification checklist (for the human reviewer)

The PR is labeled `needs-human-review` (security_sensitive). Recommended hand checks before merging:

1. On an environment with an OpenClaw config-watcher / hot-reloader, confirm the install→ensurePluginEnabled→uninstall sequence (lines 249-289) does not transiently double-handle inbound messages.
2. Confirm `clawhub.ai` returns the expected JSON schema (`{ package: { latestVersion: string } }`) and that the gateway logger sanitizes control characters in the `latest` value.
3. Re-run the four e2e scenarios from the PR description (fresh, dmwork-rebrand, dmwork-rebrand+IM, npm→ClawHub) against a real OpenClaw instance — the unit tests cover spec/sequence but not real plugin loading.

lml2468 (changes requested):
## Independent Review — PR #37 (ClawHub install + legacy npm cleanup)

**Verdict: CHANGES_REQUESTED** — the overall migration architecture is solid, but the legacy npm cleanup can nuke `channels.octo` config.

### Phase 1: Gate

CI green, mergeable. 480+/211-, 9 files. Author: caster-Q.

### 🔴 Blocking — concur with @Jerry-Xin: legacy npm cleanup can wipe `channels.octo`

`install.ts:272` calls `pluginsUninstall(NPM_PACKAGE_NAME, true)` to remove the old npm-installed `openclaw-channel-octo`. But both the old npm plugin and the new ClawHub plugin register channel id `"octo"`. OpenClaw's `plugins uninstall` is documented (and implemented — see `uninstall.ts:2-4`, `openclaw-cli.ts:647-650`) to remove `channels.<channelId>` config as part of uninstallation.

Concrete scenario:
1. User has v1.x npm-installed `openclaw-channel-octo` with `channels.octo.accounts.mybot`
2. `runInstall()` installs ClawHub `octo` → both plugins now exist
3. ClawHub install verifies healthy ✅
4. `pluginsUninstall("openclaw-channel-octo", true)` removes the npm plugin **and** `channels.octo` config
5. User loses all bot accounts/tokens 💥

**Fix**: save and restore `channels.octo` + `bindings(channel=octo)` around the uninstall, same pattern already used in the rebrand migration (`runMigration`). The existing test at `install.test.ts:423-437` sets up an `inspect` mock for the legacy plugin but has **no `channels.octo` config in the test fixture**, so it cannot catch this regression.

### 🔴 Blocking (new finding): `doctor.ts` string literal bug at lines 241 and 250

```typescript
// Line 241 — normal string literal, NOT template literal
"No (auto-fix failed; run `openclaw plugins enable ${PLUGIN_ID}`)",
// Line 250
"or `openclaw plugins enable ${PLUGIN_ID}`.",
```

These use `"..."` (normal strings), so `${PLUGIN_ID}` prints literally as `${PLUGIN_ID}` instead of resolving to `octo`. This is a functional bug — users following the printed instructions will run a command that doesn't work. Fix: use template literals `` `...` `` instead of `"..."`.

### ✅ Allen's first finding about runtime slash commands — I disagree

Allen flagged that `index.ts` uses `PLUGIN_ID` as an install spec. After tracing the code:
- `index.ts` install/update commands call `runInstall()` → which uses `CLAWHUB_INSTALL_SPEC` ("clawhub:octo") internally
- `PLUGIN_ID` in `index.ts` is only used for `resolvePluginState(PLUGIN_ID)` (health check) and display
- `runUpdate()` delegates to `runInstall()` which has all the correct spec handling

So this is **not a bug** — the indirection through `runInstall()` correctly resolves the install spec.

### 🟡 Non-blocking

1. **`getLatestClawHubVersion()` fallback chain** (`install.ts:52-71`): tries `openclaw plugins inspect octo --json` first, then falls back to `clawhub` CLI binary. The fallback to an external `clawhub` binary feels speculative — is it a real dependency? If not, the dead `try/catch` is noise.

2. **Version comparison when `targetVersion` is null** (`install.ts:193-209`): the new `pluginsUpdate(PLUGIN_ID, quiet)` delegation is cleaner than the old silent skip, but it adds a code path that's hard to test (requires `openclaw plugins update` to exist and behave correctly). Worth a comment noting this is best-effort.

### ✅ Highlights

- The two-phase migration design (detect npm plugin → install ClawHub → verify health → THEN cleanup) is the right architecture for a safe transition
- Error handling on cleanup failure (throw with manual fix instructions) is appropriate — leaving duplicate channel registrations is worse than a loud error
- Doctor's fix path now correctly delegates to `runInstall()` instead of inlining partial scenario handling — single source of truth
- Test coverage for ClawHub install specs is thorough (migration, deadlock, update, force, --from scenarios all updated)

yujiawei (changes requested):
# Code Review — PR #37 (octo-adapters) — Revised

Replacing my earlier APPROVE. After a second pass triggered by @Jerry-Xin's findings, I confirmed both flagged issues are real P0 blockers. My initial review missed them because I focused on the `cli/` diff and did not re-audit the runtime `index.ts` slash-command handlers, nor did I model the side effect of `openclaw plugins uninstall` against `channels.<id>` config. Verdict revised to CHANGES_REQUESTED.

## P0 — Runtime `/octo_install` and `/octo_update` install the wrong target

`openclaw-channel-octo/index.ts:74-100`

After this PR `PLUGIN_ID === "octo"` (ClawHub plugin id), but `index.ts` still passes `PLUGIN_ID` everywhere it previously used the npm package name:

```ts
// index.ts:78
pluginsInstall(PLUGIN_ID, true, force);
// → `openclaw plugins install octo` (no ClawHub spec; will fetch a package named "octo")

// index.ts:94
const targetVersion = execFileSync("npm", ["view", `${PLUGIN_ID}@latest`, "version"], …).trim();
// → `npm view octo@latest version` (queries the unrelated `octo` package on npmjs.org)

// index.ts:100
pluginsInstall(`${PLUGIN_ID}@latest`, true, true);
// → `openclaw plugins install octo@latest` (also wrong)
```

The CLI was updated to use `CLAWHUB_INSTALL_SPEC` and `NPM_PACKAGE_NAME` for these three concerns, but the slash-command handlers in `index.ts` were not. Net effect: `/octo_install` and `/octo_update` invoked from inside OpenClaw silently call the wrong install/update target and either install the unrelated `octo` package or fail outright. `handleInfo()` at line 65 (`plugin package: ${PLUGIN_ID}`) is also misleading for the same reason.

Suggested fix mirroring the CLI's split:
- `handleInstall` → `pluginsInstall(CLAWHUB_INSTALL_SPEC, true, force)` (or call into `runInstall()` so the npm-cleanup path also fires when invoked via slash command)
- `handleUpdate` → use the same `getLatestClawHubVersion()` path the CLI uses, then `pluginsInstall(CLAWHUB_INSTALL_SPEC, true, true)`
- `handleInfo` → print `installed plugin id: ${PLUGIN_ID}` + `npm package: ${NPM_PACKAGE_NAME}` like `cli/index.ts` already does

## P0 — Legacy npm cleanup can wipe user `channels.octo` config

`openclaw-channel-octo/cli/install.ts:264-289`

After ClawHub install verifies healthy, the cleanup path calls `pluginsUninstall(NPM_PACKAGE_NAME, true)` directly with no protection around `channels.octo`. The repo itself documents that `openclaw plugins uninstall` removes the plugin's channel config — see `cli/uninstall.ts:2-4`:

> Removes plugin + all channels.<CHANNEL_ID> config (openclaw CLI does this automatically). Phase A boundary: legacy channels.dmwork is left intact.

Both the legacy npm `openclaw-channel-octo` plugin and the new ClawHub `octo` plugin own `channel.id === "octo"` (see `src/channel.ts:354` / `:364`), so `openclaw plugins uninstall openclaw-channel-octo --force` will delete `channels.octo` along with the plugin record — wiping the bot accounts the user just migrated.

The new cleanup-path test at `cli/install.test.ts:411-483` does not include `channels.octo` in the fixture cfg (only `plugins.entries` and `plugins.installs`), so it cannot catch this regression. Run on a real OpenClaw with a configured account and you should see `channels.octo.accounts.*` disappear after the supposedly-successful migration.

Suggested fix (mirroring the existing rebrand path that already does this for `channels.dmwork`):
1. Before `pluginsUninstall(NPM_PACKAGE_NAME, true)`: `saveChannelConfigFromFile(CHANNEL_ID)` + `saveBindingsFromFile(CHANNEL_ID)`.
2. Uninstall the legacy npm plugin.
3. `restoreChannelConfigToFile(savedChannelConfig, CHANNEL_ID)` + `restoreBindingsToFile(savedBindings, CHANNEL_ID, CHANNEL_ID)`.
4. Add a regression test that puts `channels.octo.accounts.foo` in the fixture and asserts it survives the cleanup.

If OpenClaw exposes an unregister-only / keep-config mode for plugin uninstall, that would be cleaner — otherwise the save/restore dance is required.

## Why my Scenario 4 verification went wrong

I trusted the PR description's "Scenario 4 passed" claim, where the e2e fixture had `channels.octo + binding already configured + bot reachable in IM after install`. Reading the test again, that scenario goes through `runDeadlockRepair` (because `octo` plugin was not installed before the install run), not through the entry-point npm-cleanup path — so it doesn't exercise `pluginsUninstall(NPM_PACKAGE_NAME, true)` against an existing `channels.octo`. The cleanup path is exercised only when both the legacy npm plugin AND a healthy ClawHub `octo` are installed at entry, which the e2e didn't simulate.

## Carry-over from my earlier pass (still applies, all P2)

These are real but non-blocking; including them so they're not lost when the P0s above get fixed.

- **`cli/openclaw-cli.ts:912`**: `cleanupStaleStageDirectories` checks `pkg.name === PLUGIN_ID`. Staged dirs contain `package.json.name = "openclaw-channel-octo"`, so the comparison is now dead. Suggest `pkg.name === NPM_PACKAGE_NAME || pkg.name === PLUGIN_ID`.
- **`cli/install.ts:638`**: deadlock-repair `pluginsInstall(spec, quiet)` — recovery scenario, but no `force`. Consider `pluginsInstall(spec, quiet, true)` for consistency with the other recovery paths.
- **`cli/install.ts:147-154`**: `--dev` / `--next` warning fires even when `--from` overrides them; should gate on `!opts.from`.
- **`src/channel.ts:299-305`** (security_sensitive note): the `latest` string from `clawhub.ai` is interpolated into a log without sanitization; for the security_sensitive label, worth confirming the host logger strips control characters.

## Verdict — CHANGES_REQUESTED

Both P0 fixes are scoped and self-contained. Once `index.ts` slash handlers are split to use `CLAWHUB_INSTALL_SPEC` / `NPM_PACKAGE_NAME` and the legacy npm uninstall is wrapped in save/restore (with a regression test), this PR should be in good shape.

Jerry-Xin (approved):
One-line summary: The PR is relevant to `octo-adapters` and the main `npx ... install` migration path looks ready, with one non-blocking consistency concern.

💬 Non-blocking

🟡 Warning: [openclaw-channel-octo/index.ts](/private/tmp/pr-review-05FwAN/repo/openclaw-channel-octo/index.ts:73) `/octo_install` now installs `clawhub:octo` directly but does not use `runInstall()`, so it bypasses the transactional legacy npm cleanup and config preservation added for the CLI path. If this runtime command remains supported for v2/dev installs, consider delegating to `runInstall({ force })` or explicitly documenting/removing this install surface to avoid leaving both `openclaw-channel-octo` and `octo` installed after a restart.

✅ Highlights

- The `npx openclaw-channel-octo install` path correctly separates install spec (`clawhub:octo`) from installed plugin id (`octo`).
- Legacy npm cleanup is deferred until after `isHealthyInstall("octo")`, which preserves the working old install if ClawHub install fails.
- `channels.octo` and `bindings(channel=octo)` are saved/restored around legacy uninstall, covering the OpenClaw uninstall side effect.
- Doctor `--fix` now delegates to the full install flow instead of duplicating partial scenario logic.

Verification note: I attempted the focused test command, but this checkout does not have dependencies installed (`vitest: command not found`). `git diff --check` passed.

yujiawei (approved):
# Code Review — PR #37 (octo-adapters)

## Summary

Verdict: **APPROVED**.

This is a careful, well-tested migration PR. The transactional design — install ClawHub `octo` first, verify healthy via `isHealthyInstall(PLUGIN_ID)`, then uninstall the legacy npm `openclaw-channel-octo` — correctly avoids a window where the user has no working plugin. The save→uninstall→restore of `channels.octo` and `bindings(channel=octo)` (added in commit `dac291…`/`dac291…` per the diff: `install.ts:275-307`) is the right fix for the side-effecting `openclaw plugins uninstall` behavior. Tests added cover both the cleanup itself and the channel-preservation regression. Manual E2E across 4 scenarios is documented in the PR body.

No P0/P1 blockers. A few P2/nits below.

## Findings

### P2 — Rebrand / deadlock paths do not pre-disable the legacy npm `openclaw-channel-octo`

Location: `cli/install.ts:402-410` (`runMigration` step 2) and `cli/install.ts:636-687` (`runDeadlockRepair`).

If a user has *both* `openclaw-channel-dmwork` and the legacy npm `openclaw-channel-octo` v1.x installed at the time of upgrade, `detectScenario()` returns `"rebrand"` (driven by the dmwork artifacts in `hasLegacyPluginArtifacts`, `cli/openclaw-cli.ts:1504-1515`). `runMigration` step 2 then disables only `LEGACY_PLUGIN_ID` (`openclaw-channel-dmwork`). The legacy npm `openclaw-channel-octo` v1.x — which also registers `channel.id = "octo"` — remains enabled when `pluginsInstall("clawhub:octo")` is invoked at step 5 (`cli/install.ts:439-442`). Depending on the OpenClaw version's behavior on duplicate channel-id registration, this can either fail the install (then rollback) or silently leave two plugins owning `octo`.

The same omission applies to `runDeadlockRepair` (`cli/install.ts:662-667`): no pre-install disable of `NPM_PACKAGE_NAME`. The PR body's Scenario 4 testing covers the `npm openclaw-channel-octo` + `channels.octo` → deadlock-repair path and reports success, so in practice this appears to work on the OpenClaw version exercised. Still worth tightening for defense in depth.

Suggestion: before the `pluginsInstall(spec)` in both `runMigration` and `runDeadlockRepair`, capture `capturePluginState(NPM_PACKAGE_NAME)` and `pluginsDisable(NPM_PACKAGE_NAME)` when `.enabled === true`. The existing entry-level cleanup (`install.ts:275-315`) will still complete the post-install uninstall as today.

### P2 — `package.json` version still `1.0.0`

Location: `package.json` (unchanged).

PR body acknowledges this is intentional ("merging will publish `@dev`"). Behavior is consistent with the release plan, and `--dev` users opt in. The minor friction: `cli/index.ts:31-66` `info` and `--version` will report `1.0.0` for a build that materially behaves like 2.0.0 (ClawHub install, deprecated `--next/--dev`, new doctor `--fix` path). Suggest pre-publishing as `2.0.0-dev.0` (or similar prerelease tag) so the version surface matches the documented behavior of the published `@dev` tarball. Non-blocking.

### P2 — `runUpdate` does not forward `from` / `force`

Location: `cli/update.ts:15-20`.

`runUpdate` constructs `runInstall({ dev, next })` only. The `update` subcommand in `cli/index.ts:93-104` does not declare `--from` or `--force` either, so this is internally consistent — but worth confirming intentional. `openclaw-channel-octo update --force` (matching the legacy install flag) silently no-ops on the flag because commander rejects unknown options after `update`. If the intent is "update is install when not installed and update when installed," then forwarding `--force` (re-install) at minimum mirrors what users likely expect. Non-blocking.

### P2 — Hidden `--dev` option only on `install`, not on `update`

Location: `cli/index.ts:82` vs `cli/index.ts:101`.

`install` exposes both `--next` (visible) and a hidden `--dev`. `update` also exposes the same pair. OK — consistent.

### P2 — `getLatestClawHubVersion` parses unbounded subprocess output

Location: `cli/install.ts:65-75` and `cli/install.ts:76-86`.

`runCmd` returns the full captured stdout, which is then `JSON.parse`d. `openclaw plugins inspect` output is bounded in practice, but no size guard exists. Acceptable for the local-CLI source; out of scope to harden further.

### P2 (positive note) — `checkForUpdates` SemVer-validates the remote `latestVersion`

Location: `src/channel.ts:296-315`.

The added `isSemverLike` regex `/^\d+\.\d+\.\d+(?:[-+][A-Za-z0-9.-]+)?$/` correctly rejects a hostile ClawHub response trying to smuggle newlines / terminal escape sequences into the log line before interpolation. This is the right defense-in-depth for a `security_sensitive` PR and the comment cross-references it.

### P2 — Stale-stage detection accepts either `NPM_PACKAGE_NAME` or `PLUGIN_ID`

Location: `cli/openclaw-cli.ts:910-920`.

Comment explains that `PLUGIN_ID`'s semantic shift from npm-name to ClawHub-id made the previous single-comparison dead code for staged npm installs; new check tolerates both. Correct fix.

### Nit — `cli/install.ts:436` skip-reinstall log

When the migration encounters `octoAlreadyHealthy === true`, the log line `"${PLUGIN_ID} already installed and healthy — skipping reinstall."` is good. Consider also logging the captured version (from `octoSnapshot.version`) for diagnostics.

### Nit — Deprecation warning size on `--dev`/`--next`

`cli/install.ts:151-157` emits a 4-line warn block on each invocation that uses `--dev` or `--next` without `--from`. Functionally fine; a single-line warning would be less noisy.

## Verification of prior review items

- **P0-1 (runtime spec)**: `index.ts:73-89, 91-113` — `handleInstall` and `handleUpdate` now pass `CLAWHUB_INSTALL_SPEC` (= `"clawhub:octo"`) to `pluginsInstall`, not `PLUGIN_ID`. Verified.
- **P0-2 (channels.octo preservation across legacy npm uninstall)**: `cli/install.ts:275-307` — saves `channels.octo` + `bindings(channel=octo)` *before* `pluginsUninstall(NPM_PACKAGE_NAME, true)`, restores them after. Regression test at `cli/install.test.ts:485-574` exercises the side-effecting uninstall and asserts the user's `botToken` + binding survive. Verified.
- **Doctor `--fix` defers to `runInstall({force:true})`** (`cli/doctor.ts:185-218`): correctly avoids the partial subset of recovery paths that broke after `PLUGIN_ID` semantics changed. Verified.

## Security considerations (security_sensitive)

- **Hardcoded install spec** (`CLAWHUB_INSTALL_SPEC = "clawhub:octo"`): no user input flows into `pluginsInstall` other than the `--from` override, which is explicitly documented for local-tarball / alt-spec testing. Acceptable.
- **`runCmd` uses `execFileSync` with array args**: no shell expansion on POSIX. Windows path uses `cmd.exe /d /v:off /c call` with `escapeCmdArg` quoting `&|<>^()"` + `%`. Reasonable.
- **Direct JSON file writes** to `openclaw.json` (`writeConfigAtomic` + `.bak` sidecar): standard backup pattern, atomic via tmp+rename inside writer.
- **No new secret material introduced**: bot tokens (`bf_...`) and IM tokens flow through existing paths and were not touched.
- **ClawHub HTTP fetch** in `src/channel.ts:295`: hardcoded `https://clawhub.ai/api/v1/packages/octo`, no user-controlled URL components. SemVer validation on the returned `latestVersion` before any log interpolation.

Suggest human verification:
1. Confirm OpenClaw's actual behavior when two plugins (`openclaw-channel-octo` v1.x npm + ClawHub `octo`) both declare `channel.id = "octo"` during the rebrand-path install — does step 5 fail, or proceed?
2. Confirm npm `@dev` dist-tag rollout plan: `npx openclaw-channel-octo install` resolves `@latest`, which will still be v1.x post-merge, so the new behavior reaches users only when they explicitly install `@dev` or after a future v2.0.0 publish. This matches the PR's stated intent, but is worth re-confirming with downstream stakeholders.

lml2468 (approved):
## Re-review — PR #37 (ClawHub install + legacy npm cleanup)

**Verdict: APPROVED** — all 3 previous blockers resolved, plus several bonus hardening fixes.

### Previous blockers — all resolved ✅

| # | Finding | Resolution |
|---|---|---|
| 1 | `pluginsUninstall(NPM_PACKAGE_NAME)` wipes `channels.octo` | ✅ `saveChannelConfigFromFile` + `saveBindingsFromFile` before uninstall, `restoreChannelConfigToFile` + `restoreBindingsToFile` after. Mirrors the existing `runMigration` pattern. |
| 2 | `doctor.ts:241` normal string `${PLUGIN_ID}` not interpolated | ✅ Changed to template literal with escaped backticks |
| 3 | `doctor.ts:250` same bug | ✅ Fixed |

### Bonus fixes in this push

1. **Runtime slash commands** (`index.ts:81,105`): `handleInstall` now uses `CLAWHUB_INSTALL_SPEC` instead of `PLUGIN_ID`; `handleUpdate` uses `getLatestClawHubVersion()` instead of `npm view`. Both paths now correctly target ClawHub.

2. **SemVer sanitization** (`channel.ts:299-308`): version check response from ClawHub is validated against `/^\d+\.\d+\.\d+.../` before logging — prevents terminal escape injection from a hostile or corrupted response. Good security-sensitive thinking.

3. **Stale dir cleanup** (`openclaw-cli.ts:912`): `cleanupStaleStageDirectories` now matches both `NPM_PACKAGE_NAME` and `PLUGIN_ID`, preventing dead code for npm-sourced staged dirs after the PLUGIN_ID semantic shift.

4. **Regression test** (`install.test.ts:484-575`): excellent test — simulates uninstall removing `channels.octo` as a side effect, then asserts the save/restore preserves bot tokens and bindings. This is exactly the test that was missing in the first round.

### Verification

- CI green (type-check + build + test all pass)
- `doctor.ts` lines 241 and 250 now use template literals — confirmed on head branch
- `install.ts` save/restore pattern matches `runMigration` architecture
- Runtime `handleInstall`/`handleUpdate` both use `CLAWHUB_INSTALL_SPEC`

PR #38, opened 2026-05-19
fix(ci): replace paths-ignore with dorny/paths-filter preflight job

lml2468:
## Summary

Follow-up fix after cross-repo review identified the same root cause in all repos.

**Root cause:** workflow-level `paths-ignore` on the `pull_request` trigger prevents GitHub from creating any check run when only docs/markdown files change. If branch protection requires `Build` / `Test` / etc. as status checks, a docs-only PR is permanently stuck — the required check never appears.

**Fix:**
- Removed `paths-ignore` from `pull_request` trigger
- Added a `changes` preflight job using `dorny/paths-filter` (SHA-pinned)
- All downstream jobs `needs: [changes]` and gate on `needs.changes.outputs.code == 'true'`
- Docs-only PRs: downstream jobs show `skipped` (GitHub counts this as passing for required checks ✅)
- Push to main: paths-filter step is skipped; output defaults to `'true'` so all jobs run ✅
- `push` trigger retains `paths-ignore` (not a required-check path, saves runner minutes)
- Also included: `labeler.yml` fix — `pull_request_target` instead of `pull_request` for fork PR write access

## Related
- octo-server PR #84 (same fix, already reviewed and merged)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,10 @@ name: CI
 on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
+    # converted_to_draft kept: creates "skipped" check entries satisfying required-check rules
     branches: [main]
-    paths-ignore:
-      - "**/*.md"
-      - "docs/**"
-      - ".github/ISSUE_TEMPLATE/**"
+    # paths-ignore intentionally removed — use changes preflight job instead
+    # (workflow-level path filters deadlock required status checks for docs-only PRs)
   push:
     branches: [main]
     paths-ignore:
@@ -23,8 +22,35 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: github.event_name == 'pull_request'
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          filters: |
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
   build:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Build & Type-check (${{ matrix.package }})
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -26,4 +26,3 @@ jobs:
       pull-requests: write
     # secrets: intentionally omitted — reusable-pr-labeler.yml uses only the
     # inherited GITHUB_TOKEN granted via the job-level permissions above.
-


yujiawei (approved):
# Code Review — PR #38 (octo-adapters)

## Verdict: APPROVED

Clean port of the octo-server #84 fix. The root-cause analysis is correct and the implementation matches the cross-repo precedent. Two minor PR-body / metadata observations below — none block merge.

---

## 1. Verified correctness

### Root cause (validated)
`paths-ignore` on a workflow-level `pull_request` trigger prevents GitHub from instantiating a workflow run when the change set is entirely excluded by the filter. No run → no `Build` check → required-check rules deadlock docs-only PRs. Hoisting the path decision into an in-workflow `changes` preflight job and gating downstream jobs with `if:` produces a *skipped* job, which GitHub counts as a passing required check. This is the documented, GitHub-recommended pattern.

### `changes` job output expression (`ci.yml:32`)
```yaml
outputs:
  code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
```
- `push` event → short-circuits to the string `'true'` (right side never evaluated, so the skipped `filter` step is harmless).
- `pull_request` event → forwards `steps.filter.outputs.code` (`'true'` / `'false'` from dorny/paths-filter v3).
Both legs produce string literals comparable to `'true'` in the consumer. ✅

### `build` gate (`ci.yml:51-53`)
```yaml
if: |
  (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
  needs.changes.outputs.code == 'true'
```
| Event                           | clause 1 | clause 2     | result   |
|---------------------------------|----------|--------------|----------|
| push to main                    | true     | `'true'==…`  | runs     |
| PR (ready), code changed        | true     | `'true'==…`  | runs     |
| PR (ready), docs-only           | true     | `'false'==…` | skipped  |
| PR (draft) / converted_to_draft | false    | —            | skipped  |

All four states produce the right outcome, and "skipped" is the correct terminal state for required-check satisfaction.

### Filter semantics (`ci.yml:42-47`)
```
code:
  - '**'
  - '!**/*.md'
  - '!docs/**'
  - '!.github/ISSUE_TEMPLATE/**'
```
dorny/paths-filter uses picomatch ordered patterns: include-all then exclude. This is a faithful mirror of the original workflow-level `paths-ignore`. ✅

### Security / hygiene
- Both third-party actions SHA-pinned with `# v4` / `# v3` version comments (`ci.yml:34, 38`). ✅
- `persist-credentials: false` on the checkout (`ci.yml:37`). ✅
- `changes` job uses the minimal-needed permissions (`contents: read`, `pull-requests: read`) — `pull-requests: read` is required because dorny/paths-filter v3 uses the REST API to compute the diff for PR events without a full checkout. ✅
- `concurrency` group is unchanged and still keys on `pull_request.number || github.ref`, so per-PR cancellation continues to work for the extra job. ✅

### Trigger consistency
- `pull_request` trigger correctly drops `paths-ignore` (this is the whole fix).
- `push` trigger retains `paths-ignore` (acceptable — `push` is not subject to required-check rules and the saved runner minutes are real).
- `converted_to_draft` retained with inline rationale (`ci.yml:5-6`). The mechanism: on convert-to-draft, the build's draft clause short-circuits to false, producing a skipped check entry — exactly the desired "always emit a check entry" property.

### labeler.yml change
The actual diff is the removal of a single trailing blank line (`labeler.yml:29`). Behavior unchanged. The `pull_request_target` trigger described in the PR body is already on `main`; see PR-body note below.

---

## 2. Issues found

None at P0/P1. Two P2 / nit items:

### P2 — PR body line about `labeler.yml` is misleading
> *"Also included: `labeler.yml` fix — `pull_request_target` instead of `pull_request` for fork PR write access"*

Comparing `main` ↔ head, the labeler.yml hunk is just a trailing-newline deletion. The `pull_request_target` trigger is already on `main` from a previous commit. This claim should be removed from the PR description to avoid confusing reviewers / changelog authors who rely on the PR body.

### P2 — One-line whitespace churn in labeler.yml
Deleting a final newline from a YAML file (`labeler.yml:29`) is technically a no-op for behavior but counter-recommended by POSIX (`text file` should end with a newline) and most editor/linter defaults. Consider either reverting this hunk so the PR is single-file, or adding a `.editorconfig` / pre-commit step in a follow-up so the file's final-newline policy is enforced consistently. Not a blocker.

---

## 3. Recommendations

1. **Edit the PR description** to drop the `labeler.yml` `pull_request_target` line, or rewrite it to reflect the actual diff (trailing-newline normalization).
2. **(Optional, follow-up)** If you intend `labeler.yml` to *not* end with a newline, add that as an explicit `.editorconfig` rule for `*.yml`; otherwise restore the trailing newline.

---

## 4. Additional observations

- The repository's `main` branch currently has **no branch protection** (`/repos/.../branches/main/protection` returns 404). The fix is still correct — and clearly preventive for when protection is enabled — but the immediate "deadlocked PR" symptom does not exist on this repo today. This matches the PR framing as a cross-repo precaution mirroring octo-server #84.
- No other downstream jobs exist in `ci.yml`; only `build` needs `needs: [changes]`. If a future job (e.g. lint, release-please) is added, remember to gate it with the same `needs: [changes]` + `needs.changes.outputs.code == 'true'` pattern, otherwise that new required check will reintroduce the deadlock for docs-only PRs.
- For the `push` event leg, `steps.filter` is correctly skipped via `if: github.event_name == 'pull_request'`. If anyone later adds `workflow_dispatch` or a `schedule` trigger, the output expression's left-hand side (`github.event_name != 'pull_request'`) will continue to short-circuit to `'true'`, so `build` will still run — current behavior is forward-safe.

Jerry-Xin (dismissed):
This PR is relevant to `octo-adapters`, but the CI change does not reliably fix the required-check deadlock it is intended to solve.

🔴 Blocking

🔴 Critical — `.github/workflows/ci.yml:51`: the code-change gate is applied at the matrix job level. GitHub evaluates `jobs.<job_id>.if` before `strategy.matrix` is applied, while the matrix is what creates the package-specific check runs. ([docs.github.com](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax)) When a docs-only PR makes `needs.changes.outputs.code == 'false'`, the `build` matrix may never expand, so required checks such as `Build & Type-check (openclaw-channel-dmwork)` and `Build & Type-check (openclaw-channel-octo)` can still be missing instead of skipped. That reintroduces the same branch-protection deadlock this PR is meant to fix.

Move the docs/draft gating below matrix expansion. For example, keep `needs: [changes]`, let the matrix job always instantiate, add a cheap no-op step for skipped cases, and put the code/draft condition on the expensive checkout/setup/install/type-check/build/test steps.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/labeler.yml`: the actual diff only removes a trailing blank line; the `pull_request_target` behavior already exists on `main`. Consider trimming the PR description so reviewers do not spend time looking for a trigger change that is not part of this PR.

✅ Highlights

The preflight job uses pinned actions, minimal permissions, and preserves `push` path filtering, which is a good direction once the matrix check-name issue is fixed. `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12` reported no workflow syntax issues.

lml2468 (commented):
## Independent Review — PR #38 (paths-filter preflight)

**Verdict: APPROVED** (submitted as COMMENT — GitHub blocks self-approve)

### Phase 1: Gate

CI green, mergeable. +31/-6, 2 files. `Detect changed paths` job runs and passes.

### On @Jerry-Xin's blocker — I **disagree**

Allen claims that when `build` job's `if:` evaluates to false, the matrix check names will be "missing instead of skipped," causing required-check deadlock. This is factually incorrect per GitHub's own documentation:

> **"A job that is skipped will report its status as 'Success'. It will not prevent a pull request from merging, even if it is a required check."**
> — [GitHub Docs: About status checks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks)

When a job-level `if:` evaluates to false, GitHub **creates check entries in the "skipped" state**, which counts as success for required-check purposes. This is different from workflow-level `paths-ignore`, which prevents the workflow from triggering at all (no check entries created → deadlock).

The whole point of this PR is replacing `paths-ignore` (which causes missing checks) with `dorny/paths-filter` (which causes skipped checks). The distinction is:

| Mechanism | Check entries | Required check effect |
|---|---|---|
| `paths-ignore` (workflow-level) | ❌ Not created | 💀 Deadlock |
| Job-level `if: false` | ✅ Created as "skipped" | ✅ Passes |

This is also confirmed by:
- [devopsdirective.com/posts/2025/08/github-actions-required-checks-for-conditional-jobs/](https://devopsdirective.com/posts/2025/08/github-actions-required-checks-for-conditional-jobs/) (Option 1 explicitly relies on this behavior)
- The related `octo-server PR #84` using the same pattern, already merged

### Analysis of the actual change

1. **`changes` preflight job** — clean. SHA-pinned `dorny/paths-filter@6852f92c`. Minimal permissions (`contents: read`, `pull-requests: read`). Only runs checkout on PRs, not pushes.

2. **Output default for pushes** — `code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}` — correct. On `push` events, the filter step is skipped and the output defaults to `'true'`, so all jobs run on push-to-main.

3. **`paths-ignore` kept on `push` trigger** — correct. Push events don't create required checks, so `paths-ignore` is safe there and saves runner minutes.

4. **`converted_to_draft` comment** — helpful documentation of why it's kept.

5. **`labeler.yml`** — trivial trailing newline removal, no behavior change.

### 🟡 Non-blocking

1. The `persist-credentials: false` on the checkout is good practice but worth noting that `dorny/paths-filter` v3 can also work without checkout (using the GitHub API via `token` input). Not a concern here since the checkout is already minimal.

2. The `build` job's `if:` combines both the draft check and the code-change check in one expression. This is fine but means a draft PR with code changes also won't have the filter step output — however, that's correct since draft PRs should skip CI regardless.

### ✅ Highlights

- The root cause analysis is precise: `paths-ignore` → no workflow trigger → no check entries → required check deadlock
- The fix pattern (preflight filter job → job-level `if:`) is the GitHub-recommended approach
- Push behavior preserved with `paths-ignore` (saves runner time without deadlock risk)
- SHA-pinned actions throughout

Jerry-Xin (approved):
Withdrawing my earlier CHANGES_REQUESTED — the blocker was a misjudgment.

Job-level `if: false` correctly creates "Skipped" check entries for all matrix combinations, satisfying branch protection required checks. The key distinction is workflow-level `paths-ignore` (workflow never triggers → no check entries → deadlock) vs job-level `if` condition (workflow triggers, job skipped → check entries exist as Skipped → passes). `octo-server #84` validates the same pattern.

PR is clean: SHA-pinned actions, push retains `paths-ignore` for runner savings, output defaults handle push events correctly. No blockers.

PR #27, opened 2026-05-16
fix(inbound): prevent bot from re-answering previously answered questions in group chat

ken-anderson2006:
## Summary

Fixes #23

When users ask consecutive questions in a group chat with `@Bot`, the bot re-answers previously answered questions along with the new one. This PR fixes the duplicate reply issue by introducing **message_seq-based history segmentation**.

## Root Cause

1. `@Bot` messages were **not cached** in `groupHistories` (only non-mentioned messages were pushed)
2. When in-memory cache had fewer entries than `historyLimit/2`, the adapter fetched history from the DMWork/Octo API as backfill
3. The backfill filter excluded bot messages but **did not exclude already-answered `@Bot` messages** (e.g., Q1)
4. This caused Q1 to appear in `prependContext` without its answer A1, while the session transcript contained both Q1+A1 — confusing the model into re-answering Q1

## Solution

- Track `lastBotReplySeq` (the `message_seq` of the last message the bot replied to) per group session
- When building history context, partition messages into **answered** (`message_seq <= cutoffSeq`) and **new** (`message_seq > cutoffSeq`)
- Annotate answered messages with `[ALREADY ANSWERED – do NOT re-answer]` markers in the prompt
- Add `message_id` and `message_seq` fields to the API fetch response for proper tracking
- Exclude the current in-processing message from history to avoid self-duplication

## Changes

- `src/inbound.ts`: Add history segmentation logic with `partitionHistoryByCutoff()`, annotate answered context
- `src/channel.ts`: Add `lastBotReplySeqMap` module-level cache with proper cleanup
- `src/api-fetch.ts`: Return `message_id` and `message_seq` from sync API
- `src/inbound.test.ts`: Add comprehensive tests for the partitioning and annotation logic

PR #40, opened 2026-05-19
feat(ci): add CodeQL security scan and Workflow Sanity check

lml2468:
### `codeql.yml` (new)
TypeScript/JavaScript CodeQL security analysis via shared reusable workflow.
- queries: `security-extended`
- Triggers: push to main / PR opened-synced-reopened / daily cron 06:00 UTC / manual
- Skip draft PRs and docs-only changes
- Concurrency group prevents parallel duplicate runs

### `workflow-sanity.yml` (new)
Validates workflow quality on CI file changes:
- No-tabs scan (Python3 across \*.yml + \*.yaml)
- actionlint v1.7.12 static analysis (SHA-256 verified download)

## Merge order
**Merge `Mininglamp-OSS/.github` PR #26 first** — `reusable-codeql.yml` and `workflow-sanity.yml` must be on `main` before these callers work.

## Part of P1 automation rollout

PR #26, opened 2026-05-16
chore(octo): remove shim package and clean up references

caster-Q:
## Summary

Remove the planned `openclaw-channel-dmwork` shim package. The shim was
designed to silently redirect `npx -y openclaw-channel-dmwork install`
to `openclaw-channel-octo`, but this implicit redirection was considered
confusing — users would not know their plugin had changed, making
troubleshooting harder.

Migration is now explicit: BotFather (Phase C) guides users directly to
`npx -y openclaw-channel-octo install`, and the install command handles
rebrand migration from any existing dmwork configuration automatically.

## Related Issue

Design decision follow-up to #17 (Phase B).

## Changes

- `openclaw-channel-octo/shim-stub/` — deleted (shim package template)
- `openclaw-channel-octo/scripts/build-shim.mjs` — deleted
- `openclaw-channel-octo/scripts/build-shim.test.ts` — deleted
- `.github/workflows/publish-octo.yml` — removed `publish_shim` job; updated sequence comment
- `.github/workflows/publish-octo-dev.yml` — updated comment (no shim reference)
- `openclaw-channel-octo/CHANGELOG.md` — replaced shim description with accurate no-shim note
- `openclaw-channel-octo/cli/exports.test.ts` — updated comment: `./cli` is a public subpath API, not a shim entry point

## Testing

- `npm run type-check` → 0 errors
- `npm test` → 25 files / 718 tests pass (3 shim-specific tests removed with the files)

## Checklist

- [x] I have read [CONTRIBUTING.md](.github/CONTRIBUTING.md)
- [x] PR description is in English
- [x] Followed commit message conventions (Conventional Commits)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #41, opened 2026-05-19
feat(ci): add release drafter and gated release publish

lml2468:
### `release-drafter.yml` (new)
Automatically maintains a GitHub Release draft as PRs merge to main.
- Triggers on every push to main + manual dispatch
- Delegates to shared reusable workflow in .github repo
- Uses per-repo `.github/release-drafter.yml` config

### `release-publish.yml` (new)
Gate-controlled manual release publisher via workflow_dispatch.
Inputs: tag (required), validate_run_id (optional CI gate), draft flag.

### `.github/release-drafter.yml` (new)
Release drafter category config (Features / Bug Fixes / CI & Infrastructure / Documentation).

## Merge order
Requires Mininglamp-OSS/.github PR #27 merged first.

## Part of P1 automation rollout

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,20 @@
+# Release Drafter config — https://github.com/release-drafter/release-drafter
+name-template: "$NEXT_PATCH_VERSION"
+tag-template: "v$NEXT_PATCH_VERSION"
+template: |
+  ## What's Changed
+
+  $CHANGES
+
+  **Full Changelog**: https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION
+categories:
+  - title: "Features"
+    labels: [enhancement, feature]
+  - title: "Bug Fixes"
+    labels: [bug, fix]
+  - title: "CI and Infrastructure"
+    labels: [ci, infra, dependencies-changed]
+  - title: "Documentation"
+    labels: [documentation, docs]
+exclude-labels: [skip-changelog]
+change-template: "- $TITLE @$AUTHOR (#$NUMBER)"
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,15 @@
+name: Release Drafter
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  draft:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-release-drafter.yml@main
+    permissions:
+      contents: write
+      pull-requests: read
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -1,0 +1,31 @@
+name: Release Publish
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to publish (e.g. v1.4.0)"
+        required: true
+        type: string
+      validate_run_id:
+        description: "Required: successful CI run ID from the tagged commit as release evidence"
+        required: true
+        type: string
+      draft:
+        description: "Keep as draft instead of publishing"
+        required: false
+        type: boolean
+        default: false
+
+permissions: {}
+
+jobs:
+  publish:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-release-publish.yml@main
+    with:
+      tag: ${{ inputs.tag }}
+      validate_run_id: ${{ inputs.validate_run_id }}
+      draft: ${{ inputs.draft }}
+    permissions:
+      contents: write
+      actions: read


yujiawei (changes requested):
# Code Review — PR #41 (octo-adapters)

## Summary

This PR wires three new CI surfaces into octo-adapters via reusable workflows hosted in `Mininglamp-OSS/.github`: a Release Drafter, a gated Release Publisher, plus (undeclared in the PR body) CodeQL and a Workflow Sanity check. The intent and structure are sound — locked-down `permissions: {}` at the top level, narrow per-job grants, pinned action SHAs upstream, and a clean delegation to reusables. However, two correctness issues will cause workflows to fail at runtime as currently written, and the declared dependency set is incomplete. Requesting changes before merge.

---

## Blocking issues

### P0-1 — `workflow-sanity.yml` cannot be called via `uses:`; the upstream is not a reusable workflow

`.github/workflows/workflow-sanity.yml:17-19`
```yaml
jobs:
  sanity:
    uses: Mininglamp-OSS/.github/.github/workflows/workflow-sanity.yml@main
```

The referenced upstream file (added by `Mininglamp-OSS/.github` PR #26) is triggered by:
```yaml
on:
  pull_request: { paths: [...] }
  push:         { branches: [main], paths: [...] }
  workflow_dispatch:
```

It does **not** declare `on.workflow_call:`. Per the GitHub docs ("Reusing workflows → Calling a reusable workflow"), a workflow without a `workflow_call` trigger cannot be invoked via `uses:` — Actions will reject the caller with `error parsing called workflow … must contain a workflow_call trigger`. As soon as this caller fires (which the path filters guarantee on this very PR), the `Workflow Sanity` check will error out, and every subsequent PR that touches `.github/workflows/**` in octo-adapters will fail the same way.

Additionally, workflows living in `Mininglamp-OSS/.github/.github/workflows/` are **not** automatically propagated to org repos (only files under `.github/workflow-templates/` are templated, and those are copy-on-create, not live shared). So merging PR #26 alone does not give octo-adapters that check; it only runs on the `.github` repo itself.

**Fix options (pick one):**
- Inline the sanity job logic directly in `octo-adapters/.github/workflows/workflow-sanity.yml` (copy the `no-tabs` Python step and the pinned `actionlint` install from PR #26).
- Or, in `Mininglamp-OSS/.github` PR #26, add `on.workflow_call:` to that workflow (and ideally rename it to `reusable-workflow-sanity.yml` to match the naming convention already established by `reusable-release-drafter.yml` / `reusable-release-publish.yml` / `reusable-codeql.yml`), then keep this caller.

The second option is preferable because it keeps the implementation single-sourced across repos.

### P0-2 — Dependency declaration is incomplete; PR also requires `.github` PR #26

The PR body says:

> Requires Mininglamp-OSS/.github PR #27 merged first.

But this PR references four upstream workflows:

| Caller file | Upstream `uses:` | Source PR |
|---|---|---|
| `codeql.yml` | `reusable-codeql.yml@main` | **.github PR #26** (not #27) |
| `release-drafter.yml` | `reusable-release-drafter.yml@main` | .github PR #27 |
| `release-publish.yml` | `reusable-release-publish.yml@main` | .github PR #27 |
| `workflow-sanity.yml` | `workflow-sanity.yml@main` | **.github PR #26** (and see P0-1) |

If only PR #27 lands and PR #41 merges, the `CodeQL` workflow will fail on the first `push` to main and `workflow-sanity.yml` will fail on every PR touching `.github/**`, breaking required checks. Please update the PR body to require **both** `.github` #26 and #27, and consider holding this PR behind a merge-queue note ("do not merge until both upstream PRs are in").

---

## Non-blocking findings

### P1 — PR description / scope mismatch

The PR description enumerates three changes (`release-drafter.yml`, `release-publish.yml`, `.github/release-drafter.yml`) but the diff includes **five** new files — `codeql.yml` (31 LOC) and `workflow-sanity.yml` (19 LOC) are silently added. CodeQL in particular is a meaningful security surface that reviewers should be told about explicitly. Please update the PR body to list every workflow being introduced and (briefly) why CodeQL was bundled with the release-automation rollout rather than landed in its own PR.

### P2 — `release-drafter.yml` config: tag/name template asymmetry

`.github/release-drafter.yml:2-3`
```yaml
name-template: "$NEXT_PATCH_VERSION"
tag-template:  "v$NEXT_PATCH_VERSION"
```

The release name will render as e.g. `1.4.0` while the tag is `v1.4.0`. This is functional but inconsistent — most repos either prefix both or strip both. Recommend `name-template: "v$NEXT_PATCH_VERSION"` so the release title matches the tag in the GitHub Releases UI and in CHANGELOG export tooling.

### P2 — `categories` references labels that may not exist in this repo

`.github/release-drafter.yml:15`
```yaml
- title: "CI and Infrastructure"
  labels: [ci, infra, dependencies-changed]
```

`dependencies-changed` is a non-default label and (based on `gh label list` defaults) is not currently defined in octo-adapters. Release-drafter silently ignores categories whose labels never match, so this is not a correctness bug, but if the intent is to surface Renovate/Dependabot PRs under CI & Infrastructure you'll want to either (a) define the label and have Renovate apply it, or (b) reference whatever label the dependency bot actually sets (typically `dependencies` for Dependabot). Worth aligning across all org repos to keep the org-wide release notes consistent.

### P2 — `codeql.yml` concurrency group has a redundant prefix

`.github/workflows/codeql.yml:18`
```yaml
group: codeql-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.sha }}
```

`github.workflow` already evaluates to `CodeQL`, so the group expands to `codeql-CodeQL-…`. Drop the literal `codeql-` prefix or drop `${{ github.workflow }}` — keeping both does no harm but reads like a copy-paste leftover.

### P2 — `codeql.yml` `paths-ignore` could mirror the org pattern

`.github/workflows/codeql.yml:7-11` excludes `**/*.md`, `docs/**`, and `.github/ISSUE_TEMPLATE/**`. Most TypeScript repos also exclude `**/*.snap`, generated `dist/**`, and `*.lock` so that scheduled CodeQL runs do not waste minutes on diff-noise paths. Optional polish; not required for merge.

### Nit — language identifier

`language: javascript-typescript` is correct for the modern combined CodeQL pack (the upstream reusable in `.github` PR #26 also documents `go` or `javascript-typescript`). No action needed — flagging for future readers who may remember the older `javascript` / `typescript` split.

---

## What looks good

- Locked-down `permissions: {}` at workflow top level; per-job grants are narrow and correctly scoped (`security-events: write` only on CodeQL, `contents: write` only on release publisher / drafter).
- `release-publish.yml` validates semver and an optional CI `run_id` in the reusable before publishing — sensible gate design.
- Manual-dispatch-only publish path; no accidental publish on tag push or release-drafter close.
- Reusable workflows in the dependency PRs pin third-party actions by SHA (`release-drafter@67e173ca…`, `codeql-action@7c1e4cf0…`, `actions/checkout@34e11487…`). Good supply-chain hygiene; PR #41 inherits this transitively.
- `concurrency.cancel-in-progress` is correctly limited to PR runs on CodeQL; main-branch runs are preserved.

---

## Suggested next steps

1. Fix P0-1 by either inlining the workflow-sanity logic in this repo, or adding `on.workflow_call:` to the upstream and updating this caller.
2. Update the PR body to (a) declare both `.github` #26 and #27 as required-merged-first, and (b) list the five files actually added (especially CodeQL).
3. Optional but worth doing in a follow-up: align `release-drafter.yml` `name-template` with the `tag-template` (`v` prefix), and confirm the `dependencies-changed` label is what the dependency bot actually emits.

Once the P0 items are addressed, this is ready to merge alongside the upstream PRs.

Jerry-Xin (changes requested):
Project relevance passes, but the new CI/release workflow wiring has blocking execution and permission issues.

**🔴 Blocking**
- 🔴 Critical: `.github/workflows/codeql.yml:25` calls `Mininglamp-OSS/.github/.github/workflows/reusable-codeql.yml@main`, and `.github/workflows/workflow-sanity.yml:17` calls `Mininglamp-OSS/.github/.github/workflows/workflow-sanity.yml@main`, but those reusable workflows are not present on `Mininglamp-OSS/.github@main`. The stated dependency PR #27 only adds the release drafter/publish reusable workflows, so these two workflows will fail to resolve after merge unless another upstream dependency is added first or these callers are removed/retargeted. ([api.github.com](https://api.github.com/repos/Mininglamp-OSS/.github/contents/.github/workflows?ref=main))
- 🔴 Critical: `.github/workflows/release-publish.yml:24-31` exposes `validate_run_id`, but the caller grants only `contents: write`. Reading `repos/{owner}/{repo}/actions/runs/{run_id}` requires Actions read permission for the token, so the CI validation gate will fail when `validate_run_id` is supplied. Add `actions: read` here, and ensure the reusable workflow also requests/preserves that permission.

**💬 Non-blocking**
- 🟡 Warning: `.github/workflows/release-publish.yml:10-19` keeps the CI gate optional and defaults `draft` to `false`. That may be intentional, but it means a manually dispatched release can publish without evidence validation unless operators always provide `validate_run_id`.

**✅ Highlights**
- ✅ The release drafter config is small and repo-local.
- ✅ Top-level `permissions: {}` with job-scoped permissions is the right baseline pattern.

lml2468 (commented):
## Independent Review — PR #41 (release drafter + gated publish + CodeQL + sanity)

**Verdict: CHANGES_REQUESTED** (COMMENT due to self-PR restriction)

This PR is a superset of PR #40: adds `codeql.yml` + `workflow-sanity.yml` (same as #40) plus `release-drafter.yml`, `release-publish.yml`, and `.github/release-drafter.yml`.

### 🔴 Blocking

**1. Undeclared upstream dependency — PR #26 required but not listed**

The PR body declares dependency on `Mininglamp-OSS/.github` PR #27 only (release drafter/publish reusable workflows). But this PR also includes `codeql.yml` and `workflow-sanity.yml` which call reusable workflows from PR **#26** (`reusable-codeql.yml` + `workflow-sanity.yml`). Neither of these exist on `@main` yet:

```
$ gh api repos/Mininglamp-OSS/.github/contents/.github/workflows?ref=main --jq '.[].name'
→ no reusable-codeql.yml, no workflow-sanity.yml
```

After merging this PR (even after PR #27 merges), CodeQL and Sanity workflows will fail with "workflow file not found."

**2. `workflow-sanity.yml` upstream missing `workflow_call` trigger** (same as PR #40)

Even when PR #26 merges, the upstream `workflow-sanity.yml` only declares `pull_request` / `push` / `workflow_dispatch` — no `workflow_call:`. The `uses:` call will fail at runtime.

**3. `release-publish.yml` missing `actions: read` for CI validation gate**

Concur with @Jerry-Xin. The `validate_run_id` step calls `gh api "repos/.../actions/runs/${RUN_ID}"` which requires `actions: read` permission. The caller grants only `contents: write`:

```yaml
# release-publish.yml:30-31
permissions:
  contents: write  # ← missing actions: read
```

The reusable workflow's job also only declares `contents: write`. Both need `actions: read` added for the CI gate to work.

### 🟡 Non-blocking

1. **`workflow-sanity.yml` push trigger path asymmetry** — `pull_request` watches `workflows/**` + `actions/**`, but `push` only watches `workflows/**`. Mirror both paths.

2. **`release-publish.yml` optional CI gate** — `validate_run_id` defaults to empty (no validation). This means manual releases can skip CI evidence. Consider documenting this as intentional or making it required with a separate `skip_validation` flag if the intent is always-validate-with-escape-hatch.

3. **`release-drafter.yml` config** — the `$NEXT_PATCH_VERSION` template always bumps patch. For a mono-repo with two packages, this is fine as a starting point. May want to document that semver-major/minor bumps require manual tag creation + publish.

### ✅ Highlights

- `release-drafter.yml` config is clean — categories map to existing label taxonomy
- `contents: write` + `pull-requests: read` for drafter is correct least-privilege
- `workflow_dispatch` on drafter enables manual re-sync
- Top-level `permissions: {}` on all new files

lml2468:
## Fix pushed — synced with updated reusable-release-publish.yml

**validate_run_id** changed from optional to required (matches upstream reusable update).
**actions: read** permission added (needed by reusable to query run status and head_sha).

**Why validate_run_id is now required:**
The reusable now cross-checks that the CI run was triggered from the same commit as the release tag. An optional gate could be bypassed by omitting the run_id entirely, defeating the purpose of the gated release design.

Jerry-Xin (dismissed):
Summary: The PR is in scope for this repository and the workflow additions are generally aligned with the repo’s CI/release automation direction.

💬 Non-blocking

🟡 Warning: `.github/workflows/release-drafter.yml:12` and `.github/workflows/release-publish.yml:24` depend on reusable workflows in `Mininglamp-OSS/.github@main`. The release reusable workflows are not available unless the noted dependency PR is merged first, so this PR should not land before that merge order is satisfied.

🟡 Warning: `.github/workflows/codeql.yml:8` uses PR-level `paths-ignore`. That is fine if CodeQL is not a required check, but this repo’s existing CI explicitly avoids workflow-level PR path filters because skipped workflows can deadlock required checks on docs-only PRs. If CodeQL may become required, mirror the existing `ci.yml` preflight pattern instead.

🔵 Suggestion: `.github/workflows/workflow-sanity.yml:8-11` includes `.github/actions/**` for PRs but not for pushes to `main`. If local actions are added later, direct main updates to those actions would skip the sanity workflow.

🔵 Suggestion: `.github/workflows/release-publish.yml:22` could add a concurrency group keyed by `inputs.tag` to avoid two manual dispatches racing to create or update the same release.

✅ Highlights

The release publisher keeps permissions scoped to `contents: write` and `actions: read`, and delegates validation to the shared reusable workflow.

The release drafter configuration is simple and readable, with clear categories and an explicit `skip-changelog` escape hatch.

Local workflow syntax validation with `actionlint` completed without findings.

lml2468 (commented):
## Re-review — PR #41 (release drafter + gated publish + CodeQL + sanity)

**Verdict: APPROVED** (COMMENT due to self-PR restriction)

### Previous blockers — all resolved ✅

| # | Finding | Resolution |
|---|---|---|
| 1 | `workflow-sanity.yml` upstream missing `workflow_call` | ✅ PR #26 merged (`2026-05-19T11:33:10Z`). Upstream now declares `workflow_call:` + push paths include `actions/**` |
| 2 | `reusable-codeql.yml` not on `@main` | ✅ PR #26 merged. Both `reusable-codeql.yml` and `workflow-sanity.yml` confirmed on `@main` |
| 3 | `release-publish.yml` missing `actions: read` | ✅ Added at line 31. Also hardened: `validate_run_id` changed from optional (default `""`) to **required** — no CI-less releases possible |

### Merge-ordering note

PR #27 (`reusable-release-drafter.yml` + `reusable-release-publish.yml`) is still OPEN. This PR must not merge until #27 lands, or the `release-drafter.yml` and `release-publish.yml` callers will fail with "workflow file not found." The PR body already documents this dependency.

### Verification

- PR #26 merged: confirmed `workflow-sanity.yml` has `workflow_call:` on `@main`
- PR #26 merged: confirmed `reusable-codeql.yml` exists on `@main` with `workflow_call:` + correct inputs
- `release-publish.yml` on head branch: `actions: read` present, `validate_run_id` is `required: true`
- CI green

lml2468:
## Fix pushed — cleaned up extra files + synced with .github#27

**What changed:**
1. Removed `codeql.yml` and `workflow-sanity.yml` — these belong to a separate PR (#88/#54 etc.), were accidentally included due to branch inheritance
2. Release PRs now only contain: `release-drafter.yml`, `release-publish.yml`, `.github/release-drafter.yml`
3. Upstream `.github#27` was fixed for annotated tag SHA resolution

**Merge order:** `.github` PR #27 must merge first (contains reusable-release-drafter.yml and reusable-release-publish.yml)

yujiawei (approved):
# Code Review — PR #41 (octo-adapters)

## Summary

This PR adds a release-automation pair to octo-adapters: a Release Drafter that maintains a rolling draft as PRs land on `main`, and a manually dispatched, gated Release Publisher. Both delegate to reusable workflows in `Mininglamp-OSS/.github`. The scope is now tight — three files, 66 LOC — and the earlier P0 findings from the previous review pass have been addressed cleanly. Approving with one merge-ordering condition and a handful of small polish suggestions.

---

## Verification of previous blockers

| Previous P0/P1 | Resolution at `481db46b` |
|---|---|
| `workflow-sanity.yml` upstream missing `workflow_call` trigger | N/A — `workflow-sanity.yml` is no longer part of this PR. |
| `codeql.yml` referencing an upstream reusable not yet on `@main` | N/A — `codeql.yml` is no longer part of this PR. |
| `release-publish.yml` missing `actions: read` | ✅ Fixed at `.github/workflows/release-publish.yml:31`. |
| `validate_run_id` optional, allowing CI-less releases | ✅ Now `required: true` at `.github/workflows/release-publish.yml:13`. PR description still calls it "optional"; please update (see P2-1). |
| PR description scope mismatch (five files added but only three documented) | ✅ Resolved by scope reduction — current diff matches PR body. |

---

## Merge-ordering condition (must be satisfied at merge time)

The two new callers reference reusables that **do not yet exist** on `Mininglamp-OSS/.github@main`:

- `.github/workflows/release-drafter.yml:12` → `Mininglamp-OSS/.github/.github/workflows/reusable-release-drafter.yml@main`
- `.github/workflows/release-publish.yml:24` → `Mininglamp-OSS/.github/.github/workflows/reusable-release-publish.yml@main`

`gh api repos/Mininglamp-OSS/.github/contents/.github/workflows?ref=main` currently lists `reusable-codeql.yml`, `reusable-pr-labeler.yml`, `reusable-stale.yml`, `workflow-sanity.yml`, and the `octo-*` family — but neither `reusable-release-drafter.yml` nor `reusable-release-publish.yml`. Those land via `Mininglamp-OSS/.github#27`, which is still `OPEN` at the time of this review.

If this PR merges before `.github#27`:

- The next push to `main` will fail the `Release Drafter` job with `error parsing called workflow: workflow file not found`.
- Any later `workflow_dispatch` against `Release Publish` will fail the same way.

The PR body already documents this dependency ("Requires Mininglamp-OSS/.github PR #27 merged first"), so this is a merge-gate concern rather than a code defect. Calling it out here so the merge bot / human merger does not lose track of it. Not blocking the approval.

---

## Findings on the current diff

### P2-1 — PR description out of sync with `validate_run_id`'s new semantics

The PR body still says:

> Inputs: tag (required), validate_run_id (optional CI gate), draft flag.

But `release-publish.yml:13` now declares `validate_run_id` as `required: true`, and the follow-up issue comment (`#issuecomment-4487299597`) explains why. Please refresh the PR description so operators reading the merged history understand the gate is mandatory. Minor doc hygiene; not blocking.

### P2-2 — `release-publish.yml` has no `concurrency` group

`.github/workflows/release-publish.yml:21-22` declares the `publish` job with no `concurrency:` block. Two operators dispatching the same `tag` within seconds can both pass the validation steps and race the `PATCH /releases/{id}` / `POST /releases` calls in the reusable. The upstream reusable does an "update existing release if present, else create" lookup (`reusable-release-publish.yml` lines 88–110), which narrows but does not eliminate the race — two concurrent runs can both observe "no existing release" and both attempt to create.

Recommend:

```yaml
concurrency:
  group: release-publish-${{ inputs.tag }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is intentional here — once a publish has crossed the validation gate, you do not want a second dispatch to abort it mid-way through the GitHub API call sequence; queuing is the safer behavior. Jerry-Xin flagged the same point in the earlier review; carrying it forward.

### P2-3 — `release-drafter.yml` config: name/tag template asymmetry

`.github/release-drafter.yml:2-3`

```yaml
name-template: "$NEXT_PATCH_VERSION"
tag-template:  "v$NEXT_PATCH_VERSION"
```

The rendered release will display as e.g. `1.4.0` while the tag is `v1.4.0`. Functional, but readers comparing the GitHub Releases UI to `git tag` output will see a discrepancy. Recommend `name-template: "v$NEXT_PATCH_VERSION"` so the release title matches the tag (this is also the upstream `release-drafter` README's recommended pairing).

### P2-4 — `release-drafter.yml` always increments patch; no `version-resolver` configured

`name-template`/`tag-template` are both hardcoded to `$NEXT_PATCH_VERSION`. Even when a PR labelled `enhancement` or `feature` merges, the suggested next tag will be a patch bump. Release-drafter supports a `version-resolver:` stanza that maps labels to major/minor/patch — without it, anyone wanting a minor or major release has to manually edit the draft's tag every time. Worth adding while you are touching this file:

```yaml
version-resolver:
  major:
    labels: [major, breaking]
  minor:
    labels: [enhancement, feature, minor]
  patch:
    labels: [bug, fix, patch, ci, infra, dependencies-changed, documentation, docs]
  default: patch
```

Non-blocking, but the cost of adding it is ~7 lines and it prevents a foot-gun the first time you want to ship a minor release.

### P2-5 — `dependencies-changed` label may not be defined or applied

`.github/release-drafter.yml:15` references the `dependencies-changed` label under "CI and Infrastructure". `gh label list --repo Mininglamp-OSS/octo-adapters` (and the default Dependabot/Renovate label set) does not include this exact name — Dependabot uses `dependencies`, Renovate uses `dependencies` by default as well. Release-drafter silently ignores categories whose labels never match, so this is not a correctness bug, but if the intent is to surface dependency-bot PRs in the changelog, the label name needs to either be created and wired into the bot config, or replaced with whatever the bot actually emits. Worth syncing across the org so the release notes are consistent.

### Nit — `exclude-labels` and `change-template` look good

`exclude-labels: [skip-changelog]` and `change-template: "- $TITLE @$AUTHOR (#$NUMBER)"` are both clean and match the upstream reusable's expectations. No action.

---

## What looks good

- **Least-privilege permissions**: top-level `permissions: {}` on both new workflows, with narrow per-job grants (`contents: write` + `pull-requests: read` for drafter; `contents: write` + `actions: read` for publisher). Exactly the right baseline.
- **Required CI gate**: `validate_run_id` is now mandatory, and the upstream reusable cross-checks both `conclusion == success` *and* `head_sha == tag commit sha` (`reusable-release-publish.yml:81-101`). This closes the "publish a tag whose commit never passed CI" hole.
- **Strict semver validation upstream**: the reusable's tag regex rejects `v1`, `v1.2`, leading zeros, and malformed pre-release segments before any write occurs. Good defense-in-depth.
- **Manual-dispatch-only publish path**: no accidental publish on tag push, release-drafter close, or scheduled cron. The intentional human action point is preserved.
- **SHA-pinned third-party action**: the upstream reusable pins `release-drafter/release-drafter@67e173cadb2fbd3de94f4a861e0c48c913b462ae` (v6). PR #41 inherits this transitively. Good supply-chain hygiene.
- **Update-or-create logic in the publisher**: the upstream reusable looks up an existing release by tag and `PATCH`es it if found, else `POST`s a new one — so dispatching publish against a tag that already has a draft release simply promotes the draft, which is the ergonomically correct behavior for the drafter → publisher handoff.

---

## Suggested next steps

1. Hold merge until `Mininglamp-OSS/.github#27` lands, so the two reusable refs resolve on `@main`.
2. Update the PR body to reflect `validate_run_id` being required, not optional (P2-1).
3. Add the `concurrency` block to `release-publish.yml` (P2-2) — small change, removes a real race surface.
4. Optional but recommended in the same PR or a quick follow-up: align `name-template` with `tag-template` (P2-3), add a `version-resolver` block (P2-4), and reconcile the `dependencies-changed` label with whatever the bot actually applies (P2-5).

None of the remaining items rise to a merge blocker on this PR's code itself. Once the upstream dependency is satisfied at merge time, this is ready to ship.

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-adapters` and the added release automation is narrowly scoped, follows the repository’s existing reusable-workflow pattern, and has no blocking issues.

💬 Non-blocking

🟡 Warning — `.github/workflows/release-drafter.yml:12` and `.github/workflows/release-publish.yml:24`: these workflows currently depend on `Mininglamp-OSS/.github` PR #27. Until that PR is merged to `main`, the referenced reusable workflows do not exist on the target ref, so this PR must not be merged first.

🟡 Warning — `.github/workflows/release-publish.yml:10-13`: the PR description says `validate_run_id` is optional, but the workflow makes it required. The implementation matches the “gated release publish” intent and the dependent reusable workflow signature, but the PR description should be corrected to avoid operator confusion.

🔵 Suggestion — `.github/workflows/release-drafter.yml:12` and `.github/workflows/release-publish.yml:24`: referencing reusable workflows at `@main` is consistent with some existing repo automation, but it means release behavior can change without a change in this repository. Consider pinning to a version tag once the shared workflows stabilize.

✅ Highlights

The workflows use top-level `permissions: {}` with job-scoped permissions, matching the repo’s least-privilege style.

The release publisher grants only `contents: write` and `actions: read`, which aligns with creating/updating releases and validating CI evidence.

The Release Drafter config categories match the repository’s existing issue/PR label vocabulary (`feature`, `enhancement`, `bug`, `documentation`).

lml2468 (commented):
## Re-review (round 3) — PR #41 (release drafter + gated publish)

**Verdict: APPROVED** (COMMENT due to self-PR)

Scope tightened: `codeql.yml` and `workflow-sanity.yml` removed from this PR (correctly split to PR #40). Remaining files are release-only:
- `.github/release-drafter.yml` — config (unchanged)
- `.github/workflows/release-drafter.yml` — caller (unchanged)
- `.github/workflows/release-publish.yml` — caller with `actions: read` + required `validate_run_id` (unchanged from last approved round)

All three files are identical to the last approved SHA. CI green. Only remaining dependency: PR #27 merge ordering (already documented).

PR #28, opened 2026-05-17
chore: fix CODEOWNERS to reference existing maintainers team

lml2468:
## What

Fix CODEOWNERS to reference the existing `@Mininglamp-OSS/maintainers` team.

## Why

The current CODEOWNERS references `@Mininglamp-OSS/core-team`, which does not exist as a GitHub team. This means the CODEOWNERS file is effectively a no-op — GitHub cannot resolve the team and code owner review requirements are silently skipped.

## How

Replace all `@Mininglamp-OSS/core-team` references with `@Mininglamp-OSS/maintainers` (the actual team that exists in the org with members: Jerry-Xin, yujiawei, an9xyz, lml2468).

## Checklist

- [x] Code compiles with no new warnings
- [x] Linter and formatter pass
- [x] Self-reviewed the full diff from a reviewer's perspective
- [x] Rebased onto latest `main`
- [x] Commit messages follow Conventional Commits

## Testing

N/A — configuration-only change. Verified `@Mininglamp-OSS/maintainers` team exists via GitHub API.

## Breaking Changes

N/A

PR #42, opened 2026-05-19
feat(ci): simplify octo-issue-feed to new/reopen only

lml2468:
## Summary

Simplifies the `octo-issue-feed.yml` caller following the product decision to:
1. Only forward **new issue (opened)** and **reopened** events — close/label/etc. are ignored
2. Send to the shared **issue-feed group only** (per-repo project group removed)
3. Remove IssueTriage bot mention (deprecated)

### Changes
- Trigger: `[opened, closed, reopened, labeled]` → `[opened, reopened]`
- Removed: `issue_labels`, `project_group_id` from `with:` block
- Switched: `@v1` → `@main` to pick up new reusable after `.github` PR #28 merges
- Added: `permissions: {}` at workflow level

## Merge order
**Merge `.github` PR #28 first** — the reusable `octo-issue-feed.yml` on main must match this caller interface before these callers work correctly.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -2,21 +2,19 @@ name: Octo Issue Feed
 
 on:
   issues:
-    types: [opened, closed, reopened, labeled]
+    types: [opened, reopened]
 
 permissions: {}
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
     with:
       repo_name: ${{ github.event.repository.name }}
       issue_number: ${{ github.event.issue.number }}
       issue_title: ${{ github.event.issue.title }}
       issue_url: ${{ github.event.issue.html_url }}
       issue_author: ${{ github.event.issue.user.login }}
-      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
       event_action: ${{ github.event.action }}
-      project_group_id: 0262cefe4463468bb6f2accaeda623cf
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}


yujiawei (approved):
# Code Review — PR #42 (octo-adapters)

## Verdict

**APPROVED** — The change is small, focused, and matches the stated intent. Findings below are improvements / follow-ups, not merge blockers. The author has explicitly acknowledged the cross-repo merge ordering in the PR body, so the responsibility for sequencing is documented.

## What the PR does

`.github/workflows/octo-issue-feed.yml`:

- Narrows the `issues` trigger from `[opened, closed, reopened, labeled]` to `[opened, reopened]` (L5).
- Drops `issue_labels` and `project_group_id` from the `with:` block sent to the reusable workflow (L17, L19 removed).
- Switches the reusable ref from `@v1` to `@main` (L11).

## Findings

### P1 — Cross-repo merge ordering is a real failure mode (acknowledged in PR body)

The reusable on `Mininglamp-OSS/.github` `main` still declares `project_group_id` as `required: true` (see `.github/.github/workflows/octo-issue-feed.yml` input list). If this caller (PR #42) merges to `main` **before** `Mininglamp-OSS/.github` PR #28 lands, every `opened`/`reopened` issue event will fail at workflow startup with an "missing required input: project_group_id" error.

`issue_labels` is safe to drop today — it has `default: '[]'` upstream — but `project_group_id` is not.

The PR body already says "Merge `.github` PR #28 first." That's the right plan; please make sure the merger actually enforces it. A small mitigation would be to land PR #28 first and only then enable auto-merge on this PR, rather than relying on human discipline.

### P2 — `@main` is a moving target (supply chain + reproducibility)

Pinning a reusable workflow to a mutable branch ref has two costs:

1. **Supply chain**: any future commit on `Mininglamp-OSS/.github` `main` (including a compromised one) silently affects every issue event in this repo. The reusable workflow already runs with `OCTO_BOT_TOKEN`, so the blast radius is real.
2. **Reproducibility**: re-running an old workflow run no longer executes the same code.

For comparison, sibling workflows in this same repo:

- `.github/workflows/octo-pr-feed.yml` → pinned to a SHA: `…@db17008c4190b88ebd6589948efe62a02ec4d085` (good)
- `.github/workflows/octo-ci-status.yml` → `@v1` (acceptable, version tag)
- This PR → `@main` (weakest)

**Suggested follow-up (not a blocker for this PR, but please file a tracking item):** once `.github` PR #28 merges, either move this caller to a pinned SHA (matching `octo-pr-feed.yml` style) or cut a new release tag (e.g. `@v2`) on `Mininglamp-OSS/.github` and pin to that. `@main` should be treated as a short-lived transition state, not the steady state.

### P2 — PR description accuracy

The PR body lists under "Changes":
> Added: `permissions: {}` at workflow level

The diff does not actually add this line — `permissions: {}` is already present in the unchanged context (it was added in an earlier change). Minor description inaccuracy, no code impact. Worth tidying the PR body so reviewers / future archaeologists aren't misled.

### Nit — Org-wide caller consistency

If the goal is to apply this simplification org-wide, sibling repos that also call `octo-issue-feed.yml@v1` (or similar) will keep sending the old richer event set + `project_group_id` until they are migrated. Consider opening a tracking issue to migrate the rest in lockstep with `.github` PR #28, so the behavior is uniform across the org and the `@v1` tag can eventually be retired.

## Verified

- ✅ Trigger types change matches the stated product decision.
- ✅ Removed inputs (`issue_labels`, `project_group_id`) match the new reusable interface in `.github` PR #28.
- ✅ `permissions: {}` at workflow level is present (already there pre-PR, fine).
- ✅ Secrets block still forwards `OCTO_BOT_TOKEN` correctly.
- ⚠️ `@main` works today but introduces ongoing maintenance risk — see P2 above.

## Summary

Approve. Land **after** `Mininglamp-OSS/.github` PR #28 is merged. Open a follow-up to swap `@main` for a pinned SHA or a versioned tag once the upstream interface stabilizes.

Jerry-Xin (dismissed):
Summary: The PR is relevant to this repository, but the reusable workflow reference introduces a blocking secret-handling supply-chain risk.

🔴 Blocking

- 🔴 Critical — [.github/workflows/octo-issue-feed.yml:11](/private/tmp/pr-review-R0820M/repo/.github/workflows/octo-issue-feed.yml:11) switches the reusable workflow from `@v1` to mutable `@main` while still passing `OCTO_BOT_TOKEN` at lines 19-20. Any future change on `Mininglamp-OSS/.github` `main` would run with this repository secret on issue events. Pin this to an immutable commit SHA after `.github` PR #28 merges, or to a reviewed release tag if the org treats tags as protected and immutable.

💬 Non-blocking

- 🟡 Warning — [.github/workflows/octo-issue-feed.yml:11](/private/tmp/pr-review-R0820M/repo/.github/workflows/octo-issue-feed.yml:11) is merge-order coupled to `.github` PR #28. The PR description documents this, but the caller will fail if merged before the reusable workflow on `main` supports the reduced input set.

✅ Highlights

- 🔵 Suggestion — The event filter at [.github/workflows/octo-issue-feed.yml:5](/private/tmp/pr-review-R0820M/repo/.github/workflows/octo-issue-feed.yml:5) matches the stated product decision to send only `opened` and `reopened` issue events.
- 🔵 Suggestion — Removing `issue_labels` and `project_group_id` keeps the caller aligned with the simplified shared issue-feed routing model, assuming the new reusable workflow owns the shared group default.

lml2468 (commented):
[APPROVE]

Straightforward CI simplification matching the product decision to forward only `opened`/`reopened` events.

**Changes verified:**
- Trigger narrowed: `[opened, closed, reopened, labeled]` → `[opened, reopened]` — removes noise events (close/label)
- Removed `issue_labels` / `project_group_id` — consistent with reusable workflow interface change in `.github` PR #28
- `@v1` → `@main` — intentional, picks up the updated reusable interface
- `permissions: {}` added — explicit zero-permission lock, good security hygiene

🟡 **Merge order**: PR body documents this correctly — merge `.github` PR #28 first; callers will break if merged before the reusable interface is updated.

LGTM — approve when `.github` PR #28 is merged.

lml2468 (commented):
## Independent Review — PR #42 (simplify octo-issue-feed)

**Verdict: APPROVED with reservation** (COMMENT due to self-PR)

Tiny change: +2/-4, 1 file. Reduces issue-feed to `opened`/`reopened` only, drops `issue_labels` + `project_group_id`, switches `@v1` → `@main`.

### On @Jerry-Xin's blocker (`@main` + `OCTO_BOT_TOKEN`) — I **disagree** as a blocker

Allen flags the `@v1` → `@main` switch as a supply-chain risk because `OCTO_BOT_TOKEN` is passed. This is a valid security observation in principle, but in the context of this repo it is not blocking:

1. **`@main` is already the established pattern** for newer reusable workflow callers in this repo. `release-drafter.yml`, `release-publish.yml`, `stale.yml`, and `labeler.yml` all use `@main` — some with `contents: write` permissions.

2. **The `.github` repo is org-internal** (`Mininglamp-OSS/.github`). Push-to-main is controlled by org owners. The blast radius of a compromise is the same regardless of `@v1` vs `@main` since the org controls tag creation too (tags are mutable by default in GitHub).

3. **`@v1` itself is mutable** — Git tags can be force-pushed. SHA pinning (like `octo-pr-feed.yml`) is the only truly immutable option. If the policy is "pin secrets-passing workflows to SHA," that should be a repo-wide policy change applied to all existing workflows, not a blocker on this one PR.

That said, this is a legitimate security policy decision for the org. I'd recommend documenting the org's pinning policy rather than blocking individual PRs inconsistently.

### 🔴 Actual blocker — merge ordering

PR body correctly documents: **merge `.github` PR #28 first**. The upstream reusable at `@main` currently has `project_group_id: required: true`. This caller removes that input. If this PR merges before #28, every `opened`/`reopened` issue event will fail with "Required input 'project_group_id' not provided."

This is already documented in the PR body and is a merge-ordering constraint, not a code bug.

### 🟡 Non-blocking

The `permissions: {}` top-level was **already present** on `main` (not a new addition despite the PR description mentioning it). The PR description may be referring to the intent rather than a diff-visible change.

### ✅ Highlights

- Event filter reduction (`opened, reopened` only) is clean and matches the product decision
- Removing unused inputs keeps the caller aligned with the simplified reusable interface

Jerry-Xin (approved):
Replacing my earlier RC. The `@main` supply-chain concern is a valid security observation but not a per-PR blocker — `@v1` is equally mutable (force-pushable tag), and this repo already uses `@main` for other secret-passing reusable workflows (`release-drafter`, `release-publish`, `stale`, `labeler`). SHA pinning should be pursued as a repo-wide policy change if desired.

PR itself is clean: event scope narrowed to `opened`/`reopened`, dead inputs removed. Only merge-order constraint: `.github` PR #28 must land first (adds `project_group_id: required: true`).

PR #55, opened 2026-05-21
fix(accounts): tolerate case-insensitive accountId lookup

### LINKED ISSUE #54: [bug] Outbound replies silently fail when bot accountId contains uppercase characters
### Description

When OpenClaw routes an outbound reply through `outbound.sendText` (e.g. the followup queue path used for agent-generated final replies in DM sessions), the plugin throws `Error: Octo botToken is not configured` and the reply is silently dropped. The error only surfaces in OpenClaw's debug-level file log; the user's IM client receives nothing.

### Steps to Reproduce

1. Configure a bot account with a mixed-case ID (e.g. `27pBwzf2F6bfa5cd142_bot`, as generated by our botfather) in `openclaw.json` under `channels.octo.accounts`
2. Bind an isolated agent (e.g. `cc`) to that account via `bindings`
3. From the IM client, send a DM to the bot that triggers async agent work (e.g. attach an image and ask the agent to analyze it)
4. Observe OpenClaw file log: `followup queue: route-reply failed: Failed to route reply to octo: Octo botToken is not configured`
5. IM client never receives the agent's final reply (only short synchronous replies that go through the plugin `deliver` callback path are delivered)

### Expected Behavior

Agent's final reply is delivered to the IM client regardless of bot accountId casing.

### Actual Behavior

Reply is silently dropped on the followup queue / route-reply path. Only OpenClaw debug-level log shows the failure; no user-visible error.

### Adapter Version

1.0.0

### Affected Component

OpenClaw Channel Plugin

### Severity

P1 - Core Feature Broken

### Reproducibility

- [x] Consistently reproducible

### Environment

- OS: macOS Darwin 25.3.0
- Node.js: 22.x
- OpenClaw: 2026.5.18
- Octo Server: production (im.deepminer.com.cn)

### Logs / Root Cause Analysis

OpenClaw's routing layer normalizes `accountId` to lowercase via `normalizeAccountId` (in OpenClaw's `account-id-*.js` -> `canonicalizeAccountId`):

```js
function canonicalizeAccountId(value) {
  const normalized = normalizeLowercaseStringOrEmpty(value); // forces lowercase
  ...
}
```

But `resolveDmworkAccount` in `openclaw-channel-octo/src/accounts.ts` does a strict, case-sensitive lookup:

```ts
const accountConfig = channel.accounts?.[accountId] ?? channel;
```

When the routing layer passes the lowercase `27pbwzf2f6bfa5cd142_bot` to the plugin, the strict lookup misses the mixed-case key, falls through to the top-level channel config (which has no `botToken`), and `outbound.sendText` throws `Octo botToken is not configured`.

Concrete log evidence (gateway debug log, redacted):

```
[routing] resolveAgentRoute: channel=octo accountId=27pbwzf2f6bfa5cd142_bot peer=direct:... bindings=1
[routing] binding: agentId=cc accountPattern=27pBwzf2F6bfa5cd142_bot ...
[routing] match: matchedBy=binding.account agentId=cc
...
followup queue: route-reply failed: Failed to route reply to octo: Octo botToken is not configured
```

The plugin `deliver` callback path is unaffected because it uses the `account` config cached at inbound time (where the original mixed-case key is preserved). Only outbound paths that re-resolve account via `resolveDmworkAccount(accountId)` with the lowercased ID hit this bug.

### Proposed Fix

Add a case-insensitive fallback in `resolveDmworkAccount`. Strict lookup remains the primary path (O(1)); the fallback only runs when the strict lookup misses (account count is small, typically <10). Long-term, the botfather should generate lowercase bot IDs to match OpenClaw's contract; this fix bridges the gap and remains useful for historical mixed-case IDs already in production.

PR will follow shortly.

caster-Q:
## Summary

Fix silent outbound reply drops when bot accountId contains uppercase characters.

OpenClaw's routing layer normalizes `accountId` to lowercase via `normalizeAccountId`, but botfather generates mixed-case bot IDs (e.g. `27pBwzf2F6bfa5cd142_bot`). The strict lookup in `resolveDmworkAccount` fails to match the mixed-case config key when given the lowercase routing-normalized form, falls through to the top-level channel config (which has no `botToken`), and throws `Octo botToken is not configured`. The error only surfaces in OpenClaw's debug-level file log; the user's IM client receives nothing.

This affects every agent-driven reply that goes through the followup queue / route-reply path (e.g. image-processing replies, long async replies). Short synchronous replies that go through the plugin `deliver` callback path are unaffected (that path uses the cached inbound `account` config, which preserves the original mixed-case key).

## Related Issue

Fixes #54

## Changes

- `openclaw-channel-octo/src/accounts.ts`: add case-insensitive fallback in `resolveDmworkAccount`. The strict lookup remains the primary O(1) path; the fallback only runs when the strict lookup misses. Account count is small (typically <10), so the `Object.entries(...).find(...)` cost is negligible.
- `openclaw-channel-octo/src/__tests__/accounts.test.ts`: add 5 unit tests covering: strict match (primary path), case-insensitive fallback (bridge), case-insensitive fallback with multiple coexisting mixed-case accounts (correctness), missing-key top-level fallback (backward-compat), and missing botToken (returns `configured: false`).

## Testing

- [x] `npm run type-check` passes
- [x] `npm test` passes (755/755 tests, including 5 new ones in `accounts.test.ts`)
- [x] Manually verified on local gateway: with the fix applied to `~/.openclaw/npm/node_modules/openclaw-channel-octo/dist/src/accounts.js`, mixed-case bot IDs (`27pBwzf2F6bfa5cd142_bot`) successfully deliver agent replies through the route-reply path. Without the fix, `outbound.sendText` throws `Octo botToken is not configured` and the reply is silently dropped (debug-log evidence in #54).

## Long-Term Direction (out of scope for this PR)

The deeper architectural alignment is for botfather to generate lowercase bot IDs to match OpenClaw's contract (`canonicalizeAccountId` in `openclaw/dist/account-id-*.js`). This PR is a plugin-side bridge — it remains useful for historical mixed-case bot IDs already in production even after botfather is migrated.

## Checklist

- [x] Added tests for my changes (5 new unit tests in `accounts.test.ts`)
- [x] Updated documentation (no doc changes needed for an internal bug fix)
- [x] Followed commit message conventions (Conventional Commits: `fix(accounts): ...`)

### DIFF
--- a/openclaw-channel-octo/src/__tests__/accounts.test.ts
+++ b/openclaw-channel-octo/src/__tests__/accounts.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect } from "vitest";
+import { resolveDmworkAccount } from "../accounts.js";
+
+/**
+ * Unit tests for resolveDmworkAccount account-id case-insensitive fallback.
+ *
+ * Background: OpenClaw's routing layer normalizes accountId to lowercase via
+ * normalizeAccountId (canonicalizeAccountId in openclaw/dist/account-id-*.js),
+ * but botfather generates mixed-case bot IDs (e.g. "27pBwzf2F6bfa5cd142_bot").
+ * Without a case-insensitive fallback, outbound paths that re-resolve account
+ * from the lowercased ID would miss the mixed-case config key and throw
+ * "botToken is not configured", silently dropping replies.
+ */
+
+const MIXED_CASE_ID = "27pBwzf2F6bfa5cd142_bot";
+const LOWERCASE_ID = "27pbwzf2f6bfa5cd142_bot";
+const OTHER_MIXED_CASE_ID = "27pBwJ4bfWKed86272a_bot";
+
+function buildCfg(accounts: Record<string, unknown>): unknown {
+  return {
+    channels: {
+      octo: {
+        accounts,
+      },
+    },
+  };
+}
+
+describe("resolveDmworkAccount", () => {
+  describe("strict (case-sensitive) match — primary path", () => {
+    it("returns account config when accountId matches the configured key exactly", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_strict_match", apiUrl: "https://im.example.com/api" },
+      });
+
+      const account = resolveDmworkAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID });
+
+      expect(account.config.botToken).toBe("bf_strict_match");
+      expect(account.config.apiUrl).toBe("https://im.example.com/api");
+      expect(account.configured).toBe(true);
+    });
+  });
+
+  describe("case-insensitive fallback — bridges OpenClaw lowercase routing", () => {
+    it("resolves a mixed-case configured account when given its lowercase form", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_case_insensitive", apiUrl: "https://im.example.com/api" },
+      });
+
+      const account = resolveDmworkAccount({ cfg: cfg as never, accountId: LOWERCASE_ID });
+
+      expect(account.config.botToken).toBe("bf_case_insensitive");
+      expect(account.config.apiUrl).toBe("https://im.example.com/api");
+      expect(account.configured).toBe(true);
+    });
+
+    it("matches the correct account when multiple mixed-case accounts coexist", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_first", apiUrl: "https://first.example.com/api" },
+        [OTHER_MIXED_CASE_ID]: { botToken: "bf_second", apiUrl: "https://second.example.com/api" },
+      });
+
+      const first = resolveDmworkAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID.toLowerCase() });
+      const second = resolveDmworkAccount({ cfg: cfg as never, accountId: OTHER_MIXED_CASE_ID.toLowerCase() });
+
+      expect(first.config.botToken).toBe("bf_first");
+      expect(second.config.botToken).toBe("bf_second");
+    });
+  });
+
+  describe("missing account — fallback to top-level channel config", () => {
+    it("falls back to top-level channel config when accountId is not found in any case", () => {
+      const cfg = {
+        channels: {
+          octo: {
+            botToken: "bf_top_level",
+            apiUrl: "https://top-level.example.com/api",
+          },
+        },
+      };
+
+      const account = resolveDmworkAccount({ cfg: cfg as never, accountId: "nonexistent_account" });
+
+      expect(account.config.botToken).toBe("bf_top_level");
+      expect(account.config.apiUrl).toBe("https://top-level.example.com/api");
+    });
+
+    it("returns undefined botToken when neither account nor top-level config has it", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { apiUrl: "https://im.example.com/api" },
+      });
+
+      const account = resolveDmworkAccount({ cfg: cfg as never, accountId: "totally_different" });
+
+      expect(account.config.botToken).toBeUndefined();
+      expect(account.configured).toBe(false);
+    });
+  });
+});
--- a/openclaw-channel-octo/src/accounts.ts
+++ b/openclaw-channel-octo/src/accounts.ts
@@ -56,7 +56,21 @@ export function resolveDmworkAccount(params: {
 }): ResolvedDmworkAccount {
   const accountId = params.accountId ?? DEFAULT_ACCOUNT_ID;
   const channel = getChannelConfig<DmworkAccountConfig>(params.cfg);
-  const accountConfig = channel.accounts?.[accountId] ?? channel;
+  // Strict lookup first; fall back to case-insensitive match because OpenClaw's
+  // routing layer normalizes accountId to lowercase via normalizeAccountId
+  // (canonicalizeAccountId in openclaw/dist/account-id-*.js), while botfather
+  // generates mixed-case bot IDs (e.g. "27pBwzf2F6bfa5cd142_bot"). Without the
+  // fallback, outbound paths that re-resolve account from the lowercased ID
+  // miss the mixed-case config key and throw "botToken is not configured",
+  // silently dropping replies. Long-term, botfather should produce lowercase
+  // IDs to match OpenClaw's contract; this bridges the gap and keeps working
+  // for historical mixed-case bot IDs already in production.
+  const accountConfig =
+    channel.accounts?.[accountId]
+    ?? Object.entries(channel.accounts ?? {}).find(
+      ([key]) => key.toLowerCase() === accountId.toLowerCase(),
+    )?.[1]
+    ?? channel;
 
   const botToken = accountConfig.botToken ?? channel.botToken;
   const apiUrl = accountConfig.apiUrl ?? channel.apiUrl ?? DEFAULT_API_URL;


lml2468 (approved):
[APPROVE]

Solid fix for a real production silent-failure. Read the full `accounts.ts` source on the PR branch plus all 99 lines of tests.

**Root cause & fix**

OpenClaw normalizes accountId to lowercase at the routing layer, but botfather emits mixed-case IDs (e.g. `27pBwzf2F6bfa5cd142_bot`). The strict `channel.accounts?.[accountId]` lookup silently misses, falls back to the bare channel root, finds no `botToken`, and drops the reply. The three-tier fallback (strict → case-insensitive → channel root) is the right shape:

```ts
channel.accounts?.[accountId]
  ?? Object.entries(channel.accounts ?? {}).find(
      ([key]) => key.toLowerCase() === accountId.toLowerCase(),
    )?.[1]
  ?? channel;
```

- Strict match is tried first — zero regression for correctly-cased IDs. ✅
- Fallback only fires on a miss, so existing deployments with normalized IDs are unaffected. ✅
- Comment names the root cause, the symptom, and the long-term fix direction (botfather should emit lowercase). ✅

**Tests**

99 lines for a 14-line change. Five cases cover all branches: strict match, case-insensitive fallback (single and multi-account), top-level channel fallback, and the "nothing matches, returns undefined botToken" path. All meaningful branches are hit.

**Minor suggestions (non-blocking)**

🔵 `accountId.toLowerCase()` is re-evaluated on every `find()` iteration. Hoisting it outside saves repeated calls, though with O(1–2) accounts in practice it does not matter:
```ts
const accountIdLower = accountId.toLowerCase();
// ...
([key]) => key.toLowerCase() === accountIdLower,
```

🔵 No test for the degenerate case where two config keys normalize to the same lowercase (e.g. `"FooBot"` and `"foobot"`). `Object.entries` ordering is insertion-order in V8 so it is deterministic in practice, but worth a note or a future guard if botfather ever generates near-duplicate IDs.

Neither is a blocker.

Jerry-Xin (approved):
Summary: The PR is relevant to `openclaw-channel-octo` and the account lookup change is narrowly scoped, backward-compatible, and covered by focused tests.

💬 Non-blocking:
- 🔵 Suggestion: In [accounts.ts](/private/tmp/pr-review-dV9ETK/repo/openclaw-channel-octo/src/accounts.ts:68), the fallback chooses the first case-insensitive match if two configured account keys differ only by case. That configuration is already ambiguous under OpenClaw’s lowercase accountId contract, so this is not blocking, but a future guard or diagnostic could make misconfiguration easier to detect.

✅ Highlights:
- Strict exact lookup remains the primary path at [accounts.ts](/private/tmp/pr-review-dV9ETK/repo/openclaw-channel-octo/src/accounts.ts:68), preserving existing behavior and performance.
- The fallback correctly preserves top-level channel fallback semantics at [accounts.ts](/private/tmp/pr-review-dV9ETK/repo/openclaw-channel-octo/src/accounts.ts:73).
- The new tests cover exact match, lowercase route bridge, multiple mixed-case accounts, legacy top-level fallback, and missing-token behavior in [accounts.test.ts](/private/tmp/pr-review-dV9ETK/repo/openclaw-channel-octo/src/__tests__/accounts.test.ts:29).

Verification note: I attempted `npm run type-check` and the targeted Vitest run, but this checkout has no installed package dependencies, so `tsc` and `vitest` were unavailable locally.

lml2468 (commented):
[APPROVE — cannot self-approve via bot account, leaving as COMMENT]

## Code Review — PR #55

**Verdict: APPROVED.** Minimal, correct fix for a real outbound-drop bug.

### Analysis

Single production code change: `accounts.ts:68-72` adds a case-insensitive fallback to `resolveDmworkAccount`. The lookup chain is now:

1. **Strict match** `channel.accounts?.[accountId]` — O(1), handles correctly-cased IDs
2. **Case-insensitive fallback** `Object.entries().find(key.toLowerCase() === ...)` — O(n), bridges the gap
3. **Top-level channel config** — existing fallback when no account matches

The root cause is clear: OpenClaw's `normalizeAccountId` lowercases the routing key, but botfather emits mixed-case bot IDs. The fix bridges this mismatch at the adapter level while the comment correctly identifies the long-term fix (botfather should produce lowercase IDs).

### Verification

- `Object.entries().find()` is O(n) but account count is tiny — no perf concern
- If two config keys differ only by case (e.g. `FooBot` / `fooBot`), the fallback returns whichever `Object.entries` yields first — this is an inherently ambiguous config that shouldn't exist. Allen noted the same; agreed it's non-blocking
- The fix preserves the existing fallback-to-top-level behavior: if neither strict nor case-insensitive matches, it falls through to `channel` (same as before)
- 99 lines of focused tests covering strict match, case-insensitive bridge, multi-account disambiguation, and missing-account fallback

### No Blockers

---

yujiawei (approved):
# Code Review — PR #55 (octo-adapters)

## Verification

| Item | Result | Evidence |
|---|---|---|
| Fix targets the right call site | ✅ | `openclaw-channel-octo/src/accounts.ts:68-73` — strict lookup keeps the O(1) hot path, fallback only runs on miss. |
| Strict path preserved (no behavior change for matching keys) | ✅ | `accounts.ts:69` — `channel.accounts?.[accountId]` is evaluated first. |
| Top-level channel fallback preserved | ✅ | `accounts.ts:73` — `?? channel` still terminates the chain when no per-account entry matches. |
| Outbound path will see the fix | ✅ | `channel.ts:486-488` (`sendText`) and `channel.ts:563-565` (`sendMedia`) both call `resolveDmworkAccount`, so the fallback bridges OpenClaw's lowercased `ctx.accountId` to the mixed-case config key. |
| Agent-tool path consistency | ✅ | `agent-tools.ts:222-247` already does case-insensitive resolution with explicit ambiguity guarding, so the new fallback in `accounts.ts` is consistent (though stricter — see Findings §1). |
| `type-check` passes | ✅ | `npm run type-check` clean on `9b1b3ee`. |
| New unit tests pass | ✅ | `npm test -- --run accounts.test.ts` → 5/5 pass. |
| Tests cover the regression | ✅ | `__tests__/accounts.test.ts:39-67` exercises the actual mixed-case-config / lowercase-lookup scenario from #54. |

## Findings

### P2 — Silent first-match wins under pathological key collision

`accounts.ts:70-72` uses `Object.entries(...).find(...)`, which returns the first match in insertion order. If a config ever contains two account keys that differ only in case (e.g. `"BotA"` and `"bota"`), a lookup with a third casing (`"BOTA"`) will silently pick whichever was inserted first.

`agent-tools.ts:230-241` already guards against this exact case with an explicit "ambiguous" error. The new fallback in `accounts.ts` does not. The two paths are now subtly inconsistent: a malformed multi-account config would be rejected loudly at agent-tool entry but resolved silently (and possibly to the wrong bot) on the outbound path.

Real-world risk is low (no operator is likely to register two near-duplicate accounts), and the prior behavior was "drop the reply entirely" — so even silent wrong-routing is a strict improvement. Calling this out as a follow-up rather than a blocker:

```ts
const lowerId = accountId.toLowerCase();
const matches = Object.entries(channel.accounts ?? {}).filter(
  ([key]) => key.toLowerCase() === lowerId,
);
if (matches.length > 1) {
  // log.warn or pick a deterministic strategy
}
const accountConfig = channel.accounts?.[accountId] ?? matches[0]?.[1] ?? channel;
```

Not required to land. Worth a one-line code comment at minimum so the next reader knows ambiguity is intentionally not handled here.

### P2 — Fallback is `toLowerCase`-only, not full canonicalization

OpenClaw's `canonicalizeAccountId` (`openclaw/dist/account-id-*.js`) does more than lowercase: it also replaces any character outside `[a-z0-9_-]` with `-`, strips leading/trailing `-`, and truncates to 64 chars. A config key like `"my.bot"` would canonicalize upstream to `"my-bot"`, and `key.toLowerCase() === accountId.toLowerCase()` would still miss.

For the current botfather-generated ID format (`[A-Za-z0-9]+_bot`), this is irrelevant — only ASCII letters differ between the two forms. So the fix correctly handles the actual production failure mode. Flagging only because the comment at `accounts.ts:59-67` calls out `normalizeAccountId` as the source of truth; readers may assume the fallback mirrors it fully. A short note like "case-only bridge; non-alphanumeric chars are not handled because current bot IDs never contain them" would prevent confusion.

### P2 — Returned `accountId` reflects input, not matched key

`resolveDmworkAccount` returns `accountId: params.accountId ?? DEFAULT_ACCOUNT_ID` (line 92, ignoring the fallback). So when the fallback path resolves the config, the returned object reports the lowercase ID, not the mixed-case config key that actually owns the bot token.

This is harmless for every caller in this PR: the outbound paths use `account.config.botToken` directly, and the cache-key callers (`channel.ts:805`, `833-834`, `905-908`, `973`) all iterate via `listDmworkAccountIds` which returns the original config keys (mixed-case), so caches stay keyed consistently on the mixed-case form. Worth keeping in mind if a future caller starts using `account.accountId` to key cross-path state — there's a latent inconsistency between "what config matched" and "what we tell you matched."

### Nit — Test naming consistency

`__tests__/accounts.test.ts:90` test name says "returns undefined botToken when neither account nor top-level config has it", but the assertion is `expect(account.config.botToken).toBeUndefined()` — accurate; just noting `configured: false` is the more meaningful contract assertion (already covered). No change needed.

## Suggestions

- Add a one-line code comment in `accounts.ts:70-72` clarifying that the fallback intentionally does not detect ambiguous case-only collisions (since `agent-tools.ts` handles that at a higher layer). This prevents the next maintainer from "fixing" something that isn't broken.
- Long-term direction (botfather lowercase migration) is already noted in the PR description — good. Consider also linking from the source comment to that follow-up so the bridge is unambiguously temporary.

## Additional Notes

- The 5 new tests exercise the right cases. Adding a 6th asserting first-match-wins on case-collision keys would document the current behavior, but it's optional.
- Build artifact note: I had to run `npm install` inside `openclaw-channel-octo/` to satisfy peer-deps before `type-check`; this is normal for the workspace layout and not a PR concern.
- No security-relevant surface change: no new external input, no auth bypass, no privileged action introduced.

## Verdict

**APPROVED.** The fix is minimal, well-isolated to the actual failure mode, preserves the O(1) hot path, and is covered by targeted tests. The findings above are quality-of-life follow-ups and do not block merge — they can ship as a tiny follow-up PR or be folded in here if the author prefers.

PR #35, opened 2026-05-19
feat(octo): persona-clone adapter passthrough (onBehalfOf → on_behalf_of)

yujiawei:
## Summary

RFC `persona-clone` §6 (PR-B) — Adapter passthrough for the new account-level config `onBehalfOf` (Persona Clone / OBO). When set, every outbound `/v1/bot/sendMessage` request includes `on_behalf_of: <uid>` so the bot delivers messages as the real user under a server-side OBO grant.

Backend acceptance lands in PR-A (Mininglamp-OSS/octo-server#81). This PR is **backward-compatible both ways** — when `onBehalfOf` is unset/empty the request body is byte-identical to `main`, and the backend ignores the unknown field until PR-A merges.

> Note: An earlier PR `dmwork-org/dmwork-adapters#241` was opened against the wrong (frozen) repo and has been closed. This PR is the correct refile against `Mininglamp-OSS/octo-adapters` `main`, targeting the **`openclaw-channel-octo/`** subdir (NOT `openclaw-channel-dmwork/`, which is kept only for dependabot tracking per #29).

## Scope (strict, ~10 LOC core)

| File | Change |
| --- | --- |
| `openclaw-channel-octo/src/config-schema.ts` | Add optional `onBehalfOf: string` to `DmworkAccountConfig` + `DmworkConfig` + JSON Schema (account + global) |
| `openclaw-channel-octo/openclaw.plugin.json` | Mirror the same property in the manifest schema (kept in sync by `manifest-schema-sync` test) |
| `openclaw-channel-octo/src/accounts.ts` | Surface on `ResolvedDmworkAccount.config` (account override → channel default) |
| `openclaw-channel-octo/src/api-fetch.ts` | `sendMessage` + `sendMediaMessage` forward `on_behalf_of`, **omitted when unset/empty** |
| `openclaw-channel-octo/src/inbound.ts` | Pass `account.config.onBehalfOf` at the 3 send call sites: main reply, error reply, media upload |
| `openclaw-channel-octo/README.md`, `CHANGELOG.md` | Brief docs entries |
| `openclaw-channel-octo/src/api-fetch.test.ts` | 3 new unit tests: set / unset / media passthrough |

## Diff stat

```
 CHANGELOG.md                                |  3 +
 openclaw-channel-octo/README.md             |  1 +
 openclaw-channel-octo/openclaw.plugin.json  |  4 +-
 openclaw-channel-octo/src/accounts.ts       |  2 +
[...]

PR #57, opened 2026-05-21
feat(octo): OBO v2 adapter — reply routing + on_behalf_of + system hint + @AI mention

yujiawei:
## Changes

OBO v2 adapter 侧适配，配合 octo-server PR#109（已 merged）。

### 1. sendMessage + sendTyping 支持 on_behalf_of
- `api-fetch.ts`: `sendMessage` 和 `sendTyping` 均新增 `onBehalfOf` 参数
- 调用时传入 `on_behalf_of` 到 API，绕过 bot friend gate

### 2. OBO v2 reply 路由
- 识别 fan-out payload 中的 `obo_origin_channel_id`、`obo_origin_channel_type`、`obo_respond_as`
- **群聊**：回复到 origin group，以 grantor 身份（`on_behalf_of`）
- **DM**：回复到 `obo_origin_from_uid`（原始发送者），以 grantor 身份

### 3. persona_prompt 注入 LLM
- `obo_system_hint` 注入到 `GroupSystemPrompt`，LLM 收到分身人格提示词

### 4. typing indicator
- OBO 场景下 typing 发到 origin channel，携带 `on_behalf_of=grantor`

### 5. @AI（mention.ais）支持
- `MentionPayload` 新增 `ais` 字段
- `isMentioned` 逻辑加上 `mentionAis` 检查，@AI 时 bot 正确响应

## Testing
- bob 私聊 admin → James 以 admin 身份回复到 bob ✅
- persona_prompt "用幽默搞笑的风格来回复" 生效 ✅
- @James 在群里正常回复 ✅
- @AI 在群里触发 James 回复 ✅（本 PR 修复）

## Dependencies
- octo-server#109 (merged) — OBO v2 backend
- octo-server#114 (pending) — fan-out group scope bypass

### DIFF
--- a/openclaw-channel-octo/src/api-fetch.ts
+++ b/openclaw-channel-octo/src/api-fetch.ts
@@ -215,6 +215,7 @@ export async function sendMessage(params: {
   mentionEntities?: MentionEntity[];
   mentionAll?: boolean;
   replyMsgId?: string;
+  onBehalfOf?: string;
   signal?: AbortSignal;
 }): Promise<SendMessageResult | undefined> {
   const payload: Record<string, unknown> = {
@@ -247,6 +248,7 @@ export async function sendMessage(params: {
     channel_id: params.channelId,
     channel_type: params.channelType,
     payload,
+    ...(params.onBehalfOf ? { on_behalf_of: params.onBehalfOf } : {}),
   }, params.signal);
 }
 
@@ -255,11 +257,13 @@ export async function sendTyping(params: {
   botToken: string;
   channelId: string;
   channelType: ChannelType;
+  onBehalfOf?: string;
   signal?: AbortSignal;
 }): Promise<void> {
   await postJson(params.apiUrl, params.botToken, "/v1/bot/typing", {
     channel_id: params.channelId,
     channel_type: params.channelType,
+    ...(params.onBehalfOf ? { on_behalf_of: params.onBehalfOf } : {}),
   }, params.signal);
 }
 
--- a/openclaw-channel-octo/src/inbound.ts
+++ b/openclaw-channel-octo/src/inbound.ts
@@ -1204,7 +1204,10 @@ export async function handleInboundMessage(params: {
     const mentionUids = extractMentionUids(message.payload?.mention);
     const mentionAllRaw = message.payload?.mention?.all;
     const mentionAll: boolean = mentionAllRaw === true || mentionAllRaw === 1;
-    isMentioned = (!account.config.ignoreMentionAll && mentionAll) || mentionUids.includes(botUid);
+    // mention.ais=1 means @AI / @所有AI — bots should respond
+    const mentionAisRaw = message.payload?.mention?.ais;
+    const mentionAis: boolean = mentionAisRaw === true || mentionAisRaw === 1;
+    isMentioned = (!account.config.ignoreMentionAll && mentionAll) || mentionAis || mentionUids.includes(botUid);
     isExplicitBotMention = mentionUids.includes(botUid);
 
     // Defensive fallback: if payload.mention is missing/empty but the message
@@ -1565,7 +1568,11 @@ export async function handleInboundMessage(params: {
     MessageSid: String(message.message_id),
     Timestamp: message.timestamp ? message.timestamp * 1000 : undefined,
     GroupSubject: isGroup ? message.channel_id : undefined,
-    GroupSystemPrompt: undefined,
+    // OBO v2: inject obo_system_hint as GroupSystemPrompt so it reaches LLM
+    // as a system-level instruction (persona identity + reply context)
+    GroupSystemPrompt: (typeof message.payload?.obo_system_hint === "string" && message.payload.obo_system_hint.length > 0)
+      ? message.payload.obo_system_hint
+      : undefined,
     Provider: CHANNEL_ID,
     Surface: CHANNEL_ID,
     OriginatingChannel: CHANNEL_ID,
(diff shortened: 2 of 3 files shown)


lml2468 (changes requested):
[REQUEST_CHANGES]

Overall the feature is logically sound — the three-part structure (reply routing, `on_behalf_of` propagation, `@AI` mention) follows existing patterns cleanly, and the non-OBO fallback path is untouched. Two security questions need answers before this merges.

---

**🟡 Issue 1 — `obo_system_hint` is injected as `GroupSystemPrompt` unconditionally (before the `isOBOv2` check)**

```ts
// line ~1573 — runs for EVERY inbound message
GroupSystemPrompt: (typeof message.payload?.obo_system_hint === "string" && ...
  ? message.payload.obo_system_hint
  : undefined,

// line ~1600 — isOBOv2 is determined only AFTER the context is built
const isOBOv2 = Boolean(...);
```

If a sender can include `obo_system_hint` in a regular (non-OBO) message payload, they can inject arbitrary system-level instructions into the LLM — a classic prompt-injection vector. The question is whether Octo's server strips or blocks this field for non-OBO senders. If yes, please add a comment stating the trust assumption. If not, the injection should be gated:

```ts
GroupSystemPrompt: isOBOv2 && typeof message.payload?.obo_system_hint === "string" ...
```

(Or moved below the `isOBOv2` block, since `inboundContext` is consumed later.)

---

**🟡 Issue 2 — OBO mode is activated purely by payload fields, with no sender identity check**

```ts
const isOBOv2 = Boolean(
  typeof oboV2OriginChannel === "string" && oboV2OriginChannel.length > 0 &&
  typeof oboV2RespondAs === "string"     && oboV2RespondAs.length > 0
);
```

Any sender who can craft a DM with `obo_origin_channel_id` + `obo_respond_as` can:
1. Redirect the bot's reply to an arbitrary channel
2. Set `on_behalf_of` to any uid, effectively impersonating any user

If the Octo server guarantees that only the OBO routing service can inject these payload fields (e.g. by stripping them from regular user messages), this is fine — but that assumption should be documented in a comment. If that guarantee doesn't exist, a `from_uid` allowlist check should be added before trusting these fields.

---

**🔵 Suggestion — `ignoreMentionAll` does not suppress `mentionAis`**

```ts
isMentioned = (!account.config.ignoreMentionAll && mentionAll) || mentionAis || mentionUids.includes(botUid);
```

`mentionAis` bypasses `ignoreMentionAll`. This may be intentional (bots that opt out of `@all` still want `@AI`), but it's worth a short comment clarifying the design choice so the next reader doesn't assume it's an oversight.

---

**Rest of the diff looks good**
- `onBehalfOf` propagated consistently: typing interval ✅, error message ✅, `sendMessage` ✅
- Type guards on all payload fields ✅
- `obo_respond_as ?? obo_grantor_uid` fallback for backward compat ✅
- DM vs Group branch in OBO routing is clear and correct ✅

yujiawei (commented):
# Code Review — PR #57 (octo-adapters)

## 1. Verification

| Stated guarantee | Status | Evidence |
|---|---|---|
| `sendMessage` / `sendTyping` accept and forward `onBehalfOf` → `on_behalf_of` | ✅ | `openclaw-channel-octo/src/api-fetch.ts:218,251,260-267` — top-level field, spread only when truthy |
| Group OBO v2 → reply to `obo_origin_channel_id` as grantor | ✅ | `openclaw-channel-octo/src/inbound.ts:1622-1627` |
| DM OBO v2 → reply to `obo_origin_from_uid` as grantor (bypasses bot-friend gate) | ⚠️ | `inbound.ts:1614-1621` — works for text/typing/error message, **does not cover media replies** (see P1) |
| `obo_system_hint` injected as `GroupSystemPrompt` | ✅ functional, ⚠️ field semantics | `inbound.ts:1571-1575` — note the field is named `GroupSystemPrompt` but is now set in DM payloads too; PR description claims DM persona test passed, so downstream apparently accepts it, but the name is misleading |
| typing keep-alive carries `on_behalf_of` in v2 | ✅ | `inbound.ts:1657-1659` |
| `@AI` (`mention.ais`) triggers `isMentioned` | ✅ | `inbound.ts:1207-1210`, `types.ts:67` |
| `readReceipt` intentionally skipped in OBO v2 | ✅ comment present (`inbound.ts:1640`); confirm with product that no receipt is preferred over a grantor-signed receipt |

## 2. Findings

### P1 — Media replies in OBO v2 silently drop `on_behalf_of`

The reply pipeline threads `effectiveOnBehalfOf` into:

- text `sendMessage` (`inbound.ts:1777`)
- error fallback `sendMessage` (`inbound.ts:1856`)
- `sendTyping` initial + keep-alive (`inbound.ts:1642, 1658`)

…but **not** into the media path. `uploadAndSendMedia` in `inbound.ts:63-181` has no `onBehalfOf` parameter, and `sendMediaMessage` in `api-fetch.ts:67-116` does not accept or forward `on_behalf_of` at all. When the agent decides to reply with an image / file / voice / video while OBO v2 is active:

- **DM v2 (bob ↔ admin, bot only friended by admin)**: the media call will hit the same bot-friend gate that text was designed to bypass, so the upload silently fails downstream — exactly the gate this PR exists to dodge.
- **Group v2**: the media message is posted under the bot's own identity, breaking the "reply as grantor" guarantee that the rest of this PR sets up. The recipient sees a bot avatar attaching the file next to text that was authored by "admin" — a visible identity split.

The "Testing" section in the PR description only exercises text replies, so this gap won't show up in manual smoke tests.

**Fix sketch:**

```ts
// api-fetch.ts — sendMediaMessage
export async function sendMediaMessage(params: {
  ...
  mentionUids?: string[];
  mentionEntities?: MentionEntity[];
  onBehalfOf?: string;       // NEW
  signal?: AbortSignal;
}) {
  ...
  return await postJson<SendMessageResult>(params.apiUrl, params.botToken, "/v1/bot/sendMessage", {
    channel_id: params.channelId,
    channel_type: params.channelType,
    payload,
    ...(params.onBehalfOf ? { on_behalf_of: params.onBehalfOf } : {}),
  }, params.signal);
}

// inbound.ts — uploadAndSendMedia: add onBehalfOf to params + pass through to sendMediaMessage
// inbound.ts:1808 — call site: ...uploadAndSendMedia({ ..., onBehalfOf: effectiveOnBehalfOf })
```

If product policy is "v2 does not support media replies yet", make that explicit: short-circuit the media branch when `isOBOv2 && outboundMediaUrls.length > 0` and emit a single structured log/error, rather than silently posting under the wrong identity.

### P2 — `mention.ais` has no opt-out symmetric to `ignoreMentionAll`

`inbound.ts:1210`:

```ts
isMentioned = (!account.config.ignoreMentionAll && mentionAll)
              || mentionAis
              || mentionUids.includes(botUid);
```

`mention.all` already has an account-level opt-out (`config-schema.ts:13,29,54,72`, `accounts.ts:23,90`) because broadcast pings spam every bot in a noisy room. `mention.ais` has exactly the same fan-out property (one `@AI` triggers every adapter in the group) and is a new vector that didn't exist before this PR. Recommend adding a parallel `ignoreMentionAis` flag now, while the schema is already being touched — adding it later is harder because the silent-default has shipped.

### P2 — `obo_origin_channel_type` strict-numeric guard silently downgrades to Group

`inbound.ts:1613`:

```ts
const resolvedChannelType = (typeof oboV2OriginChannelType === "number"
  ? oboV2OriginChannelType
  : ChannelType.Group) as ChannelType;
```

If a future serializer ever ships the field as `"1"` / `"2"` (common after JSON round-trip through some queues), or omits it entirely, a DM origin would silently be treated as Group. The reply would then land on `oboV2OriginChannel` as a "group" address, which for a DM uid is a routing error and may also pick the wrong `replyChannelId` branch (the DM uses `obo_origin_from_uid`, group does not). Prefer a Number-coercion + explicit enum check, or fail loud:

```ts
const raw = oboV2OriginChannelType;
const num = typeof raw === "number" ? raw : Number.isFinite(Number(raw)) ? Number(raw) : NaN;
if (![ChannelType.DM, ChannelType.Group, ChannelType.CommunityTopic].includes(num)) {
  log?.warn?.(`octo: OBO v2 unknown channel_type=${String(raw)}, refusing to route`);
  return;
}
const resolvedChannelType = num as ChannelType;
```

### P2 — `GroupSystemPrompt` is now written from DM payloads too

`inbound.ts:1573-1575` sets the field unconditionally on the `ctxPayload` regardless of `isGroup`. The field name implies group context, and the sibling adapter at `openclaw-channel-dmwork/src/inbound.ts:1474` continues to hard-code it to `undefined`. PR description says DM persona was manually verified, so downstream apparently honors the field, but the naming is now misleading. Either rename to `SystemPromptOverride` / `PersonaSystemPrompt` (cross-adapter touch), or add a one-line comment at the call site documenting that the field is no longer group-exclusive in this adapter.

### P2 — No unit tests added for any of the new branches

`inbound.test.ts` already mocks `sendMessage` / `sendTyping` / `uploadAndSendMedia`. The new branches that are entirely untested:

- `isOBOv2` detection (presence of `obo_origin_channel_id` + `obo_respond_as`, plus the `obo_grantor_uid` fallback)
- DM vs Group `replyChannelId` selection
- `effectiveOnBehalfOf` propagation into `sendMessage` / `sendTyping` / `onError` fallback
- `mention.ais` triggering `isMentioned`
- `obo_system_hint` → `GroupSystemPrompt`

Given this is the routing layer for a new identity model, a handful of small unit tests asserting the outbound call shape (`expect(sendMessage).toHaveBeenCalledWith(expect.objectContaining({ onBehalfOf: "admin-uid", channelId: "bob-uid", channelType: ChannelType.DM }))`) would lock in the contract cheaply and catch the P1 regression above.

### Nits

- `inbound.ts:1658` — the keep-alive uses `...(isOBOv2 ? { onBehalfOf: effectiveOnBehalfOf } : {})`. Since `effectiveOnBehalfOf` is already `undefined` in the v1 branch and `sendTyping` only inlines `on_behalf_of` when truthy, you can just pass `onBehalfOf: effectiveOnBehalfOf` directly and drop the ternary.
- `inbound.ts:1855` — the user-facing error string is hard-coded Chinese. Pre-existing, not introduced here, but worth a flag for the OBO grantor whose audience may not be CN.
- `inbound.ts:1640` — "v2: skip readReceipt" is a deliberate choice; if product later wants a grantor-signed receipt, the comment makes that a single-line change. Good.

## 3. Recommendations

1. **Block on P1**: thread `on_behalf_of` through `sendMediaMessage` / `uploadAndSendMedia`, or explicitly short-circuit media replies in OBO v2 with a clear log. Add the corresponding test.
2. Add `ignoreMentionAis` to `OctoChannelConfig` and `OctoAccountConfig` (P2) — cheaper now than after rollout.
3. Harden `obo_origin_channel_type` coercion (P2).
4. Cover the new routing branches with unit tests in `inbound.test.ts` (P2).

## 4. Additional Observations

- The fan-out `obo_*` fields all live as untyped `[key: string]: unknown` extras on `MessagePayload` (`types.ts:90`). That's pragmatic but it means the only contract between octo-server #109 and this adapter is field names + types asserted via runtime `typeof` checks. Consider promoting `obo_origin_channel_id`, `obo_origin_channel_type`, `obo_origin_from_uid`, `obo_respond_as`, `obo_grantor_uid`, `obo_system_hint` to a named interface so the next change on either side gets a compile-time anchor.
- `extractMentionUids` is not modified, so `mention.ais` does not flow into any of the `mentionUids`-based logic (history gating, reply-mention list). This appears correct (`@AI` is not a real uid), just calling it out.

Jerry-Xin (changes requested):
Summary: The PR is in scope for this repository, but the new OBO path has blocking routing and outbound identity bugs.

🔴 Blocking

🔴 Critical — OBO origin routing is computed too late, so group fan-out can use the wrong agent/session.  
In `openclaw-channel-octo/src/inbound.ts:1057`, `isGroup` and `sessionId` are derived from the fan-out message channel, and `resolveAgentRoute` runs at `openclaw-channel-octo/src/inbound.ts:1439` before OBO fields are read at `openclaw-channel-octo/src/inbound.ts:1597`. For an OBO group message delivered through a non-group fan-out channel, the adapter will select a direct route/session, skip group member/history/GROUP.md handling, set `ChatType: "direct"`, and record the inbound message under the wrong session, while only the final outbound send goes to the origin group. This can dispatch to the wrong configured agent and lose group context. Derive the effective origin channel/type before mention gating, route resolution, context construction, and session recording.

🔴 Critical — Media replies do not carry `on_behalf_of`.  
Text replies pass `effectiveOnBehalfOf` at `openclaw-channel-octo/src/inbound.ts:1768`, but media replies call `uploadAndSendMedia` at `openclaw-channel-octo/src/inbound.ts:1808` without any OBO identity. `uploadAndSendMedia` also has no `onBehalfOf` parameter at `openclaw-channel-octo/src/inbound.ts:63`, and `sendMediaMessage` does not serialize `on_behalf_of` at `openclaw-channel-octo/src/api-fetch.ts:67`. Any OBO response containing only media, or media plus text, will send media as the bot or fail the same gate this PR is intended to bypass. Thread `onBehalfOf` through `uploadAndSendMedia` and `sendMediaMessage`.

💬 Non-blocking

🟡 Warning — The new behavior lacks direct regression tests. Please add coverage for `sendMessage`/`sendTyping` `on_behalf_of` serialization, OBO group and DM reply targets, system hint propagation, typing target identity, and `mention.ais` mention gating.

🟡 Warning — `obo_origin_channel_type` is accepted only when it is a number at `openclaw-channel-octo/src/inbound.ts:1613`; any string value defaults to group routing. If the backend contract is not strictly numeric, normalize and validate this field.

✅ Highlights

The `on_behalf_of` serialization for text and typing is small and consistent with existing `postJson` patterns.

I could not run the targeted tests locally because dependencies are not installed in this checkout; `npm test` failed with `sh: vitest: command not found`.

lml2468 (changes requested):
## Code Review — PR #57

**Verdict: CHANGES_REQUESTED.** Two P0 security blockers and one P1 functional gap.

### P0 — Security Blockers

**1. No sender trust gate on OBO v2 fields (line 1598-1606)**

```ts
const isOBOv2 = Boolean(
  typeof oboV2OriginChannel === "string" && oboV2OriginChannel.length > 0 &&
  typeof oboV2RespondAs === "string" && oboV2RespondAs.length > 0
);
```

`isOBOv2` is activated purely by payload field presence — **no `from_uid` check**. Any sender who can DM the bot can craft a message with `obo_origin_channel_id` + `obo_respond_as` and:
- Redirect the bot's reply to an arbitrary channel
- Set `on_behalf_of` to impersonate any user

This is the **exact same vulnerability** that PR #35 fixed for OBO v1 via the trust gate (`from_uid === configuredGrantor`). The v2 path must apply the same defense:

```ts
const oboV2Trusted = Boolean(
  account.config.onBehalfOf &&
  message.from_uid === account.config.onBehalfOf
);
const isOBOv2 = oboV2Trusted && Boolean(
  typeof oboV2OriginChannel === "string" && oboV2OriginChannel.length > 0 &&
  typeof oboV2RespondAs === "string" && oboV2RespondAs.length > 0
);
```

**2. `obo_system_hint` injected unconditionally (line 1573-1575)**

```ts
GroupSystemPrompt: (typeof message.payload?.obo_system_hint === "string" && ...)
  ? message.payload.obo_system_hint
  : undefined,
```

This runs for EVERY inbound message, not just OBO messages. Combined with issue #1, any sender can inject arbitrary system-level LLM instructions via a crafted payload. Must be gated:
- Either move below `isOBOv2` computation and condition on it
- Or rewrite the context field after the trust check

lml2468 already identified both issues — I'm independently confirming they are real blockers. The server "stripping non-OBO fields" assumption is undocumented and cannot be trusted at the adapter layer (defense in depth).

### P1 — Functional Gap

**3. Media replies missing `on_behalf_of` (line 1808)**

`uploadAndSendMedia` is called without `onBehalfOf`:
```ts
await uploadAndSendMedia({
  mediaUrl, apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, log,
});
```

Neither `uploadAndSendMedia` nor `sendMediaMessage` in `api-fetch.ts` accept an `onBehalfOf` parameter. If the LLM responds with an image in OBO mode, it will either fail the bot-friend gate or send as the bot's own identity, breaking the OBO contract.

Fix: thread `onBehalfOf` through `uploadAndSendMedia` → `sendMediaMessage` → API body, consistent with `sendMessage`.

### Non-blocking (P2)

4. **`ignoreMentionAll` does not suppress `mentionAis`** — likely intentional (bots that opt out of `@all` still want `@AI`), but add a brief comment confirming this is by design.

5. **Session/agent routing uses fan-out context, not origin** — `isGroup`, `sessionId`, `ChatType`, agent route all derive from the fan-out DM channel. For an OBO group message, the bot uses DM routing/context instead of group routing. Acceptable for initial v2 scope; document as known limitation and track for v2.1.

6. **`obo_origin_channel_type` accepts only numeric values** — string values silently default to group. Add validation or document the server contract.

### Agreement with Existing Reviews

- lml2468: fully agree on both security issues
- Jerry-Xin (Allen): agree on media gap (#3); disagree that "routing too late" is a correctness blocker — it's a known limitation, the replies still reach the right destination

---

yujiawei:
Superseded by PR#61 (merged 2026-05-22). All OBO v2 routing + persona clone functionality is included in PR#61.

PR #58, opened 2026-05-21
chore: improve dependabot config — add labels, groups, timezone, PR limit

lml2468:
Align dependabot config with org-wide standard:

- Add `open-pull-requests-limit: 5` to control PR volume
- Add `labels` for triage routing
- Add `groups` to batch minor/patch npm updates
- Pin schedule to Tuesday, Asia/Shanghai timezone
- Add commit-message prefix for npm ecosystem

Part of org CI/CD governance hardening (Phase 2).

PR #62, opened 2026-05-21
fix: add 'dependencies' label to release-drafter config

lml2468:
Dependabot applies the `dependencies` label to its PRs. The current release-drafter config only lists `dependencies-changed` in the CI/Infrastructure category, so Dependabot PRs don't appear in auto-generated release notes.

## Change
```diff
-    labels: [ci, infra, dependencies-changed]
+    labels: [ci, infra, dependencies-changed, dependencies]
```

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -13,7 +13,7 @@ categories:
   - title: "Bug Fixes"
     labels: [bug, fix]
   - title: "CI and Infrastructure"
-    labels: [ci, infra, dependencies-changed]
+    labels: [ci, infra, dependencies-changed, dependencies]
   - title: "Documentation"
     labels: [documentation, docs]
 exclude-labels: [skip-changelog]


yujiawei (approved):
# Code Review — PR #62 (octo-adapters)

## Summary

One-line change to `.github/release-drafter.yml` that adds the `dependencies` label to the "CI and Infrastructure" category. The change is correct, scoped, and addresses a real gap in the release notes generation pipeline.

```diff
-    labels: [ci, infra, dependencies-changed]
+    labels: [ci, infra, dependencies-changed, dependencies]
```

## Verification

- ✅ **Root cause confirmed.** `.github/dependabot.yml` configures both the `github-actions` and `npm` ecosystems to tag every PR with the `dependencies` label (lines under `labels:` in each `updates` block). Before this change, `release-drafter.yml` only listened for `dependencies-changed` (which is applied by the `PR Labeler` workflow on diff content) — so Dependabot PRs that did not also trigger that workflow would silently fall into the uncategorized bucket and miss the "CI and Infrastructure" section of auto-generated release notes.
- ✅ **Label exists.** Repo label list includes both `dependencies` and `dependencies-changed`, so the new entry resolves to a real label rather than dead config.
- ✅ **No category collision.** `dependencies` is not used by any other category in `release-drafter.yml`. Release Drafter routes a PR to the first matching category, so adding the label here cannot accidentally pull other PRs out of `Features` or `Bug Fixes`.
- ✅ **YAML structure intact.** The list remains a flow-style sequence; only one element was appended. No indentation, anchor, or other key was touched.
- ✅ **`exclude-labels: [skip-changelog]` still respected.** Behavior on PRs explicitly opted out of the changelog is unchanged.

## Findings

### P0 / P1
None.

### P2 / Nits (non-blocking, optional follow-ups)
- The category groups CI/infra and dependency-bump PRs together, which is fine for a small repo. If the volume of Dependabot PRs grows, consider splitting "Dependencies" into its own category for more readable release notes — but that is a future product decision, not a defect in this PR.
- `dependencies-changed` and `dependencies` can both be applied to the same PR (the former by the PR Labeler workflow on a dependency-file diff, the latter by Dependabot directly). Release Drafter de-duplicates by PR number, so this does not double-count; just noting the overlap so future readers don't try to "clean it up" by removing one of them — both serve different signal sources.

## Verdict

Change is minimal, correct, and immediately useful. No blockers.

Jerry-Xin (approved):
Summary: The PR is in scope for `Mininglamp-OSS/octo-adapters` and correctly updates release automation so Dependabot PRs labeled `dependencies` are categorized in generated release notes.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

🔵 Suggestion addressed: `.github/release-drafter.yml:16` now matches the actual Dependabot labels configured in `.github/dependabot.yml:11-12` and `.github/dependabot.yml:24-25`.

The change is limited to release-note categorization, has no runtime, security, performance, or error-handling impact, and does not require additional tests.

lml2468 (commented):
[APPROVE — cannot self-approve, leaving as COMMENT]

One-line YAML fix: adds `"dependencies"` to the "CI and Infrastructure" category in release-drafter, so Dependabot PRs (now labelled `dependencies` via the config introduced in #58) will appear in the changelog. Correct and complete. ✅

---

**⚠️ Unrelated: `welcome / welcome` check failed — pre-existing workflow bug**

The failure is not caused by this PR. Root cause from the run log:
1. The org membership check (`GET /orgs/Mininglamp-OSS/members/lml2468`) returns 403 (private org membership or token scope), so the workflow falls through to posting a welcome comment.
2. The `createComment` call then fails: `403 Resource not accessible by integration` (the GITHUB_TOKEN from `pull_request_target` lacks the required scope for this code path).

This is a systemic issue that will affect any PR opened by an org member whose membership cannot be verified via the token. Suggest opening a follow-up issue to fix `reusable-pr-contributor-welcome.yml` — either by ensuring `issues: write` is included in the job permissions, or by treating 403 on the membership check as "assume member, skip" instead of "proceed with welcome".

lml2468 (commented):
[APPROVE — cannot self-approve via bot account, leaving as COMMENT]

## Code Review — PR #62

**Verdict: APPROVED.** One-line config fix, correct and complete.

Adds `dependencies` label to the "CI and Infrastructure" category in `release-drafter.yml`, matching the label configured in `dependabot.yml` (PR #58). Without this, Dependabot PRs would be uncategorized in release notes.

The `welcome / welcome` CI failure is unrelated (PR welcome workflow, not a code check).

No blockers.

---

PR #69, opened 2026-05-22
feat(persona): inject persona_prompt via before_prompt_build hook

### LINKED ISSUE #68: persona_prompt 未通过 before_prompt_build hook 注入，群聊和私聊均不生效
## 问题

用户设置的分身回复风格 prompt（`persona_prompt`）在群聊和私聊场景下均未生效。DB 中值已正确写入，但 LLM 的 system prompt 中未包含任何 persona 相关内容。

### 当前架构（失败链路）

`persona_prompt` 走的是 `GroupSystemPrompt` 链路：
```
octo-server fanout → WuKongIM payload(obo_system_hint)
  → adapter 解析 message.payload.obo_system_hint
  → 设到 inbound context 的 GroupSystemPrompt
  → OpenClaw get-reply 读取 → 拼入 system prompt
```

这条链路有两个断点：

1. **群聊**：当 bot 和 grantor 都在同一个群时，`obo_fanout.go` Gate 4 跳过 fan-out（避免重复投递）。adapter 走 `buildPersonaGroupSystemPrompt()` fallback，该函数生成硬编码通用提示，**完全不含 persona_prompt**。
2. **私聊**：fan-out 执行了，`obo_system_hint` 写入 payload，但经 WuKongIM NoPersist=1/SyncOnce=1 到 bot sync API 再到 adapter 的链路上，GroupSystemPrompt 未出现在 LLM trajectory 中 — 实测确认 LLM system prompt 无任何 persona 内容。

### 正确架构参考（GROUP.md 的做法）

GROUP.md 不走 GroupSystemPrompt 链路，而是通过 before_prompt_build hook 注入：
```
hook 触发 → getGroupMdForPrompt(ctx) → 返回 { prependContext: "..." }
  → OpenClaw prepare.runtime 直接 prepend 到 system prompt
```

完全不依赖 WuKongIM payload 传递，每次 LLM 调用前都会触发。

### 1. octo-server：新增 bot-token API 获取 persona_prompt

新增 GET /v1/bot/obo-grant 端点（bot token 认证），返回当前 bot 的 active grant 信息，包含 persona_prompt。

### 2. octo-adapter：通过 before_prompt_build hook 注入 persona prompt

在 adapter 的 before_prompt_build hook 中加入 persona_prompt 注入（与 GROUP.md 同机制）。

数据来源：adapter 启动时检测 onBehalfOf 配置 → 调 GET /v1/bot/obo-grant 拉取 persona_prompt 并缓存 → 定期刷新。

### 3. 保留 GroupSystemPrompt 作为备用

inbound.ts 中原有的 obo_system_hint → GroupSystemPrompt 逻辑保留作为 fallback。

## 涉及仓库

- **octo-server** — 新增 bot-token 端点 GET /v1/bot/obo-grant
- **octo-adapters** — hook 注入 + persona prompt fetch/cache 模块

## 涉及文件

- octo-server/modules/bot_api/obo_api.go — 新增 bot-token 端点
- octo-adapters/openclaw-channel-octo/src/index.ts — hook 注入
- octo-adapters/openclaw-channel-octo/src/inbound.ts — persona prompt 缓存管理
- 新增 octo-adapters/openclaw-channel-octo/src/persona-prompt.ts — fetch/cache 模块

## 验证

- [ ] 群聊：bot 在群中，@grantor → 回复符合 persona_prompt 风格
- [ ] 私聊：u_bob 给 admin 发消息 → bot 回复符合 persona_prompt 风格
- [ ] 修改 persona_prompt 后下次回复使用最新 prompt
- [ ] persona_prompt 为空时回复无异常

yujiawei:
## Summary

Persona-clone bots (`account.config.onBehalfOf` set) currently rely on the OBO v2 fan-out copy to deliver `persona_prompt` as `obo_system_hint` → `GroupSystemPrompt`. That path only fires when messages travel through the fan-out envelope, so direct group / DM messages addressed to the persona-clone bot silently drop the `persona_prompt`.

This PR adds a second, complementary path that mirrors how `GROUP.md` is injected: an active-pull cache hydrated from `GET /v1/bot/obo-grant` and prepended to the system prompt via the existing `before_prompt_build` hook.

## Changes

- **`src/persona-prompt.ts`** (new) — per-account cache. Polls `GET /v1/bot/obo-grant` once at account start and refreshes every 60s. No-op when `onBehalfOf` is undefined. Network failures are logged and the last good cache is served so message processing never blocks.
- **`src/api-fetch.ts`** — `getBotOboGrant()` helper for the new endpoint (octo-server YUJ-1762). 404 / `has_grant=false` / inactive grants all collapse to "no hint".
- **`index.ts` `before_prompt_build` hook** — after GROUP.md (section 1) and member/history context (section 2), pushes the composed persona hint as section 3 when the cache has a value.
- **`src/channel.ts`** — wires `initPersonaPromptCache()` into the account start path and `stopPersonaPromptCache()` into the cleanup path.

## Hint format

Mirrors `buildFanoutCopyReq` in `octo-server/modules/bot_api/obo_fanout.go` so the OBO v2 path and this hook path produce structurally identical system prompts:

```
你正在以「<grantor_name>」的分身身份运作。请以 <grantor_name> 的身份回复。

<persona_prompt>
```

Channel-specific context (group name / sender name) is omitted because the hook fires per session, not per message — adding it would require coupling to inbound state and would not be available on the very first message of a session. The OBO v2 fan-out path retains the richer per-message hint.

## Backward compatibility

The legacy `inbound.ts obo_system_hint → GroupSystemPrompt` path is **preserved unchanged** as a fallback (per issue #68 禁改清单). When both paths fire (OBO v2 fan-out into a persona clone), the LLM sees two slightly redundant system messages — duplicate priming is harmless and is what the existing fan-out clients already produce.

## Dependencies

Depends on octo-server `GET /v1/bot/obo-grant` (YUJ-1762, in progress). Until that ships, persona clones running this build will keep logging `persona_prompt fetch failed (404)` every 60s and gracefully fall back to the legacy `GroupSystemPrompt` path. The 404 path is exercised in unit tests so the rollout is non-breaking even when paired with an older server.

## Tests

- 15 new unit tests in `src/persona-prompt.test.ts` covering: hint composition (5 cases), cache population from 200, has_grant=false clearing the cache, 404 handling, no-op when `onBehalfOf` is undefined, 5xx error swallowing + cache retention, Authorization header shape, periodic refresh timer, and timer replacement on repeated init.
- All 840 existing tests still pass.
- `tsc --noEmit` clean.

## Manual integration verification (post-merge, requires YUJ-1762)

1. **Group chat**: bot in a group with grantor → @grantor → bot reply matches `persona_prompt` style.
2. **DM**: `u_bob` → admin → bot reply matches `persona_prompt` style.
3. **Live edit**: change `persona_prompt` via the grantor UI → wait ≤60s → next reply uses the new prompt without process restart.

Closes #68

### DIFF
--- a/openclaw-channel-octo/index.ts
+++ b/openclaw-channel-octo/index.ts
@@ -12,8 +12,8 @@ import type { OpenClawPluginApi } from "openclaw/plugin-sdk";
 import { dmworkPlugin } from "./src/channel.js";
 import { setDmworkRuntime } from "./src/runtime.js";
 import { getGroupMdForPrompt } from "./src/group-md.js";
-import { pendingInboundContext, sessionAccountMap } from "./src/inbound.js";
-import { getPersonaPromptForSession } from "./src/persona-prompt.js";
+import { pendingInboundContext, sessionAccountMap, buildSessionAccountKey } from "./src/inbound.js";
+import { resolvePersonaHintForSession } from "./src/persona-prompt.js";
 import {
   inProcessConfigReader,
   runDoctorChecks,
@@ -303,15 +303,28 @@ const plugin: {
       // when this bot is not a persona clone, the lookup returns undefined
       // and we skip.
       //
-      // The hook ctx does not expose accountId, so we resolve it through
-      // sessionAccountMap (populated by handleInboundMessage in inbound.ts).
-      // Using ctx.accountId directly returns undefined — confirmed by
-      // PR#69 review.
-      const accountIdForPersona = sessionKey ? sessionAccountMap.get(sessionKey) : undefined;
-      if (accountIdForPersona) {
-        const personaHint = getPersonaPromptForSession(accountIdForPersona);
-        if (personaHint) systemSections.push(personaHint);
-      }
+      // The hook ctx does not expose accountId, so we cannot key the
+      // persona cache by sessionKey alone — two persona-clone bots
+      // running on the same node can legitimately share a sessionKey
+      // (OpenClaw routes per-account but session keys can collide).
+      // Keying sessionAccountMap only by sessionKey lets a later inbound
+      // overwrite an earlier one and the hook then attaches the WRONG
+      // account's persona prompt — a cross-account identity leak called
+      // out in PR#69 R3 (Jerry-Xin).
+      //
+      // Fix: sessionAccountMap is composite-keyed by
+      // `${accountId}:${sessionKey}` (see inbound.ts). The resolver
+      // iterates every registered persona account and asks
+      // sessionAccountMap whether it has been seen on this sessionKey.
+      // On 0 / >1 matches we fail safe to "no persona injection".
+      const personaHint = sessionKey
+        ? resolvePersonaHintForSession({
+            sessionKey,
+            hasAccountSession: (accountId, sk) =>
+              sessionAccountMap.has(buildSessionAccountKey(accountId, sk)),
+          })
+        : undefined;
+      if (personaHint) systemSections.push(personaHint);
 
       if (contextSections.length === 0 && systemSections.length === 0) return;
       return {
--- a/openclaw-channel-octo/index.ts
+++ b/openclaw-channel-octo/index.ts
@@ -12,7 +12,7 @@ import type { OpenClawPluginApi } from "openclaw/plugin-sdk";
 import { dmworkPlugin } from "./src/channel.js";
 import { setDmworkRuntime } from "./src/runtime.js";
 import { getGroupMdForPrompt } from "./src/group-md.js";
-import { pendingInboundContext } from "./src/inbound.js";
+import { pendingInboundContext, sessionAccountMap } from "./src/inbound.js";
 import { getPersonaPromptForSession } from "./src/persona-prompt.js";
 import {
   inProcessConfigReader,
@@ -270,12 +270,19 @@ const plugin: {
 
     console.log('[octo] registering before_prompt_build hook');
     api.on('before_prompt_build', (_event, ctx) => {
-      const sections: string[] = [];
+      // Sections destined for the user-prompt context block (group MD,
+      // member list, inbound history). These belong to the conversation
+      // surface, not the LLM's system identity.
+      const contextSections: string[] = [];
+      // Sections destined for the LLM system prompt (persona identity).
+      // System-level identity instructions must NOT live in the user-prompt
+      // prefix or the model can treat them as quotable content.
+      const systemSections: string[] = [];
 
       // 1. Group/Thread MD — wrapped in [GROUP CONTEXT] block
       const groupMdContent = getGroupMdForPrompt(ctx);
       if (groupMdContent) {
-        sections.push(`[GROUP CONTEXT]\n${groupMdContent}\n[/GROUP CONTEXT]`);
+        contextSections.push(`[GROUP CONTEXT]\n${groupMdContent}\n[/GROUP CONTEXT]`);
       }
 
       // 2. Inbound context (member list + history) — outside [GROUP CONTEXT], keeps original format
@@ -284,24 +291,33 @@ const plugin: {
         const pending = pendingInboundContext.get(sessionKey);
         if (pending) {
           pendingInboundContext.delete(sessionKey);
-          if (pending.memberListPrefix) sections.push(pending.memberListPrefix);
-          if (pending.historyPrefix) sections.push(pending.historyPrefix);
+          if (pending.memberListPrefix) contextSections.push(pending.memberListPrefix);
+          if (pending.historyPrefix) contextSections.push(pending.historyPrefix);
         }
       }
 
       // 3. Persona prompt (GH octo-adapters#68) — for persona-clone bots
       // (account.config.onBehalfOf set), pull the active persona_prompt
-      // from the per-account cache and prepend it. The cache is hydrated
-      // by initPersonaPromptCache() in channel.ts; when this bot is not
-      // a persona clone, the lookup returns undefined and we skip.
-      const accountIdForPersona = (ctx as { accountId?: string }).accountId;
+      // from the per-account cache and prepend it to the SYSTEM prompt.
+      // The cache is hydrated by initPersonaPromptCache() in channel.ts;
+      // when this bot is not a persona clone, the lookup returns undefined
+      // and we skip.
+      //
+      // The hook ctx does not expose accountId, so we resolve it through
+      // sessionAccountMap (populated by handleInboundMessage in inbound.ts).
+      // Using ctx.accountId directly returns undefined — confirmed by
+      // PR#69 review.
+      const accountIdForPersona = sessionKey ? sessionAccountMap.get(sessionKey) : undefined;
       if (accountIdForPersona) {
         const personaHint = getPersonaPromptForSession(accountIdForPersona);
-        if (personaHint) sections.push(personaHint);
+        if (personaHint) systemSections.push(personaHint);
       }
 
-      if (sections.length === 0) return;
-      return { prependContext: sections.join('\n\n') };
+      if (contextSections.length === 0 && systemSections.length === 0) return;
+      return {
+        ...(contextSections.length > 0 ? { prependContext: contextSections.join('\n\n') } : {}),
+        ...(systemSections.length > 0 ? { prependSystemContext: systemSections.join('\n\n') } : {}),
+      };
     });
   },
 };
--- a/openclaw-channel-octo/index.ts
+++ b/openclaw-channel-octo/index.ts
@@ -13,6 +13,7 @@ import { dmworkPlugin } from "./src/channel.js";
 import { setDmworkRuntime } from "./src/runtime.js";
 import { getGroupMdForPrompt } from "./src/group-md.js";
 import { pendingInboundContext } from "./src/inbound.js";
+import { getPersonaPromptForSession } from "./src/persona-prompt.js";
 import {
   inProcessConfigReader,
   runDoctorChecks,
@@ -286,6 +287,17 @@ const plugin: {
           if (pending.memberListPrefix) sections.push(pending.memberListPrefix);
           if (pending.historyPrefix) sections.push(pending.historyPrefix);
         }
+      }
+
+      // 3. Persona prompt (GH octo-adapters#68) — for persona-clone bots
+      // (account.config.onBehalfOf set), pull the active persona_prompt
+      // from the per-account cache and prepend it. The cache is hydrated
+      // by initPersonaPromptCache() in channel.ts; when this bot is not
+      // a persona clone, the lookup returns undefined and we skip.
+      const accountIdForPersona = (ctx as { accountId?: string }).accountId;
+      if (accountIdForPersona) {
+        const personaHint = getPersonaPromptForSession(accountIdForPersona);
+        if (personaHint) sections.push(personaHint);
       }
 
       if (sections.length === 0) return;
--- a/openclaw-channel-octo/src/api-fetch.ts
+++ b/openclaw-channel-octo/src/api-fetch.ts
@@ -640,6 +640,69 @@ export async function deleteVoiceContext(params: {
   });
 }
 
+// ---- OBO Grant API (persona clone introspection) ----
+
+/**
+ * Bot-side view of its own OBO grant — used by persona clones to fetch
+ * the active `persona_prompt` so it can be injected into the LLM system
+ * prompt via the before_prompt_build hook (GH octo-adapters#68).
+ *
+ * Returned by GET /v1/bot/obo-grant (octo-server YUJ-1762). The bot is
+ * identified by its botToken; the server resolves the grant where this
+ * bot is the grantee.
+ */
+export interface BotOboGrant {
+  /** False / absent when the bot has no active grant (regular non-persona bot). */
+  has_grant: boolean;
+  grantor_uid?: string;
+  grantor_name?: string;
+  persona_prompt?: string;
+  /** Whether the grant is currently active (mode != "paused" & not revoked). */
+  active?: boolean;
+}
+
+/**
+ * GET /v1/bot/obo-grant — fetch this bot's own OBO grant info.
+ *
+ * Returns null when:
+ *  - the bot has no grant (404)
+ *  - the server reports has_grant=false
+ *  - the response is malformed
+ *
+ * Throws on transport / 5xx errors so the caller's retry-on-next-tick
+ * cadence (see persona-prompt.ts) can decide whether to log and skip.
+ */
+export async function getBotOboGrant(params: {
+  apiUrl: string;
+  botToken: string;
+}): Promise<BotOboGrant | null> {
+  const url = `${params.apiUrl.replace(/\/+$/, "")}/v1/bot/obo-grant`;
+  const resp = await fetch(url, {
+    method: "GET",
+    headers: { Authorization: `Bearer ${params.botToken}` },
+    signal: AbortSignal.timeout(DEFAULT_TIMEOUT_MS),
+  });
+  // 404 = no grant for this bot (regular bot, not a persona clone).
+  if (resp.status === 404) return null;
+  if (!resp.ok) {
+    const text = await resp.text().catch(() => "");
+    throw new Error(
+      `Bot API GET /v1/bot/obo-grant failed (${resp.status}): ${text || resp.statusText}`,
+    );
+  }
+  const raw = (await resp.json().catch(() => null)) as Record<string, unknown> | null;
+  if (!raw || typeof raw !== "object") return null;
+  const hasGrant = raw.has_grant === true;
+  if (!hasGrant) return null;
+  return {
+    has_grant: true,
+    grantor_uid: typeof raw.grantor_uid === "string" ? raw.grantor_uid : undefined,
+    grantor_name: typeof raw.grantor_name === "string" ? raw.grantor_name : undefined,
+    persona_prompt: typeof raw.persona_prompt === "string" ? raw.persona_prompt : undefined,
+    active: raw.active === true,
+  };
+}
+
 /** Decoded payload from base64 message content */
 interface SyncMessagePayload {
   type?: number;
--- a/openclaw-channel-octo/src/channel.ts
+++ b/openclaw-channel-octo/src/channel.ts
@@ -35,6 +35,7 @@ import { createDmworkManagementTools } from "./agent-tools.js";
 import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
+import { initPersonaPromptCache, stopPersonaPromptCache } from "./persona-prompt.js";
 import path from "node:path";
 import os from "node:os";
 import { mkdir, readFile, writeFile, unlink } from "node:fs/promises";
@@ -792,6 +793,20 @@ export const dmworkPlugin: ChannelPlugin<ResolvedDmworkAccount> = {
         log,
       }).catch(() => {});
 
+      // Start persona_prompt cache refresh loop for persona-clone bots
+      // (GH octo-adapters#68). No-op for regular bots. Polls
+      // GET /v1/bot/obo-grant so persona_prompt edits propagate without
+      // requiring a fan-out copy or process restart.
+      initPersonaPromptCache(
+        {
+          accountId: account.accountId,
+          apiUrl: account.config.apiUrl,
+          botToken: account.config.botToken!,
+          onBehalfOf: account.config.onBehalfOf,
+        },
+        log,
+      );
+
       // Prefetch GROUP.md and group members for all groups (fire-and-forget)
       const groupMdCache = getOrCreateGroupMdCache(account.accountId);
       (async () => {
@@ -1081,6 +1096,7 @@ export const dmworkPlugin: ChannelPlugin<ResolvedDmworkAccount> = {
           socket.disconnect();
           if (heartbeatTimer) { clearInterval(heartbeatTimer); heartbeatTimer = null; }
           if (cooldownReconnectTimer) { clearTimeout(cooldownReconnectTimer); cooldownReconnectTimer = null; }
+          stopPersonaPromptCache(account.accountId);
           ctx.setStatus({
             accountId: account.accountId,
             running: false,
--- a/openclaw-channel-octo/src/inbound.test.ts
+++ b/openclaw-channel-octo/src/inbound.test.ts
@@ -17,6 +17,7 @@ import {
   resolveCommandBody,
   resolveCommandAuthorized,
   pendingInboundContext,
+  sessionAccountMap,
   segmentHistoryEntries,
   type ResolveFileResult,
 } from "./inbound.js";
@@ -1309,6 +1310,36 @@ describe("pendingInboundContext", () => {
     pendingInboundContext.set(key, { historyPrefix: "new", memberListPrefix: "ml" });
     expect(pendingInboundContext.get(key)?.historyPrefix).toBe("new");
     expect(pendingInboundContext.get(key)?.memberListPrefix).toBe("ml");
+  });
+});
+
+describe("sessionAccountMap", () => {
+  beforeEach(() => {
+    sessionAccountMap.clear();
+  });
+
+  it("should store and retrieve accountId by sessionKey", () => {
+    const key = "octo:group:abc";
+    sessionAccountMap.set(key, "bot_account_1");
+    expect(sessionAccountMap.get(key)).toBe("bot_account_1");
+  });
+
+  it("should keep separate entries for different sessionKeys", () => {
+    sessionAccountMap.set("k1", "acct_a");
+    sessionAccountMap.set("k2", "acct_b");
+    expect(sessionAccountMap.get("k1")).toBe("acct_a");
+    expect(sessionAccountMap.get("k2")).toBe("acct_b");
+  });
+
+  it("should overwrite on repeated set for same sessionKey", () => {
+    const key = "octo:dm:reroute";
+    sessionAccountMap.set(key, "acct_old");
+    sessionAccountMap.set(key, "acct_new");
+    expect(sessionAccountMap.get(key)).toBe("acct_new");
+  });
+
+  it("returns undefined for unknown sessionKey", () => {
+    expect(sessionAccountMap.get("nope")).toBeUndefined();
   });
 });
 
--- a/openclaw-channel-octo/src/inbound.test.ts
+++ b/openclaw-channel-octo/src/inbound.test.ts
@@ -18,6 +18,7 @@ import {
   resolveCommandAuthorized,
   pendingInboundContext,
   sessionAccountMap,
+  buildSessionAccountKey,
   segmentHistoryEntries,
   type ResolveFileResult,
 } from "./inbound.js";
@@ -1313,33 +1314,57 @@ describe("pendingInboundContext", () => {
   });
 });
 
-describe("sessionAccountMap", () => {
+describe("sessionAccountMap (composite-keyed)", () => {
   beforeEach(() => {
     sessionAccountMap.clear();
   });
 
-  it("should store and retrieve accountId by sessionKey", () => {
-    const key = "octo:group:abc";
-    sessionAccountMap.set(key, "bot_account_1");
-    expect(sessionAccountMap.get(key)).toBe("bot_account_1");
+  it("buildSessionAccountKey concatenates accountId and sessionKey", () => {
+    expect(buildSessionAccountKey("acct_a", "octo:group:abc")).toBe("acct_a:octo:group:abc");
   });
 
-  it("should keep separate entries for different sessionKeys", () => {
-    sessionAccountMap.set("k1", "acct_a");
-    sessionAccountMap.set("k2", "acct_b");
-    expect(sessionAccountMap.get("k1")).toBe("acct_a");
-    expect(sessionAccountMap.get("k2")).toBe("acct_b");
+  it("stores and retrieves accountId by the composite key", () => {
+    const sessionKey = "octo:group:abc";
+    sessionAccountMap.set(buildSessionAccountKey("bot_account_1", sessionKey), "bot_account_1");
+    expect(sessionAccountMap.get(buildSessionAccountKey("bot_account_1", sessionKey))).toBe("bot_account_1");
   });
 
-  it("should overwrite on repeated set for same sessionKey", () => {
-    const key = "octo:dm:reroute";
-    sessionAccountMap.set(key, "acct_old");
-    sessionAccountMap.set(key, "acct_new");
-    expect(sessionAccountMap.get(key)).toBe("acct_new");
+  it("keeps separate entries for different sessionKeys", () => {
+    sessionAccountMap.set(buildSessionAccountKey("acct_a", "k1"), "acct_a");
+    sessionAccountMap.set(buildSessionAccountKey("acct_b", "k2"), "acct_b");
+    expect(sessionAccountMap.get(buildSessionAccountKey("acct_a", "k1"))).toBe("acct_a");
+    expect(sessionAccountMap.get(buildSessionAccountKey("acct_b", "k2"))).toBe("acct_b");
   });
 
-  it("returns undefined for unknown sessionKey", () => {
-    expect(sessionAccountMap.get("nope")).toBeUndefined();
+  it("does NOT overwrite when two accounts share the same sessionKey (multi-account isolation)", () => {
+    // 🔴 PR#69 R3 regression guard: two distinct accounts can legitimately
+    // share the same sessionKey. The composite-key map must keep both
+    // entries so the hook can disambiguate per-account; otherwise the
+    // second account's persona prompt would leak into the first account's
+    // prompt build.
+    const sharedSessionKey = "agent:default:octo:group:shared_group";
+    sessionAccountMap.set(buildSessionAccountKey("acct_persona_a", sharedSessionKey), "acct_persona_a");
+    sessionAccountMap.set(buildSessionAccountKey("acct_persona_b", sharedSessionKey), "acct_persona_b");
+    expect(sessionAccountMap.get(buildSessionAccountKey("acct_persona_a", sharedSessionKey))).toBe("acct_persona_a");
+    expect(sessionAccountMap.get(buildSessionAccountKey("acct_persona_b", sharedSessionKey))).toBe("acct_persona_b");
+    expect(sessionAccountMap.size).toBe(2);
+  });
+
+  it("overwrites on repeated set for the same (accountId, sessionKey) pair", () => {
+    const key = buildSessionAccountKey("acct_a", "octo:dm:reroute");
+    sessionAccountMap.set(key, "acct_a");
+    sessionAccountMap.set(key, "acct_a"); // idempotent
+    expect(sessionAccountMap.get(key)).toBe("acct_a");
+    expect(sessionAccountMap.size).toBe(1);
+  });
+
+  it("returns undefined for unknown composite keys", () => {
+    expect(sessionAccountMap.get(buildSessionAccountKey("acct_a", "nope"))).toBeUndefined();
+    // And: looking up by raw sessionKey alone (the old buggy access pattern)
+    // never matches a composite-keyed entry — guards against accidental
+    // regression to the pre-fix call site.
+    sessionAccountMap.set(buildSessionAccountKey("acct_a", "octo:group:x"), "acct_a");
+    expect(sessionAccountMap.get("octo:group:x")).toBeUndefined();
   });
 });
 
--- a/openclaw-channel-octo/src/inbound.ts
+++ b/openclaw-channel-octo/src/inbound.ts
@@ -29,6 +29,19 @@ import { randomUUID } from "node:crypto";
 // handleInboundMessage writes here; the hook reads and clears per sessionKey.
 export const pendingInboundContext = new Map<string, { historyPrefix: string; memberListPrefix: string }>();
 
+// Per-session accountId map. The before_prompt_build hook receives ctx.sessionKey
+// but not ctx.accountId, so we record the binding here on every inbound message
+// (right next to pendingInboundContext.set) and the hook resolves accountId by
+// looking up sessionKey. Persona-prompt injection (GH octo-adapters#68) depends
+// on this mapping — without it, getPersonaPromptForSession can never be called
+// with the correct accountId from the hook.
+//
+// Lifetime: entries are kept (not deleted) so the hook works on every prompt
+// build for the session, not just the first one after an inbound message.
+// Sessions are bounded by accounts, so the map size is bounded by the active
+// session count per account — no unbounded growth in practice.
+export const sessionAccountMap = new Map<string, string>();
+
 export type DmworkStatusSink = (patch: {
   lastInboundAt?: number;
   lastOutboundAt?: number;
@@ -1564,6 +1577,11 @@ export async function handleInboundMessage(params: {
   if (historyPrefix || memberListPrefix) {
     pendingInboundContext.set(route.sessionKey, { historyPrefix, memberListPrefix });
   }
+
+  // Record sessionKey → accountId so the before_prompt_build hook can resolve
+  // accountId from ctx.sessionKey (hook ctx does not expose accountId).
+  // Required by persona-prompt injection (GH octo-adapters#68).
+  sessionAccountMap.set(route.sessionKey, account.accountId);
 
   const finalBody = quotePrefix ? (quotePrefix + rawBody) : rawBody;
 
--- a/openclaw-channel-octo/src/inbound.ts
+++ b/openclaw-channel-octo/src/inbound.ts
@@ -29,18 +29,41 @@ import { randomUUID } from "node:crypto";
 // handleInboundMessage writes here; the hook reads and clears per sessionKey.
 export const pendingInboundContext = new Map<string, { historyPrefix: string; memberListPrefix: string }>();
 
-// Per-session accountId map. The before_prompt_build hook receives ctx.sessionKey
-// but not ctx.accountId, so we record the binding here on every inbound message
-// (right next to pendingInboundContext.set) and the hook resolves accountId by
-// looking up sessionKey. Persona-prompt injection (GH octo-adapters#68) depends
-// on this mapping — without it, getPersonaPromptForSession can never be called
-// with the correct accountId from the hook.
+// Per-(account, session) registry. The before_prompt_build hook receives
+// ctx.sessionKey but not ctx.accountId, so we record on every inbound message
+// (right next to pendingInboundContext.set) the fact that some accountId has
+// been seen on a given sessionKey. Persona-prompt injection (GH
+// octo-adapters#68) depends on this — without it, getPersonaPromptForSession
+// can never be called with the correct accountId from the hook.
+//
+// 🔴 Multi-account isolation (PR#69 R3, Jerry-Xin):
+// The map is keyed by the COMPOSITE `${accountId}:${sessionKey}`, NOT by
+// `sessionKey` alone. Two distinct bot accounts (e.g. a persona clone and a
+// regular bot, or two persona clones) running on the same OpenClaw node can
+// legitimately share the same `sessionKey` — OpenClaw routes per-account but
+// the resulting session keys can collide. Keying only by `sessionKey` means
+// the second account's inbound `.set` overwrites the first, and the hook
+// then attaches the wrong account's persona prompt — a cross-account
+// identity leak.
+//
+// With the composite key, both accounts get separate entries and the hook
+// resolves persona identity by iterating the registered persona accounts
+// and checking `sessionAccountMap.has(buildSessionAccountKey(candidate, ctx.sessionKey))`.
+// If exactly one persona account matches we use it; on 0 or >1 matches we
+// fail safe to "no persona injection" rather than risk attaching the wrong
+// identity.
 //
 // Lifetime: entries are kept (not deleted) so the hook works on every prompt
 // build for the session, not just the first one after an inbound message.
-// Sessions are bounded by accounts, so the map size is bounded by the active
-// session count per account — no unbounded growth in practice.
+// Sessions are bounded by accounts, so the map size is bounded by
+// (active accounts × active session keys per account) — no unbounded growth
+// in practice.
 export const sessionAccountMap = new Map<string, string>();
+
+/** Build the composite key used to record `(accountId, sessionKey)` pairs. */
+export function buildSessionAccountKey(accountId: string, sessionKey: string): string {
+  return `${accountId}:${sessionKey}`;
+}
 
 export type DmworkStatusSink = (patch: {
   lastInboundAt?: number;
@@ -1578,10 +1601,12 @@ export async function handleInboundMessage(params: {
     pendingInboundContext.set(route.sessionKey, { historyPrefix, memberListPrefix });
   }
 
-  // Record sessionKey → accountId so the before_prompt_build hook can resolve
-  // accountId from ctx.sessionKey (hook ctx does not expose accountId).
+  // Record (accountId, sessionKey) so the before_prompt_build hook can
+  // resolve persona identity from ctx.sessionKey (hook ctx does not expose
+  // accountId). The composite key is required for multi-account isolation —
+  // see the doc comment on sessionAccountMap above.
   // Required by persona-prompt injection (GH octo-adapters#68).
-  sessionAccountMap.set(route.sessionKey, account.accountId);
+  sessionAccountMap.set(buildSessionAccountKey(account.accountId, route.sessionKey), account.accountId);
 
   const finalBody = quotePrefix ? (quotePrefix + rawBody) : rawBody;
 
--- a/openclaw-channel-octo/src/persona-prompt.test.ts
+++ b/openclaw-channel-octo/src/persona-prompt.test.ts
@@ -1,0 +1,291 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+import {
+  composePersonaHint,
+  getPersonaPromptForSession,
+  initPersonaPromptCache,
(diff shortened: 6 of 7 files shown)


lml2468 (approved):
## Code Review — PR #69

[APPROVE]

**Verdict: APPROVED.** Well-structured feature, solid error handling, no security regressions. CI green, 840 tests pass.

---

### Design verified

The PR correctly solves the gap: OBO v2 fan-out delivers `obo_system_hint` only when messages travel through the relay envelope; direct group/DM messages to the persona clone never got the `persona_prompt`. The active-pull cache via `before_prompt_build` fills that gap cleanly.

Key design decisions that check out:

- **Fail-open on network error:** `refreshPersonaPromptCache` catches exceptions, warns, and returns without clearing the cache — so a transient 5xx never drops the persona hint mid-conversation ✅
- **Fail-safe for non-persona bots:** `initPersonaPromptCache` / `refreshPersonaPromptCache` both short-circuit on `!account.onBehalfOf`, so the entire module is a no-op for regular bots ✅
- **Timer hygiene:** `timer.unref?.()` prevents the refresh loop from keeping the process alive; `stopPersonaPromptCache` is wired into the socket disconnect path ✅
- **Idempotent init:** re-calling `initPersonaPromptCache` for the same account clears the old timer before starting a new one — no timer leak ✅
- **`has_grant=false` / 404 → clears cache:** a revoked grant stops injecting the prompt within one refresh cycle ✅
- **OBO v2 path untouched:** legacy `GroupSystemPrompt` path in `inbound.ts` is preserved as documented (#68 禁改清单) ✅

---

### Security

- `getBotOboGrant` uses the bot's own `botToken` as `Bearer` auth → only the bot's own grant is exposed ✅
- The persona_prompt content comes from server-side grantor config (admin-controlled), not from inbound message payload → no user-injection vector ✅
- When OBO v2 and `before_prompt_build` both fire, the LLM sees two slightly redundant hints — harmless, documented ✅

---

### `api-fetch.ts` additions

`getBotOboGrant` implementation:
- URL strips trailing slashes: `apiUrl.replace(/\/+$/, "")` ✅
- 404 → `null` (no grant); non-OK non-404 → throws (caller's retry cadence handles); malformed JSON → `null` ✅
- All fields type-guarded individually before assignment ✅

---

### `index.ts` hook wiring

```ts
const accountIdForPersona = (ctx as { accountId?: string }).accountId;
```

The type cast is necessary because the SDK's `before_prompt_build` context type doesn't expose `accountId` at the TypeScript level, but it is present at runtime (consistent with existing hook usage patterns in this file). The `if (accountIdForPersona)` guard makes the cast safe. ✅

---

### Test coverage

15 tests in `persona-prompt.test.ts`:
- `composePersonaHint`: all null/undefined/empty/inactive/missing-name branches ✅
- `refreshPersonaPromptCache`: 200/has_grant=false/404/no-op/5xx-retain/auth-header ✅
- Timer behavior: initial + periodic fire, no-op without `onBehalfOf`, repeated-init no-leak ✅
- `_resetPersonaPromptCacheForTests` / `setPersonaPromptRefreshIntervalMs` support complete test isolation ✅

---

### 💬 P2 — Non-blocking

- 🔵 `active === undefined` is treated as active (falls through the `grant.active === false` gate). This is the right default (older server versions may not return `active`), but worth a code comment to make the intent explicit.
- 🔵 `setPersonaPromptRefreshIntervalMs` is exported from the production module. A `/* @internal */` JSDoc annotation or a separate test-only file would signal its test-only scope more clearly. Current `_`-prefix naming convention is adequate for now.
- 🔵 `(ctx as { accountId?: string }).accountId` — worth a follow-up to the SDK maintainers to add `accountId` to the typed `before_prompt_build` context so this cast can be removed.

Jerry-Xin (changes requested):
Relevant PR, but the implementation has blocking correctness issues that prevent the promised persona prompt path from working reliably.

**🔴 Blocking**

🔴 Critical: `before_prompt_build` cannot find the account cache key.  
[index.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/index.ts:297) reads `ctx.accountId`, but the inbound session metadata stores `AccountId` at [inbound.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/src/inbound.ts:1737), and the OpenClaw hook context does not expose this lowercase field. In practice `accountIdForPersona` will be undefined, so direct DM/group messages still will not receive the cached persona prompt. Store a `sessionKey -> accountId` mapping during inbound handling, or otherwise pass/recover the account id through a supported hook context path.

🔴 Critical: the new path uses prompt context, not system context.  
[index.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/index.ts:304) returns `{ prependContext: ... }`, which prepends to the user prompt, while the legacy `GroupSystemPrompt` path is system-level context. This does not produce structurally equivalent system prompting as described in the PR. Persona identity instructions should use `prependSystemContext` or another system-prompt path.

🔴 Critical: stopped or reconfigured accounts can be repopulated by an in-flight fetch.  
[persona-prompt.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/src/persona-prompt.ts:168) starts an uncancelled async refresh, and [persona-prompt.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/src/persona-prompt.ts:125) writes the cache after the await without checking whether the account is still active. If [persona-prompt.ts](/private/tmp/pr-review-wO1USN/repo/openclaw-channel-octo/src/persona-prompt.ts:182) clears state during shutdown/reconfigure, the old request can resolve later and restore a stale persona hint. Add a generation token, active-account registry, or abort signal so stale refreshes cannot write after stop.

**✅ Highlights**

The API helper keeps 404/no-grant handling non-fatal, and the tests cover many useful cache behaviors. The lifecycle cleanup hook is also wired in the expected account stop path.

lml2468 (dismissed):
**Review (R1)** — feat(persona): inject persona_prompt via before_prompt_build hook

### Summary

CI is green, and the non-persona/security shape is mostly right: accounts without `onBehalfOf` no-op, the new active-pull path reads `persona_prompt` from the bot-authenticated `/v1/bot/obo-grant` endpoint rather than from inbound user payloads, and the existing OBO payload path still gates `obo_system_hint` by configured grantor.

However, the new direct-message/group prompt path is still not deliverable. I verified the full hook path against the head branch and `openclaw@2026.5.7`: the hook context does not expose lowercase `accountId`, the returned field injects into user prompt context instead of system prompt context, and the cache lifecycle has stale-write/cold-start gaps.

### Findings

- 🔴 Critical: `before_prompt_build` still cannot find the persona cache key. `openclaw-channel-octo/src/inbound.ts:1720-1738` finalizes the inbound context with `AccountId`, while `openclaw-channel-octo/index.ts:297-300` reads `(ctx as { accountId?: string }).accountId`. In `openclaw@2026.5.7`, `PluginHookAgentContext` includes `runId/jobId/agentId/sessionKey/sessionId/workspaceDir/modelProviderId/modelId/messageProvider/trigger/channelId`, not `accountId`; the run path carries `agentAccountId` internally but the hook context builder only adds `messageProvider` and `channelId`. In practice `accountIdForPersona` is undefined, so direct DM/group messages never read the `persona_prompt` cache. Fix by passing account id through a supported hook context, or by maintaining a `sessionKey -> accountId` map from inbound handling and looking up by `ctx.sessionKey`.

- 🔴 Critical: the persona hint is injected into the user prompt, not the system prompt. `openclaw-channel-octo/index.ts:303-304` returns `{ prependContext: ... }`; OpenClaw applies `prependContext` by prepending it to the effective user prompt, while `prependSystemContext` is what gets composed into the agent system prompt. This is not equivalent to the existing `GroupSystemPrompt` path in `openclaw-channel-octo/src/inbound.ts:1747`, and contradicts this PR's own `persona-prompt.ts` comments saying the hint must reach the LLM system prompt. Return `prependSystemContext` for the persona section, or route through an existing system-context path.

- 🔴 Critical: stopped/reconfigured accounts can be repopulated by an in-flight refresh. `openclaw-channel-octo/src/persona-prompt.ts:168` starts an uncancelled initial fetch, periodic refreshes do the same at `openclaw-channel-octo/src/persona-prompt.ts:170-172`, and `refreshPersonaPromptCache` writes `_cache.set(...)` after the await at `openclaw-channel-octo/src/persona-prompt.ts:111-125` without checking whether the account is still active. `stopPersonaPromptCache` clears the timer and cache at `openclaw-channel-octo/src/persona-prompt.ts:182-187`, but an old request can resolve later and restore stale persona instructions after shutdown, grant change, or account reconfiguration. Add a generation token/active registry/AbortController and test the stale-resolution case.

- 🔴 Major: `active` is optional in the API model but absence is parsed as inactive. `BotOboGrant.active` is declared optional at `openclaw-channel-octo/src/api-fetch.ts:654-661`, and `composePersonaHint` only suppresses the prompt when `grant.active === false` at `openclaw-channel-octo/src/persona-prompt.ts:85-90`. But `getBotOboGrant` returns `active: raw.active === true` at `openclaw-channel-octo/src/api-fetch.ts:697-703`, which turns an omitted/older-server `active` field into `false` and drops an otherwise valid `persona_prompt`. Preserve undefined with `typeof raw.active === "boolean" ? raw.active : undefined`, or make the endpoint contract require `active` and test the missing-field behavior.

- 🟠 Major: cold start has a first-turn race. `initPersonaPromptCache` deliberately fire-and-forgets the initial fetch at `openclaw-channel-octo/src/persona-prompt.ts:166-168`, and `getPersonaPromptForSession` returns undefined until that request completes at `openclaw-channel-octo/src/persona-prompt.ts:190-205`. The first inbound prompt after account start can therefore run without the required persona instructions. If fail-open is intentional, document it and make the UX/product tradeoff explicit; otherwise await the initial hydration before accepting turns for persona-clone accounts or synchronously fetch on first prompt build with bounded timeout/caching.

### Verdict

CHANGES_REQUESTED. The feature shape is reasonable, but the direct prompt injection path does not currently work reliably enough to merge.

yujiawei (commented):
# Code Review — PR #69 (octo-adapters)

## Summary

The persona-prompt cache module (`src/persona-prompt.ts`) is well-built and the 15 unit tests cover its behavior cleanly. However, the integration into the `before_prompt_build` hook in `index.ts` has a runtime-fatal correctness bug: the hint is read by a key that does not exist on the SDK's hook context object, so the persona prompt is never actually injected. The whole feature this PR ships is dead on arrival in production. Blocking on that, plus a few smaller notes.

## Findings

### P0 — `persona_prompt` is never injected: `ctx.accountId` does not exist on `before_prompt_build`

`openclaw-channel-octo/index.ts:297`:

```ts
const accountIdForPersona = (ctx as { accountId?: string }).accountId;
if (accountIdForPersona) {
  const personaHint = getPersonaPromptForSession(accountIdForPersona);
  if (personaHint) sections.push(personaHint);
}
```

The structural cast `(ctx as { accountId?: string })` defeats the type checker, but at runtime there is no `accountId` field on the hook context.

Evidence (against the pinned dev dep `openclaw@^2026.5.4`, which the PR builds against):

- SDK type definition — `PluginHookAgentContext` in `openclaw/dist/plugin-sdk/src/plugins/hook-types.d.ts`:

  ```ts
  export type PluginHookAgentContext = {
      runId?: string; jobId?: string; trace?: DiagnosticTraceContext;
      agentId?: string; sessionKey?: string; sessionId?: string;
      workspaceDir?: string; modelProviderId?: string; modelId?: string;
      messageProvider?: string; trigger?: string; channelId?: string;
  };
  ```

  No `accountId`.

- SDK runtime builder — `openclaw/dist/lifecycle-hook-helpers-9z237Zsl.js:6 buildAgentHookContext()` spreads exactly those 11 fields onto the object handed to every `before_prompt_build` listener. No `accountId` is ever populated.

- Existing convention in this codebase — `src/group-md.ts:345 getGroupMdForPrompt()` does *not* read `ctx.accountId`; it parses `sessionKey` for the groupNo and calls `resolveAccountId(agentId, groupNo)` precisely because the runtime context does not expose `accountId`.

Consequence: `accountIdForPersona` is `undefined` on every call → the `if` is always false → `getPersonaPromptForSession` is never invoked → no persona hint ever reaches `prependContext`. The OBO v2 fan-out → `obo_system_hint` → `GroupSystemPrompt` path remains the only working delivery channel, which is exactly the path the PR set out to complement. The DM / direct group case the PR is supposed to fix still drops `persona_prompt`.

Why CI did not catch it: `persona-prompt.test.ts` exercises `composePersonaHint`, cache population, refresh timer, and `getPersonaPromptForSession` — all in isolation from the hook. There is no test that drives the `before_prompt_build` hook end-to-end with a synthetic `PluginHookAgentContext` and asserts the returned `prependContext` contains the hint. That is the test that would have failed.

Fix direction (not prescriptive — pick what fits the codebase):

1. Mirror what `group-md.ts` does: extract `agentId` and parse `sessionKey` with `OCTO_GROUP_RE`, then resolve `accountId` via `resolveAccountId(agentId, groupNo)`. Works for group/thread sessions only.
2. The PR's primary motivation is DM delivery, where `sessionKey` has no `groupNo`. For that, key the persona cache by something the hook actually has (`agentId`, or `agentId` plus a 1:1 `agentId → accountId` registry populated alongside `initPersonaPromptCache` in `channel.ts`). The current `accountId` keying is fine for the cache itself; you only need an extra lookup path from `agentId` (or `channelId`) at read time.
3. Add an integration test that:
   - seeds the cache via `refreshPersonaPromptCache` for an account,
   - invokes the registered `before_prompt_build` handler with a stub `PluginHookAgentContext` matching what the runtime actually delivers (no `accountId`),
   - asserts `prependContext` contains the persona hint.

   That test must fail on `4ee18d98` and pass after the fix.

### P1 — "structurally identical" claim does not match what's shipped

The PR description says: *"Mirrors `buildFanoutCopyReq` in `octo-server/modules/bot_api/obo_fanout.go` so the OBO v2 path and this hook path produce structurally identical system prompts."* The implementation in `composePersonaHint` (`src/persona-prompt.ts:75-98`) intentionally omits the channel/group/sender context that `buildFanoutCopyReq` includes; the comment block on lines 75-83 acknowledges this. Result: when both paths fire on a persona clone, the LLM sees two system messages with **different** prefixes (channel-agnostic here, channel-specific in the fan-out copy). That is functionally redundant priming, but it is not "structurally identical." Tighten either the PR description or the doc comment so future maintainers don't assume the formats are interchangeable / dedupable.

### P1 — `refreshPersonaPromptCache` will spam logs on a persistent 5xx upstream

`src/persona-prompt.ts:116-121` logs at `warn` on every failed fetch with no dedup. With the documented 60s cadence, a long-broken upstream produces 1440 warn lines per day per persona-clone account. Consider one of:

- compare the new error message to the last logged one for the account and suppress repeats, or
- drop subsequent identical errors to `debug` until success.

The 404 case is already silent after the first cycle (since `getBotOboGrant` returns `null` → cache stays `undefined` → the change-detection branch on line 140 does not fire); the 5xx case lacks that suppression.

### P2 — `setPersonaPromptRefreshIntervalMs` is exported but is a test hook

`src/persona-prompt.ts:63` is exported and mutates a module-global. Pair it with `_resetPersonaPromptCacheForTests` semantics by renaming to `_setPersonaPromptRefreshIntervalMsForTests` (matching the underscore-prefix convention already used on line 209) or mark `@internal` so it is not picked up as a public surface.

### P2 — `changed=${prev?.hint !== hint}` is always `true` in the branch where it logs

`src/persona-prompt.ts:140-144`: the `if (prev?.hint !== hint)` guard makes the `changed` interpolation always `true` inside the branch. Cosmetic — remove the field or compute it outside the guard if you want a richer log line.

### Nit — `account.config.botToken!` non-null assertion

`src/channel.ts:798` uses `!` to coerce `botToken`. The codebase already does this elsewhere, so consistent with house style, but worth flagging that a malformed config would silently send `Bearer undefined` to `/v1/bot/obo-grant` until the upstream rejects it.

## Suggested next steps

1. Replace the `ctx.accountId` lookup with a path that uses what the SDK actually provides (`agentId` plus parsed `sessionKey`, or a parallel `agentId → accountId` registry seeded in `channel.ts`).
2. Add the hook-level integration test described above so this regression cannot recur silently.
3. Tighten the "structurally identical" wording.
4. Add error-dedup to the refresh loop's 5xx log path.

yujiawei:
Pushed fix commit 825745d addressing the 3 blocking reviews:

**B1 — `ctx.accountId` undefined**
Added `sessionAccountMap` (sessionKey → accountId) in `inbound.ts` next to `pendingInboundContext`. `handleInboundMessage` populates it; the hook resolves accountId via `sessionAccountMap.get(ctx.sessionKey)`. Persona prompt now actually reaches the LLM on non-OBO paths.

**B2 — `prependSystemContext` for persona**
Split hook return into `contextSections` (GROUP.md / member list / history → `prependContext`) and `systemSections` (persona hint → `prependSystemContext`). The SDK supports both fields (openclaw CHANGELOG #35177).

**B3 — abort in-flight fetches**
Added per-account `_generation` counter in `persona-prompt.ts`. `initPersonaPromptCache` and `stopPersonaPromptCache` bump generation; `refreshPersonaPromptCache` snapshots it pre-fetch and bails out before mutating `_cache` if it changed. Covers both stop→stale-resolve and reconfigure→stale-resolve races.

**Tests**
- New `sessionAccountMap` unit tests in `inbound.test.ts`.
- New generation-guard race tests (stop-wins / reconfigure-wins) in `persona-prompt.test.ts`.
- Full suite: 846/846 passing. `tsc --noEmit` and `npm run build` clean.

No GROUP.md logic touched, no existing functionality removed.

yujiawei (commented):
# Code Review — PR #69 (octo-adapters)

## Verdict

**APPROVED** (no P0/P1 blockers found). The three review blockers (B1/B2/B3) called out in the follow-up commit are addressed correctly. Suggestions below are P2 / nits and do not block merge.

## 1. Verification of declared fixes

| Item | Status | Evidence |
|---|---|---|
| B1 — `sessionAccountMap` so the hook can resolve `accountId` from `sessionKey` | ✅ | `openclaw-channel-octo/src/inbound.ts:42` (export), `:1582` (populated in `handleInboundMessage`), `openclaw-channel-octo/index.ts:301` (hook lookup). Mirrors the existing `pendingInboundContext` pattern and uses the same key. |
| B2 — persona hint routed to `prependSystemContext` instead of `prependContext` | ✅ | `openclaw-channel-octo/index.ts:273-320`. SDK contract `PluginHookBeforePromptBuildResult` does declare both `prependContext` and `prependSystemContext` (`openclaw/plugin-sdk/.../hook-before-agent-start.types.d.ts:16-30`), and the docstring even calls out the "static plugin guidance" caching use case the persona prompt fits. |
| B3 — generation token aborts stale in-flight fetches around stop/reconfigure | ✅ | `openclaw-channel-octo/src/persona-prompt.ts:75-86, 134-160, 197-216, 230-242`. Both stop and init bump generation; `refreshPersonaPromptCache` snapshots `myGeneration` and bails on mismatch both in the catch branch and post-fetch. The two new race tests in `persona-prompt.test.ts:573-685` cover the stop-wins and reconfigure-wins scenarios. |

## 2. Architectural notes

- **Path duality is intentional**: the legacy `inbound.ts obo_system_hint → GroupSystemPrompt` route is preserved untouched (per the issue #68 禁改清单 / 禁改 list). Confirmed by reading the diff — `inbound.ts` only gains the new `sessionAccountMap` global + one `.set` call; nothing about the legacy hint path is changed.
- **Hint format parity with `buildFanoutCopyReq`**: the channel-agnostic variant in `composePersonaHint` correctly drops per-message channel context (group name / sender name) because the hook fires per-session, not per-message. Sensible trade-off; the OBO v2 fan-out path keeps the richer per-message hint.
- **Failure semantics**: 404 / `has_grant=false` / inactive grants collapse to "no hint" cleanly. 5xx errors keep the last good cache and just `log.warn` — message processing never blocks. Good defensive design.
- **First-message gap**: if an inbound message arrives before the initial `GET /v1/bot/obo-grant` resolves, that first reply will be sent without the persona hint. Not blocking — the next reply (within ≤60s) will have it. Worth noting in operator-facing release notes if there is one.

## 3. Suggestions (P2 / non-blocking)

### S1 — `sessionAccountMap` is never pruned on account stop
`openclaw-channel-octo/src/inbound.ts:31-43` keeps entries for the lifetime of the process. The comment justifies this ("entries are kept (not deleted) so the hook works on every prompt build for the session, not just the first one after an inbound message") and the size is bounded by the active session count, so this is not a real leak in practice. However:

- After `stopPersonaPromptCache(accountId)` runs, any `sessionAccountMap` entry that still points to the stopped account would just no-op on the next prompt build (cache is empty), so there is no correctness issue.
- Long-running gateways that churn through many ephemeral DM sessions across many bots would accumulate entries indefinitely. Consider, in a follow-up, sweeping `sessionAccountMap` entries whose `accountId` equals the stopped account inside `stopPersonaPromptCache` — or building an inverse index. Not required for this PR.

### S2 — `_generation` map grows monotonically across account lifecycles
`openclaw-channel-octo/src/persona-prompt.ts:75-86, 234-238`. `stopPersonaPromptCache` deletes `_cache`, `_timers`, and `_firstFetchLogged` for an account but **deliberately retains** `_generation` (otherwise resetting it to 0 could let a very-late stale fetch slip through after a future init). The trade-off is sound; the leak is `O(unique accountIds ever started)`. No action needed.

### S3 — `composePersonaHint`: `grantor_uid` is not trimmed
`openclaw-channel-octo/src/persona-prompt.ts:115-120`:

```ts
const grantorName = (grant.grantor_name ?? "").trim() || grant.grantor_uid;
if (!grantorName) return undefined;
```

If `grantor_name` is empty/whitespace and `grantor_uid` happens to be whitespace-only (very unlikely server bug), the resulting hint would render `「   」`. Cheap fix: `(grant.grantor_name ?? "").trim() || (grant.grantor_uid ?? "").trim()`. Edge-case only.

### S4 — `composePersonaHint`: `active === undefined` is treated as active
`openclaw-channel-octo/src/persona-prompt.ts:111`: `if (grant.active === false) return undefined;` — this lets a payload that omits `active` entirely through as "active". Worth aligning explicitly with the YUJ-1762 server contract (will the server always include `active`?). If the server can legitimately omit it, the current default is fine; otherwise consider being strict: `if (grant.active !== true) return undefined;`.

### S5 — Race window where two interval ticks can overlap
If a fetch ever exceeds `_refreshIntervalMs` (60s), two interval-triggered refreshes could be in flight at once, both capturing the same `myGeneration`, and the slower one would silently overwrite the faster one (last-write-wins). In practice `DEFAULT_TIMEOUT_MS` in `api-fetch.ts` looks much shorter than 60s, so this cannot happen — but it would be more robust to recursively schedule via `setTimeout` after each fetch settles, instead of `setInterval`. Optional hardening.

### S6 — Comment hygiene
Several comment blocks are notably long (e.g. `inbound.ts:31-43`, `persona-prompt.ts:1-26, 67-86, 124-138`). They are accurate and well-written but slightly heavy for the eventual reader. Up to author taste — happy to leave as is given how subtle the generation guard is.

## 4. Other findings while reading

- `channel.ts:807` passes `account.config.botToken!` (non-null assertion). Pre-existing pattern in this file (the new code matches the surrounding style), so not raised here, just noted.
- `getBotOboGrant` uses `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` and `DEFAULT_TIMEOUT_MS` is unchanged in this PR — assumed to already be defined in `api-fetch.ts`. Confirmed by the patch context (`DEFAULT_TIMEOUT_MS` is referenced, not introduced).
- Tests look thorough: 15 unit tests for the new module covering hint composition (5 cases), cache population, has_grant=false clearing, 404, no-op when `onBehalfOf` undefined, 5xx swallowing, Authorization header shape, periodic refresh timer, init idempotency, and the two B3 race scenarios. Plus 4 new tests for `sessionAccountMap`. The PR body claim of 846 total tests passing matches.

## Summary

Solid follow-up on the three blocking review items. The persona injection path, the system-vs-context split, and the generation guard all check out against the codebase and the openclaw SDK contract. Suggestions above are improvements, not blockers.

Jerry-Xin (changes requested):
One blocking issue: the PR is relevant to `octo-adapters`, but the new persona injection can attach the wrong account’s persona prompt in multi-account sessions.

🔴 Blocking

🔴 Critical — `sessionAccountMap` is keyed only by `sessionKey`, but `sessionKey` is not guaranteed to be account-scoped. The hook resolves persona identity with `sessionAccountMap.get(sessionKey)` in `openclaw-channel-octo/index.ts:310-313`, and the map is populated globally in `openclaw-channel-octo/src/inbound.ts:1581-1584`. For default DM routing, OpenClaw can use the same main session key across accounts for the same agent; group keys also do not necessarily include `accountId`. Since inbound queues are scoped by account in `openclaw-channel-octo/src/channel.ts:123-152`, two accounts can process messages concurrently and overwrite the same session mapping before prompt build runs. Result: account A’s prompt can be built with account B’s persona cache, leaking or applying the wrong system identity instructions. This is a cross-account isolation bug and should be fixed before merge.

A safer design needs an account-scoped lookup at prompt-build time, or the session mapping must be made unambiguous and concurrency-safe. If the hook cannot receive `accountId`, avoid persistent `sessionKey -> accountId` identity for system prompt injection unless collisions are impossible by construction.

💬 Non-blocking

🟡 Warning — Tests cover the cache and basic map operations, but not the actual hook behavior or multi-account/session-key collision case. Add a regression test where two accounts share the same `sessionKey` and verify persona prompt injection cannot cross accounts.

✅ Highlights

The cache lifecycle has good defensive handling for refresh failures and stop/reconfigure races, and `prependSystemContext` is the right prompt surface for persona identity rather than user-context prefixing.

yujiawei (commented):
# Code Review — PR #69 (octo-adapters)

## Verdict

**APPROVED** — no P0/P1 blockers found at head `c26c683`. All three previously‑blocking concerns from earlier reviews (B1 `ctx.accountId` undefined, B2 wrong context surface, B3 stop/reconfigure races) are correctly addressed in commits `825745d` and `c26c683`. The cross‑account isolation regression flagged on `825745d` is also resolved in `c26c683`. Remaining items below are P2 / non‑blocking.

## 1. Verification of fixes against head

| Item | Status | Evidence |
|---|---|---|
| B1 — hook can resolve `accountId` from `sessionKey` | ✅ Fixed | `openclaw-channel-octo/index.ts:320‑326` uses `resolvePersonaHintForSession` keyed by `sessionKey`; `src/inbound.ts:1609` registers `(accountId, sessionKey)` next to `pendingInboundContext.set`. |
| B2 — persona hint routed to **system** prompt, not user context | ✅ Fixed | `openclaw-channel-octo/index.ts:280, 327, 332` separate `systemSections` → `prependSystemContext`, distinct from `contextSections` → `prependContext`. |
| B3 — stop / reconfigure races cannot resurrect a stale cache | ✅ Fixed | `src/persona-prompt.ts:68‑78, 140, 150, 160, 208, 229` introduce a per‑account generation counter bumped on init/stop; `refreshPersonaPromptCache` snapshots `myGeneration` and bails before writing on mismatch (both error and success paths). Tests `persona-prompt.test.ts:405‑585` (stop‑wins / reconfigure‑wins). |
| R3 — composite key prevents cross‑account leak when two persona clones share a `sessionKey` | ✅ Fixed | `src/inbound.ts:61‑66` exposes `buildSessionAccountKey`; `src/persona-prompt.ts:286‑301` resolves by iterating `_timers.keys()` and fails safe on `matches.length !== 1`. Regression test `persona-prompt.test.ts` "two persona accounts share the same sessionKey" asserts `undefined` is returned. |

The wiring in `index.ts:320‑326` correctly passes a closure `sessionAccountMap.has(buildSessionAccountKey(accountId, sk))` rather than a `.get(sessionKey)`, so the old pre‑fix call site cannot silently regress.

## 2. Architectural notes

- **Path duality preserved** — the legacy `inbound.ts obo_system_hint → GroupSystemPrompt` path is untouched (per the #68 禁改清单). Diff confirms `inbound.ts` only gains the new `sessionAccountMap` global + one `.set` call; the legacy hint extraction at the OBO v2 site is intact.
- **System vs context split is correct** — `prependSystemContext` is the right surface for identity instructions; persona content should not be addressable as `prependContext` user‑prompt prefix where the model can treat it as quotable.
- **Resolver is conservative on ambiguity** — `resolvePersonaHintForSession` short‑circuits after the second match (`if (matches.length > 1) return undefined`), so a many‑persona node only pays O(matches) before bailing.
- **Failure semantics** — `getBotOboGrant` collapses 404 / `has_grant=false` / malformed JSON to `null`; 5xx throws to the caller; `refreshPersonaPromptCache` swallows the throw and keeps the last good `_cache` entry. Message processing never blocks on this loop.
- **Lifecycle** — `initPersonaPromptCache` is wired in the account start path (`src/channel.ts:800‑808`) and `stopPersonaPromptCache` in the disconnect cleanup (`src/channel.ts:1099`). Idempotent re‑init clears the old timer (`persona-prompt.ts:202‑203`).

## 3. P1 (worth addressing before merge but not strictly blocking)

### P1 — `getBotOboGrant` coerces an omitted `active` field to `false`

`src/api-fetch.ts:702` parses the response as:

```ts
active: raw.active === true,
```

That turns any response where the server omits `active` (older deployment, future field gating, partial payload) into `active: false`. `composePersonaHint` then rejects the grant on line 110 (`if (grant.active === false) return undefined`). The PR's own type declaration at `api-fetch.ts:660‑661` marks `active` as optional, so the contract anticipates absence — but the parser does not preserve it.

If the YUJ‑1762 contract is "the server always returns `active`", this is fine; tighten the type to required (`active: boolean`) so the contract is explicit. Otherwise, preserve undefined:

```ts
active: typeof raw.active === "boolean" ? raw.active : undefined,
```

and keep `composePersonaHint`'s `=== false` gate as‑is. A unit test that fixes a server response without `active` and asserts the hint is still composed would lock this down. This was raised in the prior R1 review and remains open at `c26c683`.

### P1 — 5xx upstream logs once every 60s per persona‑clone account

`src/persona-prompt.ts:151‑154` logs at `warn` on every failed fetch with no dedup. A long‑broken upstream produces 1440 warn lines / day / persona‑clone account. The 404 branch is already silent after the first cycle (it returns `null` → `hint` stays `undefined` → the change‑detection branch on line 179 does not fire); 5xx lacks that suppression. Compare against the last logged error per account and either suppress repeats or downgrade subsequent identical errors to `debug` until the next success. Also raised in R1, not addressed.

## 4. P2 / nits

- **No end‑to‑end hook integration test.** `resolvePersonaHintForSession` is exercised as a pure helper, and the wiring in `index.ts:320‑326` is small, but no test drives the registered `before_prompt_build` handler with a stub `PluginHookAgentContext` matching what the SDK runtime actually delivers (no `accountId`, with a `sessionKey`) and asserts the returned `prependSystemContext`. The unit tests would still pass if the hook wiring regressed (e.g. someone re‑introduced `ctx.accountId`). Worth one integration test that seeds `sessionAccountMap` + cache and invokes the handler.
- **Composite key separator.** `buildSessionAccountKey` uses `:` which also appears inside `sessionKey` (`agent:<agentId>:octo:group:<groupNo>`). The current form is unambiguous in practice because `accountId` does not contain `:`, but a separator that cannot appear in either field (e.g. `\x00`) — or a `Map<string, Set<string>>` indexed first by `sessionKey` — would be defensively stronger.
- **`sessionAccountMap` not pruned on `stopPersonaPromptCache`.** Long‑running gateways that churn many ephemeral DM sessions across many bots accumulate entries monotonically. Bound is `(active accounts × active session keys)`, so not a true leak, but a sweep of entries whose value equals the stopped accountId on stop would be cheap.
- **`_generation` map intentionally retained on stop** (`src/persona-prompt.ts:223‑235`) to avoid letting a very‑late stale fetch slip through after a future init. Trade‑off is sound, but adds another monotonically‑growing map keyed by every `accountId` ever started. Acceptable; documenting the rationale is already done in‑place.
- **Race window between two overlapping interval ticks.** If a fetch ever exceeds `_refreshIntervalMs` (60s), two interval‑triggered refreshes can be in flight simultaneously and both capture the same generation token; the slower one will overwrite the faster one (last‑write‑wins). In practice `DEFAULT_TIMEOUT_MS` in `api-fetch.ts` is much shorter than 60s so this cannot happen today, but `setTimeout` chaining post‑settlement would remove the residual risk.
- **`composePersonaHint` does not trim `grantor_uid`** (`persona-prompt.ts:113`). If `grantor_name` is whitespace and `grantor_uid` is also whitespace‑only the hint renders `「   」`. Edge case only.
- **`channel.ts:804` `account.config.botToken!`** non‑null assertion. Matches the surrounding style in this file — not introduced by this PR. Noted for completeness.
- **First‑message cold‑start gap.** If an inbound arrives before the initial `GET /v1/bot/obo-grant` resolves, that first reply ships without the persona hint; the next reply (within ≤60s) will have it. Acknowledged in the PR description as accepted. Worth a one‑line operator note in release notes when YUJ‑1762 lands.

## 5. Other observations while reading

- `src/api-fetch.ts:679` `apiUrl.replace(/\/+$/, "")` correctly normalises trailing slashes.
- `getBotOboGrant` uses `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` (timeout already defined upstream).
- Tests look thorough: 15 unit tests for hint composition + cache, the two B3 generation‑guard race tests, 4 new `sessionAccountMap` composite‑key tests, and 6 new `resolvePersonaHintForSession` tests covering single‑match / zero / two‑persona collision (the core R3 regression guard) / mixed persona‑plus‑regular / empty sessionKey / cold‑start. PR body claims 856 passing — consistent with the diff.
- CI: build / type‑check / CodeQL workflows are currently in progress at review time; no broken pre‑existing checks were observed.

## Summary

The architecture is sound, the three blocking review items and the multi‑account isolation issue are correctly resolved, and the test coverage on the touched code is high. The two P1 items (`active` parsing & 5xx log spam) are worth a small follow‑up but are not merge blockers given the current YUJ‑1762 contract.

Jerry-Xin (changes requested):
One-line summary: The PR is relevant to `octo-adapters`, but the persona cache lifecycle can leak stale identity prompts across account reconfiguration, which is a blocking security/correctness issue.

🔴 Blocking

🔴 Critical — Stale persona cache survives reconfiguration to non-persona account  
[openclaw-channel-octo/src/persona-prompt.ts](/private/tmp/pr-review-Buqcjq/repo/openclaw-channel-octo/src/persona-prompt.ts:194): `initPersonaPromptCache()` returns immediately when `onBehalfOf` is missing, but it does not clear any existing timer/cache for the same `accountId`. If an account is changed from persona-clone to regular bot without a full `stopPersonaPromptCache()` first, the old timer remains registered and the old cached persona hint can still be injected. This is an identity leak. Call `stopPersonaPromptCache(account.accountId)` or otherwise clear `_timers`, `_cache`, and `_firstFetchLogged` when re-initializing as non-persona.

🔴 Critical — Repeated init keeps the previous persona hint until the new fetch completes  
[openclaw-channel-octo/src/persona-prompt.ts](/private/tmp/pr-review-Buqcjq/repo/openclaw-channel-octo/src/persona-prompt.ts:200): on repeated init for the same `accountId`, the code clears the old interval and bumps generation, but leaves `_cache` intact while the new fire-and-forget fetch is pending. During that window, `before_prompt_build` can inject the previous grantor’s `persona_prompt` into the new configuration’s sessions. Clear or mark the cache unusable before starting the new fetch, and add a regression test for grantor/token change with a delayed refresh.

💬 Non-blocking

🟡 Warning — `sessionAccountMap` has no cleanup path  
[openclaw-channel-octo/src/inbound.ts](/private/tmp/pr-review-Buqcjq/repo/openclaw-channel-octo/src/inbound.ts:56): entries are intentionally retained for session reuse, but long-running multi-account nodes can accumulate every session key ever seen. Consider cleanup on account stop/session end or bounding by TTL/LRU.

🟡 Warning — I could not run local tests/type-check in this workspace because `vitest` and `tsc` are not installed (`node_modules` missing). The PR’s claimed test results may still be valid in CI.

✅ Highlights

The new system-prompt placement uses the supported `prependSystemContext` hook field, and the resolver’s ambiguous multi-persona behavior fails closed instead of risking cross-account prompt injection.

yujiawei (commented):
# Code Review — PR #69 (octo-adapters)

## Verdict

**APPROVED** at head `685f0f7e` — the two R4 blocking concerns (persona→regular reconfigure leak; in-flight refetch window serving the previous grantor's hint) are addressed correctly, and the cumulative state of the patch is internally consistent against the openclaw SDK contract and the surrounding channel runtime. Remaining items are P1/P2 follow-ups carried over from earlier rounds, not merge blockers.

## 1. Verification of R4 blocking fixes against head `685f0f7e`

| Item | Status | Evidence |
|---|---|---|
| R4-B1 — `initPersonaPromptCache` early-returns without leaking stale persona state when an account is reconfigured from persona-clone → regular bot | ✅ Fixed | `src/persona-prompt.ts:198-206`. The `if (!account.onBehalfOf)` branch now calls `stopPersonaPromptCache(account.accountId)` before returning. `stopPersonaPromptCache` bumps generation, clears `_timers`, `_cache`, and `_firstFetchLogged`. New regression test at `src/persona-prompt.test.ts:754-791` ("clears the cached hint when the account is reconfigured to drop onBehalfOf") asserts both `getPersonaPromptForSession` returns `undefined` and `getRegisteredPersonaAccountIds()` no longer contains the account. |
| R4-B2 — re-init with a different grantor does not serve the previous grantor's hint during the new fetch's in-flight window | ✅ Fixed | `src/persona-prompt.ts:217-227`. Immediately after bumping generation, `_cache.delete(account.accountId)` and `_firstFetchLogged.delete(...)` run synchronously, before `void refreshPersonaPromptCache(...)`. `getPersonaPromptForSession` therefore returns `undefined` during the gap until the new fetch resolves. New regression test at `src/persona-prompt.test.ts:793-852` ("returns undefined during the refetch window when re-init switches grantor") explicitly holds the second fetch open via a manually-resolved promise and asserts the cache is empty during the in-flight window, then matches the new grantor (`"bob voice"`) once the fetch resolves and never contains the old grantor (`"alice voice"`). |

The fix shape is also internally consistent: the generation guard inside `refreshPersonaPromptCache` (`src/persona-prompt.ts:142, 159`) means that even if a *previous* in-flight fetch from the old grantor resolves after the new init's eager `_cache.delete`, the post-await branch will see a generation mismatch and bail out, so the old grantor's hint cannot resurrect the cache. The interaction between Fix 2 (eager delete on reconfigure) and the existing generation guard is the right layered defence.

## 2. Verification that prior approvals at `c26c683` still hold at `685f0f7e`

The HEAD diff vs `c26c683` only touches `src/persona-prompt.ts` (the `initPersonaPromptCache` early-return path) and `src/persona-prompt.test.ts` (two new tests). The previously-validated fixes remain intact:

- **B1 — hook resolves `accountId` from `sessionKey`**: `index.ts:320-326` still uses `resolvePersonaHintForSession({ sessionKey, hasAccountSession })` with a closure over the composite-keyed `sessionAccountMap.has(buildSessionAccountKey(...))`. Unchanged.
- **B2 — persona hint routed to `prependSystemContext`, not `prependContext`**: `index.ts:280, 327-332` still separate `systemSections` from `contextSections`. Unchanged.
- **B3 — generation-token guard for stop/reconfigure races**: `src/persona-prompt.ts:75-86, 140-160` still bump generation on init/stop and snapshot it in `refreshPersonaPromptCache`. The new Fix 2 layers on top of (does not replace) this guard.
- **R3 — composite-key resolver fails safe on multi-persona sessionKey collisions**: `src/persona-prompt.ts:286-301` still iterates `_timers.keys()` and returns `undefined` on `matches.length !== 1`. Regression test "two persona accounts share the same sessionKey" still present.

## 3. Architectural notes (re-verified at HEAD)

- **Path duality preserved.** The legacy `inbound.ts obo_system_hint → GroupSystemPrompt` path is untouched (per the issue #68 禁改清单). `inbound.ts` only gains the `sessionAccountMap` global and one `.set` call adjacent to `pendingInboundContext.set`; the OBO v2 hint-extraction site is intact.
- **`prependSystemContext` is the right surface** for identity instructions — persona content should not be reachable as a `prependContext` user-prompt prefix where the model can quote or argue against it. SDK contract `PluginHookBeforePromptBuildResult` (openclaw plugin-sdk) does declare both fields.
- **Resolver is conservative on ambiguity.** `resolvePersonaHintForSession` short-circuits after the second match (`if (matches.length > 1) return undefined`), so a many-persona node only pays O(matches) before bailing.
- **Failure semantics.** `getBotOboGrant` collapses 404 / `has_grant=false` / malformed JSON to `null`; non-OK non-404 throws; `refreshPersonaPromptCache` catches and keeps the last good `_cache` entry. Message processing never blocks on this loop.
- **Lifecycle.** `initPersonaPromptCache` is wired in the account start path (`src/channel.ts:796-808`) and `stopPersonaPromptCache` in the disconnect cleanup (`src/channel.ts:1099`). Idempotent re-init clears the old timer first.
- **Tradeoff acknowledgement.** The eager `_cache.delete` in Fix 2 means every (re-)init — even when the grantor hasn't actually changed — opens a brief window where `getPersonaPromptForSession` returns `undefined`. Any inbound message arriving during the gap will be answered without the persona hint, then subsequent messages get it back within one refresh tick. This is the correct tradeoff against the identity-leak risk that Fix 2 closes, but it is operationally visible on hot reconnects. Worth a one-line operator note.

## 4. P1 items still open from prior rounds

These were flagged in R1 / R3 reviews and remain unaddressed at this HEAD. They are not strict merge blockers given the current YUJ-1762 server contract, but should be on the follow-up list.

### P1 — `getBotOboGrant` coerces an omitted `active` field to `false`

`src/api-fetch.ts:702`:

```ts
active: raw.active === true,
```

If the server omits `active` (older deployment, partial payload, future field gating), this turns it into `false`. `composePersonaHint` then rejects the grant via `if (grant.active === false) return undefined` (`src/persona-prompt.ts:110`). The PR's own type at `src/api-fetch.ts:660-661` marks `active` as optional, so the contract anticipates absence — but the parser doesn't preserve it.

Two acceptable resolutions:
1. **Tighten the contract**: declare `active: boolean` (required) and confirm YUJ-1762 always returns it. Add a server-side test that the field is always populated.
2. **Preserve undefined**: `active: typeof raw.active === "boolean" ? raw.active : undefined`, with a unit test that fixes a server response without `active` and asserts the hint is still composed.

Pick one; the inconsistency between the optional-type declaration and the strict parser is the real bug.

### P1 — 5xx upstream logs once every 60s per persona-clone account

`src/persona-prompt.ts:151-154` logs at `warn` on every failed fetch with no dedup. A long-broken upstream produces ~1,440 warn lines/day/persona-clone account. The 404 branch is silent after the first cycle (it returns `null` → `hint` stays `undefined` → the change-detection branch on line 179 doesn't fire); 5xx lacks that suppression.

Fix: compare against the last logged error per account and either suppress repeats or downgrade subsequent identical errors to `debug` until the next success.

## 5. P2 / nits

- **No end-to-end hook integration test.** The wiring at `index.ts:320-326` is small but untested. `resolvePersonaHintForSession` is exercised as a pure helper and `sessionAccountMap` is exercised structurally, but no test drives the registered `before_prompt_build` handler with a stub `PluginHookAgentContext` matching what the SDK runtime actually delivers (no `accountId`, with a `sessionKey`) and asserts the returned `prependSystemContext` contains the hint. The unit tests would still pass if someone re-introduced `ctx.accountId` lookup or swapped `prependSystemContext` back to `prependContext`. One integration test would lock the hook wiring down.
- **Composite key separator collision.** `buildSessionAccountKey` (`src/inbound.ts:65`) uses `:` which also appears inside `sessionKey` (`agent:<agentId>:octo:group:<groupNo>`). Current form is unambiguous in practice because `accountId` does not contain `:`, but a separator that cannot appear in either field (e.g. `\x00`) — or a `Map<string, Set<string>>` indexed first by `sessionKey` — would be defensively stronger.
- **`sessionAccountMap` not pruned on `stopPersonaPromptCache`.** Long-running gateways that churn many ephemeral DM sessions across many bots accumulate entries monotonically. Bound is `(active accounts × active session keys)`, so not a true leak, but a sweep of entries whose value equals the stopped accountId on stop would be cheap.
- **`_generation` map retained on stop** (`src/persona-prompt.ts:223-235`). Trade-off is sound (avoids letting a very-late stale fetch slip through after a future init), but adds another monotonically-growing map keyed by every `accountId` ever started. With the new R4 Fix 1, `_generation` is now also bumped for accounts that were never persona clones (because the `!onBehalfOf` early-return now calls `stopPersonaPromptCache`, which calls `_bumpGeneration`). Bound is `O(unique accountIds ever started)`. Acceptable, but the new growth surface (regular bots, not just persona clones) is worth a follow-up comment in code.
- **Race window between two overlapping interval ticks.** If a fetch ever exceeds `_refreshIntervalMs` (60s), two interval-triggered refreshes can be in flight simultaneously and both capture the same generation token; the slower one will overwrite the faster one (last-write-wins). In practice `DEFAULT_TIMEOUT_MS` in `api-fetch.ts` is much shorter than 60s so this cannot happen today, but `setTimeout` chaining post-settlement would remove the residual risk.
- **`composePersonaHint` does not trim `grantor_uid`** (`src/persona-prompt.ts:118`). If `grantor_name` is whitespace and `grantor_uid` is also whitespace-only the hint renders `「   」`. Edge case only — server contract presumably prevents this.
- **Cold-start gap on first message.** If an inbound arrives before the initial `GET /v1/bot/obo-grant` resolves, that first reply ships without the persona hint; the next reply (within ≤60s) will have it. Acknowledged in the PR description as accepted. Worth a one-line operator note in release notes when YUJ-1762 lands.

## 6. Other observations while reading

- `src/api-fetch.ts:679` `apiUrl.replace(/\/+$/, "")` correctly normalises trailing slashes.
- `getBotOboGrant` uses `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` (timeout already defined upstream).
- `channel.ts:804` `account.config.botToken!` non-null assertion matches the surrounding style in this file — not introduced by this PR.
- Tests: 15 unit tests for hint composition + cache, 2 B3 generation-guard race tests, 4 `sessionAccountMap` composite-key tests, 6 `resolvePersonaHintForSession` tests covering single-match / zero / multi-persona collision / mixed persona-plus-regular / empty sessionKey / cold-start, plus 2 new R4 tests for the reconfigure paths. PR body claims 856 passing — consistent with the diff.
- The new R4 test for the persona→regular switch (`src/persona-prompt.test.ts:754-791`) also asserts `getRegisteredPersonaAccountIds().not.toContain(...)`, which provides a useful secondary check that the timer is genuinely de-registered, not just that the cache is cleared.

## Summary

The two R4 blocking concerns are correctly addressed at HEAD `685f0f7e`. The eager-delete + generation-guard layering is sound, the new tests cover both reconfigure paths, and prior fixes from earlier rounds remain intact. The two P1 follow-ups (`active` parsing & 5xx log dedup) and the small P2 set are improvements rather than merge blockers given the current YUJ-1762 contract.

Jerry-Xin (approved):
The PR is in scope for `octo-adapters` and the implementation is mergeable; I found no blocking correctness or security issues.

💬 Non-blocking

- 🟡 Warning: `openclaw-channel-octo/src/api-fetch.ts:697-703` normalizes `active` to `false` whenever the field is absent. Since `BotOboGrant.active` is typed optional and `composePersonaHint()` only treats explicit `false` as inactive, an otherwise valid `{ has_grant: true, persona_prompt: ... }` response without `active` would silently suppress persona injection. Consider preserving `undefined` unless `raw.active` is actually boolean, or adding a test that locks the intended API contract.

- 🟡 Warning: `openclaw-channel-octo/src/inbound.ts:56-61` and `openclaw-channel-octo/src/inbound.ts:1604-1609` keep `sessionAccountMap` entries forever. That makes persona injection available across turns, but the “no unbounded growth” comment is stronger than the code guarantees for long-lived processes with many unique DM/group session keys. A TTL, session-end cleanup, or account-stop cleanup would make the lifecycle explicit.

✅ Highlights

- The new `prependSystemContext` usage in `openclaw-channel-octo/index.ts:330-332` matches the declared OpenClaw hook contract.
- Cache stop/reconfigure handling in `persona-prompt.ts` has good stale-fetch guards and avoids serving an old grantor prompt during reconfiguration.
- Focused tests are strong: I ran `npm run type-check` and `npx vitest run src/persona-prompt.test.ts src/inbound.test.ts`; both passed.

lml2468 (approved):
**Re-review (R2)** — Previous blockers: P0 accountId lookup, P0 prependContext vs prependSystemContext, P0 stale refresh race

### Previous Findings Status

- [x] **P0 ctx.accountId undefined** — Fixed. Hook now uses `ctx.sessionKey` (which IS available in `PluginHookAgentContext`) + composite-keyed `sessionAccountMap` populated during inbound handling. `resolvePersonaHintForSession` iterates registered persona accounts and checks membership via `hasAccountSession` closure. Multi-account collision fails safe (0 or >1 matches → undefined). Verified composite key: `buildSessionAccountKey("acct", "sk")` → `"acct:sk"`. Test coverage includes cross-account isolation, regular-bot coexistence, and cold-start undefined cases.

- [x] **P0 prependContext → prependSystemContext** — Fixed. Hook now separates `contextSections` (group MD, member list, history → `prependContext`) from `systemSections` (persona identity → `prependSystemContext`). This correctly routes persona instructions to system prompt via `composeSystemPromptWithHookContext` rather than user prompt prefix.

- [x] **P0 stale refresh after stop/reconfigure** — Fixed. Added `_generation` counter bumped by both `initPersonaPromptCache` and `stopPersonaPromptCache`. `refreshPersonaPromptCache` captures generation before fetch, checks after await — stale results are silently dropped. Tests cover stop-during-flight and reconfigure-during-flight scenarios with explicit resolution ordering.

### Non-blocking observations (not duplicating Allen's flags)

- 🟡 `active: raw.active === true` still normalizes absent/undefined to `false`, which `composePersonaHint`'s `grant.active === false` gate then treats as inactive. If the server omits `active` on older versions, valid persona_prompts are silently dropped. Suggested: `typeof raw.active === "boolean" ? raw.active : undefined`. Allen flagged this same point — tracking as follow-up.

- 🟡 Cold-start first-turn race is still present (fire-and-forget initial fetch). The code comments document the intent — not blocking.

- 🟡 `initPersonaPromptCache` now eagerly clears `_cache` + `_firstFetchLogged` on reconfigure, and handles persona→regular transitions by calling `stopPersonaPromptCache` first. Both are correct defensive behaviors.

### Verdict

APPROVED. All three P0 blockers are properly resolved. The generation guard, composite-key isolation, and system/context prompt separation are clean and well-tested.

PR #72, opened 2026-05-23
feat: add check-sprint-on-merge workflow

lml2468:
## Summary

Adds a caller workflow that invokes the org-level reusable `check-sprint` workflow on every PR to `main`.

Merges will be **blocked** if the PR's Sprint field on the [Octo Board](https://github.com/orgs/Mininglamp-OSS/projects/2) is not set to the current sprint iteration.

See [Mininglamp-OSS/.github #42](https://github.com/Mininglamp-OSS/.github/pull/42) for the reusable workflow implementation.

> **Note for deployer:** Enable the required status check `check-sprint / check-sprint` on the `main` branch protection *after* this PR is merged.

### DIFF
--- a/.github/workflows/check-sprint.yml
+++ b/.github/workflows/check-sprint.yml
@@ -1,0 +1,24 @@
+# Calls the org-level reusable workflow to block merge when the PR's
+# Sprint field on the Octo Board is missing or doesn't match the
+# current sprint iteration.
+#
+# Uses pull_request_target so the workflow can access secrets for the
+# Projects API. No code from the PR is checked out or executed.
+
+name: Check Sprint
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [main]
+
+permissions: {}
+
+jobs:
+  check-sprint:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main
+    with:
+      pr-number: ${{ github.event.pull_request.number }}
+      repo-name: ${{ github.repository }}
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}
--- a/.github/workflows/check-sprint.yml
+++ b/.github/workflows/check-sprint.yml
@@ -1,9 +1,33 @@
 # Calls the org-level reusable workflow to block merge when the PR's
-# Sprint field on the Octo Board is missing or doesn't match the
+# Sprint field on the Octo Board is missing or does not match the
 # current sprint iteration.
 #
-# Uses pull_request_target so the workflow can access secrets for the
-# Projects API. No code from the PR is checked out or executed.
+# Uses pull_request_target so the workflow can access PROJECT_TOKEN for
+# the GitHub Projects API. No PR code is checked out or executed here.
+#
+# KNOWN LIMITATION — stale required status after sprint/field change:
+#   This check re-evaluates only on pull_request_target events
+#   (opened, synchronize, reopened). Two scenarios can leave a stale
+#   green status on branch protection:
+#
+#   1. Sprint rollover: the org's active sprint advances while this PR
+#      has no new commits. Branch protection may allow merge on the old
+#      green result from the previous sprint.
+#      Partial mitigation: enforce-freshness: true (see below) rejects
+#      PRs whose head commit predates the current sprint start.
+#
+#   2. Sprint field edited via Projects UI after a green run: no
+#      pull_request* event fires, so the check is not re-run.
+#      No automated mitigation exists (GitHub Projects has no webhook
+#      events that can trigger Actions workflows).
+#
+# RECOMMENDED MITIGATION for both cases:
+#   In branch protection for main, enable
+#   "Require branches to be up to date before merging".
+#   A new push re-triggers this check, keeping the result fresh.
+#
+# See the reusable workflow header for full technical details:
+#   Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml
 
 name: Check Sprint
 
@@ -20,5 +44,6 @@ jobs:
     with:
       pr-number: ${{ github.event.pull_request.number }}
       repo-name: ${{ github.repository }}
+      enforce-freshness: true
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}


lml2468 (commented):
[COMMENT]

Clean caller workflow — no issues found.

**What it does:** Adds a `pull_request_target` caller that invokes the org-level `reusable-check-sprint` workflow on every PR to `main`, blocking merge when the Sprint field is missing or stale.

**Security:**
- `pull_request_target` is appropriate here: secrets are needed for the Projects API, and no PR code is checked out or executed. ✅
- `permissions: {}` drops all default token scopes — good practice. ✅

**Inputs / secrets:** `pr-number` and `repo-name` are correctly derived from the event context; `PROJECT_TOKEN` is passed through as a secret. ✅

**Minor note (non-blocking):** The reusable workflow is pinned to `@main`. For an org-internal `.github` repo this is fine and easier to maintain; a SHA pin would be stricter but is not necessary here.

LGTM 🚀

Jerry-Xin (changes requested):
This PR is in scope for `Mininglamp-OSS/octo-adapters`, but the workflow currently has blocking reliability issues for the intended merge gate.

🔴 Blocking

- 🔴 Critical — [.github/workflows/check-sprint.yml:19](/Users/mlamp/.openclaw/tmp/pr-review-qck7xxld/repo/.github/workflows/check-sprint.yml:19): The caller references `Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main`, but that reusable workflow is not present on the org `.github` repo’s `main` branch yet; it only exists in the referenced implementation PR. If this merges first, this workflow will fail to resolve/run for every PR. Merge the reusable workflow first, or otherwise ensure the referenced workflow exists on `main` before landing this caller.

- 🔴 Critical — [.github/workflows/check-sprint.yml:11](/Users/mlamp/.openclaw/tmp/pr-review-qck7xxld/repo/.github/workflows/check-sprint.yml:11): The check only runs on `opened`, `synchronize`, and `reopened`, but the validation target is time-dependent and project-field-dependent. A PR can pass during one sprint, remain open into the next sprint without new commits, and still keep a green required check at merge time. Similarly, Project Sprint field changes do not trigger this workflow. That means the branch protection check can become stale and allow a merge that no longer matches the current sprint. The enforcement needs a strategy that revalidates close to merge time, such as a scheduled/status-refresh mechanism for open PRs, a merge-queue/merge-group compatible gate if applicable, or another required check that is updated when sprint state changes.

💬 Non-blocking

- 🔵 Suggestion — [.github/workflows/check-sprint.yml:11](/Users/mlamp/.openclaw/tmp/pr-review-qck7xxld/repo/.github/workflows/check-sprint.yml:11): Consider adding the same `# zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed` annotation used by nearby `pull_request_target` workflows. The security model is documented in the file, but the annotation keeps workflow scanning consistent with the repository’s existing pattern.

✅ Highlights

- The PR is relevant to this repository because it adds a repository-level merge policy workflow.
- `permissions: {}` is appropriate for this metadata-only caller.
- The workflow passes only PR metadata into the reusable workflow and does not check out or execute PR code.

yujiawei (dismissed):
# Code Review — PR #72 (octo-adapters)

Small, focused PR (1 file / 24 lines) that wires this repo into the org-level reusable Sprint-discipline check. The implementation is correct, the security model is sound, and it follows the same pattern already established by `.github/workflows/codeql.yml` in this repo. Approving with a few non-blocking notes for the rollout.

## What it does

Adds `.github/workflows/check-sprint.yml`, a thin caller that invokes `Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main` on every `pull_request_target` against `main` and passes the PR number + repo slug plus the `PROJECT_TOKEN` secret.

## Verification

| Item | Status | Evidence |
|---|---|---|
| YAML is valid and lint-clean | ✅ | PR checks: `sanity / actionlint` PASS, `sanity / No tabs in workflow files` PASS |
| `pull_request_target` is the correct trigger | ✅ | Needs secrets for Projects API; no PR code is checked out or executed (`check-sprint.yml:11–14` rationale comment) |
| Workflow-level `permissions: {}` | ✅ | `check-sprint.yml:15` — GITHUB_TOKEN carries zero scopes; all API access is through `PROJECT_TOKEN` PAT |
| Trigger scope limited to PRs targeting `main` | ✅ | `check-sprint.yml:11–13` |
| Caller `uses:` ref consistent with existing repo convention | ✅ | `codeql.yml:25` already pins the same org reusable repo at `@main` |
| Required-check name matches GitHub's `<caller-job-id> / <reusable-job-name>` rule | ✅ | Caller job id `check-sprint` (`check-sprint.yml:18`) + reusable job name `check-sprint` ⇒ `check-sprint / check-sprint`, matching the deployer note |
| Inputs and secret declarations match the reusable workflow's contract | ✅ | Reusable `workflow_call` expects `pr-number` (number), `repo-name` (string), and secret `PROJECT_TOKEN` — all supplied |

## Findings

### P2 — Cross-repo merge ordering is load-bearing (not blocking, but worth calling out)

`uses: Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main` resolves at run time. The reusable workflow is being introduced in `Mininglamp-OSS/.github#42`, which at the time of this review is still **open** (not merged to `main`).

Consequence if this PR is merged first: every subsequent PR run will fail the workflow load with "workflow file not found", visible as a red check on PRs until `#42` lands. It will not block merges (the deployer note correctly defers enabling branch protection until after merge), but it will create noise.

Suggestion: extend the deployer note from
> Enable the required status check `check-sprint / check-sprint` on the `main` branch protection *after* this PR is merged.

to something like
> Merge `Mininglamp-OSS/.github#42` **first**, then merge this PR, then enable the required status check `check-sprint / check-sprint` on `main`'s branch protection.

### P2 — Consider also firing on `ready_for_review`

The trigger list is `[opened, synchronize, reopened]`. A PR opened as a draft and later marked ready-for-review will not re-fire this check unless a subsequent push triggers `synchronize`. `codeql.yml:7` in this same repo already lists `ready_for_review` for exactly this reason. Adding it here would keep parity and avoid a stale-check edge case for draft → ready transitions with no further commits.

### Nit — Pinning to a tag/SHA vs `@main`

Pinning org reusables to `@main` is the established pattern in this repo (`codeql.yml:25`), so this PR is consistent and I'm not asking to change it. Just noting that the org could choose to pin reusable workflows to immutable refs (SHA or version tag) in the future for tighter supply-chain hygiene, since the reusable workflow runs with access to `PROJECT_TOKEN`. Trust boundary is internal-org-only, so the marginal risk is low.

### Nit — Header comment phrasing

`check-sprint.yml:1–6` is clear and useful. Consider one tiny addition: explicitly note that the reusable workflow lives in `Mininglamp-OSS/.github` (currently only implied by the `uses:` line), so a future reader doesn't have to scroll down to find out where the actual logic lives.

## Out-of-scope observations (informational, not blockers for this PR)

While reading the reusable workflow source (`Mininglamp-OSS/.github#42`) for context, two things stood out that should be addressed in *that* PR's review, not here:

1. The Projects v2 GraphQL `iterations` field returns only currently-configured iterations, not completed ones. If the current calendar day falls in a gap between sprints (or before iteration #1 is configured), `current_sprint` becomes `None` and the workflow fails for every PR. Worth confirming the iteration schedule has no gaps, or handling that case explicitly.
2. `get_today_shanghai()` shells out to `/usr/bin/date` with `TZ=Asia/Shanghai`. Equivalent in pure Python via `zoneinfo.ZoneInfo("Asia/Shanghai")` would remove the subprocess and avoid a host-dependency.

Neither affects this PR — flagging only so they can be raised on `#42`.

## Conclusion

Code is correct, security model is appropriate, and the change is consistent with existing conventions in this repo. Approving. The only real coordination item is making the cross-repo merge order explicit in the deployer note so the rollout is smooth.

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-adapters` and adds a narrowly scoped repository governance workflow that matches the existing reusable-workflow pattern.

💬 Non-blocking
- 🔵 Suggestion — [.github/workflows/check-sprint.yml](/Users/mlamp/.openclaw/tmp/pr-review-0zttm_ye/repo/.github/workflows/check-sprint.yml:35): Consider adding `ready_for_review` or `edited` if the team wants the check to run when a draft PR becomes ready or when a PR is retargeted to `main`. Current required-check behavior should still block merges without a passing status, but users may need to push/reopen/rerun manually in those cases.
- 🔵 Suggestion — [.github/workflows/check-sprint.yml](/Users/mlamp/.openclaw/tmp/pr-review-0zttm_ye/repo/.github/workflows/check-sprint.yml:43): Using `@main` is consistent with this repo’s org-level workflow pattern, but it means future changes to the reusable workflow affect this merge gate immediately. A version tag or SHA would be safer if deterministic policy behavior is required.

✅ Highlights
- `pull_request_target` is justified here because the workflow only calls a reusable metadata/API workflow and does not check out or execute PR code.
- `permissions: {}` keeps the default `GITHUB_TOKEN` minimized.
- The caller inputs and `PROJECT_TOKEN` secret match the referenced reusable workflow contract.
- YAML parsed successfully. `actionlint` was not installed locally, so I could not run that static check.

yujiawei (approved):
# Code Review — PR #72 (octo-adapters)

Re-review after the `fix(ci): document stale-check limitation; add enforce-freshness` update (`bc6db9f`). The two blocking items raised previously have been addressed sufficiently to merge. Approving with three non-blocking suggestions for the deployment rollout.

## Verification of prior blockers

| Prior blocking item | Status now | Evidence |
|---|---|---|
| Cross-repo dependency: `reusable-check-sprint.yml@main` must exist on `Mininglamp-OSS/.github` before this lands | ✅ Resolved | `GET repos/Mininglamp-OSS/.github/contents/.github/workflows/reusable-check-sprint.yml?ref=main` → 200, 17 KB, sha `202158405fca8ccc052570eed65f1e790eda622a`. The caller `uses:` line at `.github/workflows/check-sprint.yml:42` will resolve. |
| Stale check on sprint rollover / Project field edit (only refires on `opened`/`synchronize`/`reopened`) | ✅ Addressed in scope | `enforce-freshness: true` (`.github/workflows/check-sprint.yml:46`) makes the check fail when the head commit predates the current sprint start, which closes the rollover scenario whenever the check is re-evaluated. The remaining residual (Project field edited via UI on an up-to-date branch with no new push) is a genuine GitHub Platform limitation — Projects v2 has no webhook event Actions can subscribe to — and is now called out explicitly in the workflow header (`check-sprint.yml:7–25`). The recommended branch-protection mitigation ("Require branches to be up to date before merging") is documented at `check-sprint.yml:27–30`. Together with `enforce-freshness: true`, this is the strongest guarantee achievable in the current platform. |

## Verification of the change itself

| Item | Status | Evidence |
|---|---|---|
| YAML validates | ✅ | `sanity / actionlint` and `sanity / No tabs in workflow files` are passing on `bc6db9f`. |
| `pull_request_target` is correct for this use case | ✅ | Needs `PROJECT_TOKEN` access for fork PRs; no PR code is checked out or executed. |
| Workflow-level `permissions: {}` | ✅ | `check-sprint.yml:39` — `GITHUB_TOKEN` carries zero scopes; all API access flows through `PROJECT_TOKEN`. |
| Trigger scope limited to PRs targeting `main` | ✅ | `check-sprint.yml:35–37`. |
| Caller `uses:` ref matches repo convention | ✅ | `codeql.yml:25` already pins the same org reusable repo at `@main`. |
| Required-check name | ✅ | Caller job id `check-sprint` (`check-sprint.yml:42`) + reusable job name `check-sprint` ⇒ `check-sprint / check-sprint`, matching the deployer note. |
| Inputs/secret declarations match reusable contract | ✅ | Reusable `workflow_call` expects `pr-number` (number), `repo-name` (string), optional `enforce-freshness` (bool, default false), and secret `PROJECT_TOKEN` — all supplied with correct types. |
| `enforce-freshness: true` is wired through to the reusable workflow | ✅ | Input is declared at the reusable side (`reusable-check-sprint.yml:78–82`) and consumed in the freshness gate. |

## Non-blocking suggestions

### Nit — PR body should also mention the "Require branches to be up to date" branch-protection setting

The workflow file now clearly documents that `enforce-freshness: true` alone is not enough — it needs branch protection's "Require branches to be up to date before merging" to actually re-trigger after a sprint rollover. The PR body, however, still says only:

> Enable the required status check `check-sprint / check-sprint` on the `main` branch protection *after* this PR is merged.

It would be worth extending the deployer note to:

> After merging, on `main` branch protection enable both:
> 1. Required status check: `check-sprint / check-sprint`
> 2. **Require branches to be up to date before merging** (so a sprint rollover or PR aging re-triggers the check via the resulting push instead of letting a stale green status carry the merge).

Without (2), `enforce-freshness: true` only catches sprint rollover on PRs that happen to receive a new push between rollover and merge. With (2), it catches every case where the branch is rebased/updated before merge.

### Nit — Consider adding `ready_for_review` to the trigger list

A PR opened as a draft and later marked ready-for-review will not fire any of `opened` / `synchronize` / `reopened` on its own; the check will only run when a push later arrives. `codeql.yml:7` in this same repo already includes `ready_for_review` for the same reason. Adding it here would close that edge case and keep the file consistent with the rest of the repo's workflow triggers.

### Nit — Add the existing `zizmor` annotation pattern for `pull_request_target`

Other `pull_request_target` workflows in the repo (e.g., `auto-add-to-project.yml`) use the `# zizmor: ignore[dangerous-triggers]` annotation for metadata-only automations. The security model here is documented prose-style in the header, which is great for a human reader, but adding the annotation keeps workflow scanners consistent with the rest of the repo's pattern.

## Out-of-scope observations (informational; for the `Mininglamp-OSS/.github` PR's review, not this one)

Two items in the reusable workflow source that should be raised there if not already, but do not affect this caller PR:

1. The Projects v2 `iterations` field only returns currently-configured iterations. If the current calendar day falls in a gap between configured iterations (or before iteration #1), `current_sprint` becomes `None` and the workflow fails for every PR. Worth confirming the iteration schedule has no gaps or handling that case explicitly.
2. `get_today_shanghai()` (in the reusable workflow) could use `zoneinfo.ZoneInfo("Asia/Shanghai")` instead of shelling out to `/usr/bin/date`, removing the subprocess and the host-binary dependency. The reusable workflow already imports `zoneinfo`.

## Conclusion

The two blockers from the prior review round are resolved (reusable workflow exists on `.github@main`; stale-check limitation is documented honestly and mitigated with `enforce-freshness: true` plus a documented branch-protection requirement). The change is small, correct, security-sound, and consistent with existing repo conventions. The remaining items are deployment-note and consistency polish — none are merge blockers.

PR #76, opened 2026-05-26
chore(create-openclaw-octo): rename subdir and shed plugin source (1/3)

caster-Q:
## Summary

First of two PRs migrating this repo's CLI package from `openclaw-channel-octo` (npm-installed plugin + CLI) to `create-openclaw-octo` (CLI-only, ClawHub-bound).

This PR is **rename + repackage**, no behavioral changes to CLI commands. PR2 will add the legacy-install detection + force-upgrade pre-flight.

## Related Issue

N/A — internal migration tracked in the implementation plan doc.

## Changes

- Delete `openclaw-channel-dmwork/` (long-deprecated, removed from CI matrix)
- `git mv openclaw-channel-octo/ create-openclaw-octo/`
- Remove plugin runtime sources from the renamed subdir (`src/`, `skills/`, `index.ts`, `openclaw.plugin.json`) — plugin lives at `Mininglamp-OSS/openclaw-channel-octo` now; this subdir is CLI-only
- Inline the constants/version stubs into `cli/constants.ts` + `cli/version.ts`; rewrite imports in `cli/openclaw-cli.ts`, `cli/utils.ts`, `cli/index.ts` to use the new locations
- `package.json`: rename to `create-openclaw-octo`, drop `openclaw` plugin-extension block, drop `axios`/`ws`/`crypto-js`/etc runtime deps, drop `bundledDependencies`, refresh `files`/`bin`, update `prebuild` to write `cli/version.ts`
- `tsconfig.json`: drop stale `index.ts` + `src/**/*.ts` from `include`
- Update `.github/CODEOWNERS`, `.github/dependabot.yml`, `.gitignore`
- Rename CI workflows: `publish-octo.yml` (tag prefix `create-openclaw-octo/v*`), `publish-octo-dev.yml` (path filter), `ci.yml` (matrix entry, drop dmwork)
- Add top-level `permissions: {}` to publish workflows (per repo convention)
- Soften forward-looking comment about the redirect-shim subdir (subdir doesn't exist yet; lands in a future PR)
- Align user-facing CLI strings with the new package name
- Rewrite `create-openclaw-octo/README.md` for the CLI-only role (npm rename banner, migration table, issue-routing guide)

## Testing

- [x] Unit tests: `npm test` → **80/80 passing**
- [x] Type-check: `npx tsc --noEmit` → clean
- [x] Build: `npm run build` → succeeds (prebuild writes `cli/version.ts`)
[...]

PR #81, opened 2026-05-26
feat(create-openclaw-octo): force-upgrade legacy installs (2/2)

caster-Q:
## Summary

Second of two PRs in the CLI rename + force-upgrade migration (PR1 = #76, merged).

This PR adds the **force-upgrade pre-flight** that detects legacy installs (old npm 1.x of `openclaw-channel-octo`, the very-old `dmwork` plugin id, npm-residue after migration, etc.) and either auto-migrates (in `install`) or surfaces a clear blocking notice (in `bind` / `quickstart` / `remove-account`).

It also folds in the two P2 nits from #76's post-merge review (no code changes, infra/docs only).

## Related Issue

N/A — internal migration tracked in the implementation plan doc.

### Force-upgrade pre-flight (the substantive part)

- `cli/openclaw-cli.ts`:
  - `detectInstallState()` — 5-state classifier: `octo-clawhub` / `octo-npm-legacy` / `dmwork-legacy` / `broken` / `none`. Belt-and-suspenders npm fallback (cfg `entries` + `installs` + `hasLegacyNpmResidue()`) so we still detect 1.0.0-style installs on OpenClaw versions that don't support `plugins inspect`.
  - `compareOpenClawVersion()` — integer-tuple compare on dotted version strings (NOT semver — OpenClaw uses calendar-versioned `YYYY.M.D`).
  - `detectOpenClawState()` — 3-tier: `ok` / `warn` / `block`, gated on `OPENCLAW_PEER_MIN = "2026.4.15"` and `OPENCLAW_RECOMMENDED = "2026.5.18"`.
- `cli/utils.ts`:
  - `printUpgradeNotice(status, ...)` — open-right box format; `block` (✗) for hard stops, `warn` (⚠) for advisories.
  - `enforceHealthyClawHubInstall()` — shared pre-flight used by `bind` / `quickstart` / `remove-account`. Returns cleanly only when state is `octo-clawhub` and OpenClaw is `ok` / `warn`; otherwise prints upgrade notice and `process.exit(1)`.
  - `renderInstallStatusBanner()` — non-blocking variant for `doctor` (returns string, doesn't exit).
- `cli/bind.ts`, `cli/quickstart.ts`, `cli/remove-account.ts`:
  - Replace the old `isHealthyInstall(PLUGIN_ID) + ensureOpenClawCompat()` pair with a single `enforceHealthyClawHubInstall()` call. Single source of truth.
- `cli/index.ts`:
  - Remove the redundant `npm package: ...` line from `info` output (it was a constant — adds noise, no signal).
- Tests: 14 new cases for `compareOpenClawVersion` (8) + `detectOpenClawState` (6). Both placed BEFORE the `getConfigFilePathSafe` test block to avoid `vi.doMock` mock pollution. **94/94 passing locally.**

### Post-merge nit cleanup from #76

- `.github/workflows/ci.yml`: drop stale paths-filter patterns (`package.json`, `pnpm-lock.yaml`, `*.config.*`). No root `package.json` exists; repo uses npm not pnpm. Harmless but misleading.
- `create-openclaw-octo/README.md`: add explicit guidance for users who installed the old name globally (`npm i -g openclaw-channel-octo`) — the rename does not auto-clean their on-disk shim binary; recommend `npm uninstall -g openclaw-channel-octo` + `npx` workflow.

## Testing

- [x] Unit tests: `npm test` → **94/94 passing** (80 from PR1 + 14 new)
- [x] Type-check: `npx tsc --noEmit` → clean
- [x] Build: `npm run build` → succeeds
- [x] Manual scenario testing (per implementation plan, scenarios A/B/C/D):
  - **A** ClawHub octo healthy → `bind` proceeds normally
  - **B** ClawHub octo + npm residue → warn banner, proceed
  - **C** npm 1.0.0 only (`openclaw-channel-octo`) → block, prompt force-upgrade
  - **D** very-old `dmwork` plugin id → block, prompt force-upgrade

## Checklist

- [x] Added tests for my changes (14 new cases)
- [x] Updated documentation (README global-install warning)

## Commit list

```
3049931  feat: add Plan A detection helpers + upgrade notice
6252ae4  feat: wire plan A pre-flight into bind/quickstart/remove-account
462b125  test: cover compareOpenClawVersion + detectOpenClawState
5aa84f0  docs: drop redundant 'npm package' line from info output
14f3d83  chore: address PR #76 P2 review nits
```

## Series tracker

- [x] PR 1/2 (#76): rename subdir, shed plugin source, inline cli/constants + cli/version, rename npm package, infra
- [ ] **PR 2/2 (this PR)**: force-upgrade pre-flight + tests + post-merge nit cleanup

After this lands, the next external-facing step is publishing `create-openclaw-octo@1.0.1` to npm `@latest` (separate atomic-release plan; not part of this PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,9 +46,6 @@ jobs:
           filters: |
             code:
               - 'create-openclaw-octo/**'
-              - 'package.json'
-              - 'pnpm-lock.yaml'
-              - '*.config.*'
               - '.github/workflows/**'
 
   build:
--- a/create-openclaw-octo/README.md
+++ b/create-openclaw-octo/README.md
@@ -8,6 +8,8 @@ CLI tools for the [OpenClaw Octo channel plugin](https://github.com/Mininglamp-O
 
 > **Renamed from `openclaw-channel-octo`** (npm).
 > The old npm package name has been replaced by `create-openclaw-octo`. The legacy npm name will print a redirect notice; please use `npx -y create-openclaw-octo ...` going forward.
+>
+> If you previously installed the old name globally (`npm i -g openclaw-channel-octo`), the `openclaw-channel-octo` shim binary on your `$PATH` will not be updated by the rename. Remove it with `npm uninstall -g openclaw-channel-octo` and switch to the `npx -y create-openclaw-octo ...` workflow above. (No global install of the new name is needed; `npx` always fetches the latest.)
 
 Repository: https://github.com/Mininglamp-OSS/octo-adapters
 
--- a/create-openclaw-octo/cli/account-writes-octo.test.ts
+++ b/create-openclaw-octo/cli/account-writes-octo.test.ts
@@ -13,14 +13,14 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 // Mocks must be declared before importing the modules under test.
 // ---------------------------------------------------------------------------
 
-// Stub utils.js so ensureOpenClawCompat doesn't try to spawn `openclaw`.
+// Stub utils.js so enforceHealthyClawHubInstall doesn't try to spawn `openclaw`.
 // Re-export the real CHANNEL_ID/PLUGIN_ID/etc. via importOriginal so the
 // module under test sees the canonical helper behaviour for paths/objects.
 vi.mock("./utils.js", async (importOriginal) => {
   const actual = await importOriginal<typeof import("./utils.js")>();
   return {
     ...actual,
-    ensureOpenClawCompat: vi.fn(),
+    enforceHealthyClawHubInstall: vi.fn(),
   };
 });
 
--- a/create-openclaw-octo/cli/bind.ts
+++ b/create-openclaw-octo/cli/bind.ts
@@ -5,7 +5,6 @@
  */
 
 import {
-  isHealthyInstall,
   readConfigFromFile,
   writeConfigAtomic,
 } from "./openclaw-cli.js";
@@ -13,7 +12,7 @@ import {
   CHANNEL_ID,
   RECOMMENDED_DM_SCOPE,
   ensureChannelConfigObject,
-  ensureOpenClawCompat,
+  enforceHealthyClawHubInstall,
   validateAccountId,
 } from "./utils.js";
 
@@ -25,14 +24,8 @@ export interface BindOptions {
 }
 
 export async function runBind(opts: BindOptions): Promise<void> {
-  ensureOpenClawCompat();
-
-  // 1. Pre-flight: plugin must be healthy
-  if (!isHealthyInstall()) {
-    console.error("Octo plugin is not installed or in an unhealthy state.");
-    console.error("Please run first: npx -y create-openclaw-octo install");
-    process.exit(1);
-  }
+  // 1. Pre-flight: enforce healthy ClawHub octo + recent OpenClaw (exits on legacy/missing/broken)
+  enforceHealthyClawHubInstall();
 
   // 2. Validate params
   if (!opts.botToken.startsWith("bf_")) {
--- a/create-openclaw-octo/cli/index.ts
+++ b/create-openclaw-octo/cli/index.ts
@@ -14,7 +14,7 @@ import {
 } from "./doctor.js";
 import { runUninstall } from "./uninstall.js";
 import { runRemoveAccount } from "./remove-account.js";
-import { ensureOpenClawCompat, NPM_PACKAGE_NAME, PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
+import { ensureOpenClawCompat, PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
 import { getOpenClawVersionStrict, resolvePluginState } from "./openclaw-cli.js";
 import { PLUGIN_VERSION } from "./version.js";
 
@@ -57,7 +57,6 @@ program
     console.log(`${b}openclaw:${r} ${g}${openclawVersion}${r}`);
     console.log(`${b}installed plugin id:${r} ${g}${PLUGIN_ID}${r} (ClawHub)`);
     console.log(`${b}installed plugin version:${r} ${g}${installedVersion}${r}`);
-    console.log(`${b}npm package:${r} ${g}${NPM_PACKAGE_NAME}${r}`);
     console.log();
     console.log(`${b}Environment:${r}`);
     console.log(`${b}OS:${r} ${g}${process.platform} ${process.arch}${r}`);
--- a/create-openclaw-octo/cli/index.ts
+++ b/create-openclaw-octo/cli/index.ts
@@ -14,7 +14,7 @@ import {
 } from "./doctor.js";
 import { runUninstall } from "./uninstall.js";
 import { runRemoveAccount } from "./remove-account.js";
-import { ensureOpenClawCompat, NPM_PACKAGE_NAME, PLUGIN_ID } from "./utils.js";
+import { ensureOpenClawCompat, NPM_PACKAGE_NAME, PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
 import { getOpenClawVersionStrict, resolvePluginState } from "./openclaw-cli.js";
 import { PLUGIN_VERSION } from "./version.js";
 
@@ -141,7 +141,6 @@ program
   .option("--fix", "Attempt to automatically fix issues", false)
   .option("--json", "Output JSON", false)
   .action(async (opts) => {
-    ensureOpenClawCompat();
     const result = await runDoctorChecks({
       reader: cliConfigReader,
       accountId: opts.accountId,
@@ -150,6 +149,13 @@ program
     if (opts.json) {
       console.log(JSON.stringify(result, null, 2));
     } else {
+      // Render upgrade-status banner first (only when there's actually an
+      // issue to surface — null on the healthy path).
+      const banner = renderInstallStatusBanner();
+      if (banner) {
+        console.error(banner);
+        console.error("");
+      }
       console.log(formatDoctorResult(result));
     }
   });
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -554,6 +554,132 @@ describe("resolvePluginState", () => {
   });
 });
 
+// ---------------------------------------------------------------------------
+// compareOpenClawVersion — pure function, no mocks needed
+//
+// IMPORTANT: This block (and detectOpenClawState below) MUST stay before the
+// `getConfigFilePathSafe` block. That block uses `vi.doMock("node:child_process",
+// ...)` with a LOCAL mock fn, which permanently re-binds the module mock so the
+// global `mockExecFileSync` reference (set at the top of this file) no longer
+// reaches the openclaw-cli.js module instance after that. Tests below
+// `getConfigFilePathSafe` can't reliably control execFileSync via that ref.
+// ---------------------------------------------------------------------------
+
+describe("compareOpenClawVersion", () => {
+  it("returns 0 for equal versions", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.4.15", "2026.4.15")).toBe(0);
+  });
+
+  it("returns -1 when a < b in patch", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.4.14", "2026.4.15")).toBe(-1);
+  });
+
+  it("returns 1 when a > b in patch", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.4.16", "2026.4.15")).toBe(1);
+  });
+
+  it("returns 1 when a > b in minor (major equal)", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.5.0", "2026.4.99")).toBe(1);
+  });
+
+  it("returns 1 when a > b in major", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2027.0.0", "2026.99.99")).toBe(1);
+  });
+
+  it("regression: double-digit minor is integer-compared, not lex-compared", async () => {
+    // Lexical compare would say "10" < "9". Integer compare says "10" > "9".
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.10.0", "2026.9.0")).toBe(1);
+    expect(compareOpenClawVersion("2026.9.0", "2026.10.0")).toBe(-1);
+  });
+
+  it("regression: double-digit patch is integer-compared, not lex-compared", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.5.18", "2026.5.2")).toBe(1);
+    expect(compareOpenClawVersion("2026.5.2", "2026.5.18")).toBe(-1);
+  });
+
+  it("treats missing segments as 0", async () => {
+    const { compareOpenClawVersion } = await loadModule();
+    expect(compareOpenClawVersion("2026.5", "2026.5.0")).toBe(0);
+    expect(compareOpenClawVersion("2026", "2026.0.0")).toBe(0);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// detectOpenClawState — three tiers: block / warn / ok
+// ---------------------------------------------------------------------------
+
+describe("detectOpenClawState", () => {
+  it("block when openclaw is not on PATH (ENOENT)", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => {
+      const err = new Error("spawn openclaw ENOENT") as any;
+      err.code = "ENOENT";
+      throw err;
+    });
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("block");
+    if (state.kind === "block") {
+      expect(state.version).toBeNull();
+      expect(state.reason).toMatch(/not installed|not on PATH/i);
+    }
+  });
+
+  it("block when version < OPENCLAW_PEER_MIN (2026.4.15)", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.3.10 (abc1234)\n" as any);
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("block");
+    if (state.kind === "block") {
+      expect(state.version).toBe("2026.3.10");
+      expect(state.reason).toContain("2026.3.10");
+      expect(state.reason).toContain("2026.4.15");
+    }
+  });
+
+  it("warn when peer-min <= version < recommended (2026.5.18)", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.4.20 (abc1234)\n" as any);
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("warn");
+    if (state.kind === "warn") {
+      expect(state.version).toBe("2026.4.20");
+      expect(state.reason).toContain("2026.4.20");
+      expect(state.reason).toContain("2026.5.18");
+    }
+  });
+
+  it("warn at exact peer-min (boundary)", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.4.15 (abc1234)\n" as any);
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("warn");
+  });
+
+  it("ok at exact recommended (boundary)", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.5.18 (abc1234)\n" as any);
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("ok");
+    if (state.kind === "ok") {
+      expect(state.version).toBe("2026.5.18");
+    }
+  });
+
+  it("ok when version > recommended", async () => {
+    const { detectOpenClawState } = await loadModule();
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2027.1.0 (abc1234)\n" as any);
+    const state = detectOpenClawState();
+    expect(state.kind).toBe("ok");
+  });
+});
+
 describe("getConfigFilePathSafe (Windows relative path)", () => {
   it("should resolve Windows relative path .\\.\\.openclaw\\openclaw.json to homedir", async () => {
     vi.resetModules();
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -693,6 +693,8 @@ describe("detectInstallState", () => {
     legacyNpmInEntries?: boolean;
     legacyNpmInInstalls?: boolean;
     npmDirResidue?: boolean;
+    dmworkChannel?: boolean;
+    dmworkBinding?: boolean;
   }) {
     return async () => {
       const { existsSync, readFileSync } = await import("node:fs");
@@ -722,6 +724,12 @@ describe("detectInstallState", () => {
           installPath: "~/.openclaw/npm/node_modules/openclaw-channel-octo",
         };
       }
+      if (opts.dmworkChannel) {
+        cfg.channels = { dmwork: { accounts: { "u1": {} } } };
+      }
+      if (opts.dmworkBinding) {
+        cfg.bindings = [{ match: { channel: "dmwork" }, accountId: "u1", agent: "main" }];
+      }
       vi.mocked(readFileSync).mockReturnValue(JSON.stringify(cfg));
       vi.mocked(existsSync).mockImplementation((p: unknown) => {
         const path = String(p);
@@ -741,6 +749,7 @@ describe("detectInstallState", () => {
     expect(state.kind).toBe("octo-clawhub");
     if (state.kind === "octo-clawhub") {
       expect(state.legacyNpmActive).toBe(false);
+      expect(state.legacyDmworkResidue).toBe(false);
       expect(state.npmResidue).toBe(false);
       expect(state.version).toBe("1.0.12");
     }
@@ -779,6 +788,27 @@ describe("detectInstallState", () => {
     if (state.kind === "octo-clawhub") {
       expect(state.legacyNpmActive).toBe(true);
       expect(state.npmResidue).toBe(true);
+    }
+  });
+
+  it("octo-clawhub with dmwork channel residue — legacyDmworkResidue=true (BLOCK)", async () => {
+    await setupOctoClawHub({ dmworkChannel: true })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyDmworkResidue).toBe(true);
+      expect(state.legacyNpmActive).toBe(false);
+    }
+  });
+
+  it("octo-clawhub with dmwork binding residue — legacyDmworkResidue=true (BLOCK)", async () => {
+    await setupOctoClawHub({ dmworkBinding: true })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyDmworkResidue).toBe(true);
     }
   });
 
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -680,6 +680,144 @@ describe("detectOpenClawState", () => {
   });
 });
 
+// ---------------------------------------------------------------------------
+// detectInstallState — also kept BEFORE the `getConfigFilePathSafe` block for
+// the same `vi.doMock` re-binding reason called out at line 560.
+// ---------------------------------------------------------------------------
+
+describe("detectInstallState", () => {
+  // Helper: ClawHub octo present in cfg + extensions dir, with optional
+  // `legacyNpm` toggle to add cfg.plugins.entries["openclaw-channel-octo"]
+  // (the dual-active signal).
+  function setupOctoClawHub(opts: {
+    legacyNpmInEntries?: boolean;
+    legacyNpmInInstalls?: boolean;
+    npmDirResidue?: boolean;
+  }) {
+    return async () => {
+      const { existsSync, readFileSync } = await import("node:fs");
+      mockExecFileSync.mockImplementation((cmd, args) => {
+        const argsArr = args as string[];
+        if (argsArr[0] === "config" && argsArr[1] === "file") {
+          return "/home/user/.openclaw/openclaw.json";
+        }
+        if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
+          // simulate old OpenClaw → fallback path inside isHealthyInstall
+          throw new Error("error: unknown command 'inspect'");
+        }
+        return "";
+      });
+      const cfg: any = {
+        plugins: {
+          entries: { octo: { enabled: true } },
+          installs: { octo: { version: "1.0.12", installPath: "~/.openclaw/extensions/octo" } },
+        },
+      };
+      if (opts.legacyNpmInEntries) {
+        cfg.plugins.entries["openclaw-channel-octo"] = { enabled: true };
+      }
+      if (opts.legacyNpmInInstalls) {
+        cfg.plugins.installs["openclaw-channel-octo"] = {
+          version: "1.0.0",
+          installPath: "~/.openclaw/npm/node_modules/openclaw-channel-octo",
+        };
+      }
+      vi.mocked(readFileSync).mockReturnValue(JSON.stringify(cfg));
+      vi.mocked(existsSync).mockImplementation((p: unknown) => {
+        const path = String(p);
+        if (path.endsWith("/extensions/octo")) return true;
+        if (path.endsWith(".openclaw/npm/node_modules/openclaw-channel-octo")) {
+          return Boolean(opts.npmDirResidue);
+        }
+        return false;
+      });
+    };
+  }
+
+  it("octo-clawhub healthy and clean (no residue, no legacy active)", async () => {
+    await setupOctoClawHub({})();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyNpmActive).toBe(false);
+      expect(state.npmResidue).toBe(false);
+      expect(state.version).toBe("1.0.12");
+    }
+  });
+
+  it("octo-clawhub with directory residue only — npmResidue=true, legacyNpmActive=false", async () => {
+    await setupOctoClawHub({ npmDirResidue: true })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.npmResidue).toBe(true);
+      expect(state.legacyNpmActive).toBe(false);
+    }
+  });
+
+  it("octo-clawhub with legacy npm still registered in cfg.entries — legacyNpmActive=true (BLOCK)", async () => {
+    await setupOctoClawHub({ legacyNpmInEntries: true, legacyNpmInInstalls: true })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyNpmActive).toBe(true);
+    }
+  });
+
+  it("octo-clawhub: legacyNpmActive takes precedence regardless of dir residue", async () => {
+    await setupOctoClawHub({
+      legacyNpmInEntries: true,
+      legacyNpmInInstalls: true,
+      npmDirResidue: true,
+    })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyNpmActive).toBe(true);
+      expect(state.npmResidue).toBe(true);
+    }
+  });
+
+  it("octo-npm-legacy when ClawHub octo absent and NPM_PACKAGE_NAME healthy", async () => {
+    const { existsSync, readFileSync } = await import("node:fs");
+    mockExecFileSync.mockImplementation((cmd, args) => {
+      const argsArr = args as string[];
+      if (argsArr[0] === "config" && argsArr[1] === "file") {
+        return "/home/user/.openclaw/openclaw.json";
+      }
+      if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
+        throw new Error("error: unknown command 'inspect'");
+      }
+      return "";
+    });
+    // No `octo` entry; only legacy npm is present
+    vi.mocked(readFileSync).mockReturnValue(JSON.stringify({
+      plugins: {
+        entries: { "openclaw-channel-octo": { enabled: true } },
+        installs: {
+          "openclaw-channel-octo": {
+            version: "1.0.0",
+            installPath: "~/.openclaw/extensions/openclaw-channel-octo",
+          },
+        },
+      },
+    }));
+    vi.mocked(existsSync).mockImplementation((p: unknown) =>
+      String(p).endsWith("/extensions/openclaw-channel-octo"),
+    );
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-npm-legacy");
+    if (state.kind === "octo-npm-legacy") {
+      expect(state.version).toBe("1.0.0");
+    }
+  });
+});
+
 describe("getConfigFilePathSafe (Windows relative path)", () => {
   it("should resolve Windows relative path .\\.\\.openclaw\\openclaw.json to homedir", async () => {
     vi.resetModules();
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -695,6 +695,8 @@ describe("detectInstallState", () => {
     npmDirResidue?: boolean;
     dmworkChannel?: boolean;
     dmworkBinding?: boolean;
+    veryLegacyDmworkInEntries?: boolean;
+    veryLegacyDmworkInInstalls?: boolean;
   }) {
     return async () => {
       const { existsSync, readFileSync } = await import("node:fs");
@@ -729,6 +731,15 @@ describe("detectInstallState", () => {
       }
       if (opts.dmworkBinding) {
         cfg.bindings = [{ match: { channel: "dmwork" }, accountId: "u1", agent: "main" }];
+      }
+      if (opts.veryLegacyDmworkInEntries) {
+        cfg.plugins.entries["dmwork"] = { enabled: true };
+      }
+      if (opts.veryLegacyDmworkInInstalls) {
+        cfg.plugins.installs["dmwork"] = {
+          version: "0.5.21",
+          installPath: "~/.openclaw/extensions/dmwork",
+        };
       }
       vi.mocked(readFileSync).mockReturnValue(JSON.stringify(cfg));
       vi.mocked(existsSync).mockImplementation((p: unknown) => {
@@ -809,6 +820,26 @@ describe("detectInstallState", () => {
     expect(state.kind).toBe("octo-clawhub");
     if (state.kind === "octo-clawhub") {
       expect(state.legacyDmworkResidue).toBe(true);
+    }
+  });
+
+  it("octo-clawhub with very-legacy `dmwork` plugin entries — legacyDmworkResidue=true (BLOCK)", async () => {
+    // Regression for the `hasVeryLegacyPluginArtifacts` precedence —
+    // detectScenario() in install.ts treats `dmwork` (very-legacy id) as
+    // priority 1 (legacy-to-octo migration) above `openclaw-channel-dmwork`
+    // (intermediate id) at priority 2 (rebrand). The pre-flight must
+    // surface BOTH ids so a stale `dmwork` entry without channels.dmwork
+    // or bindings still blocks bind/quickstart/remove-account.
+    await setupOctoClawHub({
+      veryLegacyDmworkInEntries: true,
+      veryLegacyDmworkInInstalls: true,
+    })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyDmworkResidue).toBe(true);
+      expect(state.legacyNpmActive).toBe(false);
     }
   });
 
--- a/create-openclaw-octo/cli/openclaw-cli.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.ts
@@ -1017,7 +1017,12 @@ export function isHealthyInstall(pluginId: string = PLUGIN_ID): boolean {
  * state we surface as a warn but don't block on).
  */
 export type InstallState =
-  | { kind: "octo-clawhub"; version: string | null; npmResidue: boolean }
+  | {
+      kind: "octo-clawhub";
+      version: string | null;
+      npmResidue: boolean;
+      legacyNpmActive: boolean;
+    }
   | { kind: "octo-npm-legacy"; version: string | null }
   | { kind: "dmwork-legacy"; version: string | null }
   | { kind: "broken"; details: string }
@@ -1029,6 +1034,27 @@ function hasLegacyNpmResidue(): boolean {
   return existsSync(resolve(homedir(), LEGACY_NPM_RESIDUE_PATH));
 }
 
+/**
+ * Whether the legacy npm-installed `openclaw-channel-octo` is still registered
+ * in OpenClaw's plugin entries (i.e., OpenClaw still attempts to load it on
+ * startup). Both legacy npm v1.x and current ClawHub octo register channel
+ * id "octo", so a healthy ClawHub install plus a still-registered legacy npm
+ * install double-register the channel — install.ts treats this as a hard
+ * cleanup error after install. detectInstallState surfaces it before
+ * bind/quickstart/remove-account write any config so the user is prompted
+ * to migrate first instead of silently writing into a duplicated channel.
+ *
+ * Signal source: cfg.plugins.entries[NPM_PACKAGE_NAME]. Note that
+ * cfg.plugins.installs has been observed null on current OpenClaw schemas
+ * (install metadata moved into `plugins inspect <id> --json`), so installs
+ * cannot be used as a registration signal. Tracked in #82 for the broader
+ * detection-layer cleanup.
+ */
+function isLegacyNpmStillRegistered(): boolean {
+  const cfg = readConfigFromFile();
+  return Boolean(cfg?.plugins?.entries?.[NPM_PACKAGE_NAME]);
+}
+
 export function detectInstallState(): InstallState {
   // Priority order: ClawHub (target) → npm-legacy → dmwork-legacy → broken → none
   if (isHealthyInstall(PLUGIN_ID)) {
@@ -1037,6 +1063,7 @@ export function detectInstallState(): InstallState {
       kind: "octo-clawhub",
       version: state.version,
       npmResidue: hasLegacyNpmResidue(),
+      legacyNpmActive: isLegacyNpmStillRegistered(),
     };
   }
   if (isHealthyInstall(NPM_PACKAGE_NAME)) {
--- a/create-openclaw-octo/cli/openclaw-cli.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.ts
@@ -1065,13 +1065,16 @@ export function detectInstallState(): InstallState {
       version: state.version,
       npmResidue: hasLegacyNpmResidue(),
       legacyNpmActive: isLegacyNpmStillRegistered(cfg),
-      // Mirror install.ts detectScenario priority: dmwork artifacts (plugin
-      // dir/entries/installs/allow + channels.dmwork + bindings(channel=dmwork))
-      // require a `rebrand` migration even when octo is otherwise healthy.
-      // Block the pre-flight in this state so the user is prompted to run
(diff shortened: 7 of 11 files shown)


lml2468 (approved):
[APPROVE]

**Scope:** PR 2/2 — force-upgrade pre-flight for legacy installs + post-merge nit cleanup from #76.

**Code quality: solid.** Well-typed discriminated unions, exhaustive switch coverage, clean separation of concerns.

**Key additions reviewed:**

1. **`detectInstallState()`** (5-state classifier) — correct priority ordering: ClawHub → npm-legacy → dmwork-legacy → broken → none. Belt-and-suspenders fallback for old OpenClaw runtimes without `plugins inspect` is a good defensive measure.

2. **`compareOpenClawVersion()`** — integer-segment comparison, correctly avoids SemVer (calendar-versioned). Tests cover double-digit segment regression and missing-segment edge cases.

3. **`detectOpenClawState()`** — clean 3-tier (block/warn/ok) with well-documented version bounds (`OPENCLAW_PEER_MIN`, `OPENCLAW_RECOMMENDED`).

4. **`enforceHealthyClawHubInstall()`** — single-source-of-truth pre-flight replacing the old `isHealthyInstall() + ensureOpenClawCompat()` pair. Exhaustive `switch` on `plugin.kind` — no fall-through risk.

5. **`renderInstallStatusBanner()`** — non-blocking variant for `doctor`. Returns `null` on happy path (no noise).

6. **Tests** — 14 new cases with good boundary coverage. vi.doMock ordering comment is appreciated.

7. **Wiring** — `bind`, `quickstart`, `remove-account` correctly replaced with single `enforceHealthyClawHubInstall()` call. `doctor` uses the non-blocking `renderInstallStatusBanner()` variant.

**CI:** Build & Type-check ✅, CodeQL pending (expected to pass), `check-sprint` ❌ (Sprint W22 not set).

No blocking issues found. LGTM.

lml2468 (dismissed):
REVIEW_STATE: APPROVED

## Summary

Clean, well-designed force-upgrade pre-flight implementation. All CI green.

### Architecture ✅

- **5-state install classifier** (`detectInstallState`): ClawHub → npm-legacy → dmwork-legacy → broken → none. Belt-and-suspenders detection with npm residue path fallback for old runtimes that lack `plugins inspect`.
- **3-tier version gate** (`detectOpenClawState`): block / warn / ok with clear thresholds (`OPENCLAW_PEER_MIN = "2026.4.15"`, `OPENCLAW_RECOMMENDED = "2026.5.18"`).
- **Single enforcement point** (`enforceHealthyClawHubInstall`): replaces the scattered `ensureOpenClawCompat()` + `isHealthyInstall()` pair. Used by `bind`, `quickstart`, `remove-account`.
- **Non-blocking variant** (`renderInstallStatusBanner`): for `doctor` — surfaces issues without exiting.

### Code quality ✅

- `compareOpenClawVersion` uses integer-segment ordering (not semver) — correct for OpenClaw's calendar versioning.
- Proper boundary handling (peer-min is warn, recommended is ok).
- Type-driven state machine (`InstallState`, `OpenClawState`) with discriminated unions.
- Clean stderr UX: boxed notices with actionable commands.

### Tests ✅

- 14 new tests: 8 for `compareOpenClawVersion` (including regression for double-digit lex-vs-integer), 6 for `detectOpenClawState` (ENOENT, boundary conditions, all tiers).
- Placement before `vi.doMock` block is explicitly documented to avoid mock pollution.
- 94/94 passing per PR body.

### Minor nits (non-blocking, no action needed)

- The `switch (plugin.kind)` in `enforceHealthyClawHubInstall` doesn't have a default case — TypeScript exhaustiveness checking via return type handles this, but an explicit `default: never` assertion would make it future-proof.

No blocking issues. Ship it.

Jerry-Xin (changes requested):
Summary: The PR is in scope for this repository, but the new pre-flight can incorrectly allow a dangerous dual-install legacy state.

**🔴 Blocking**
- 🔴 Critical: `detectInstallState()` classifies `octo` as healthy before checking whether legacy installs are still live. In [create-openclaw-octo/cli/openclaw-cli.ts:1034](/Users/mlamp/.openclaw/tmp/pr-review-kha1hwgv/repo/create-openclaw-octo/cli/openclaw-cli.ts:1034), the function returns `octo-clawhub` immediately when `PLUGIN_ID` is healthy, so a system with both healthy ClawHub `octo` and healthy legacy `openclaw-channel-octo` is treated as allowed. `enforceHealthyClawHubInstall()` then only warns on `npmResidue` and proceeds in [create-openclaw-octo/cli/utils.ts:244](/Users/mlamp/.openclaw/tmp/pr-review-kha1hwgv/repo/create-openclaw-octo/cli/utils.ts:244). That misses the exact risky state the installer comments describe as duplicate `"octo"` channel registration requiring cleanup in [create-openclaw-octo/cli/install.ts:263](/Users/mlamp/.openclaw/tmp/pr-review-kha1hwgv/repo/create-openclaw-octo/cli/install.ts:263). Please distinguish “filesystem residue only” from a still-registered healthy legacy npm install, and block/force-upgrade when legacy entries/installs are still live alongside ClawHub.

**💬 Non-blocking**
- 🟡 Warning: There are no direct tests for `detectInstallState()` or `enforceHealthyClawHubInstall()` dual-install classification. Add coverage for ClawHub-only, ClawHub + npm directory residue only, ClawHub + registered legacy npm, npm-only legacy, and dmwork-only legacy.

**✅ Highlights**
- The PR is relevant to `create-openclaw-octo` migration behavior and belongs in this repo.
- `compareOpenClawVersion()` and `detectOpenClawState()` coverage is useful, especially the integer segment comparisons.
- Verified locally after `npm ci`: `npm test` passed 94/94, `npm run type-check` passed, and `npm run build` passed.

yujiawei (dismissed):
# Code Review — PR #81 (octo-adapters)

## TL;DR

Solid second half of the rename + force-upgrade migration. The new pre-flight (`enforceHealthyClawHubInstall`) consolidates three previously-scattered checks into a single source of truth, and the boxed upgrade notice is a clear UX upgrade over the old plain-text errors. The core classifier (`detectInstallState`) handles the tricky belt-and-suspenders case for npm 1.x installs that don't live under `extensions/`. Type-check, build, and 94/94 tests pass locally.

No blocking correctness, security, or data-loss issues found. A few P2 nits below — none gate merge.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Build (`npm run build`) | ✅ | clean tsc emit |
| Type-check (`npx tsc --noEmit`) | ✅ | clean |
| Tests (`npm test`) | ✅ | 94/94 passing (matches PR claim) |
| Diff scope matches description | ✅ | `git diff main...pr81 --stat`: 11 files / +501 / -38; 14 new test cases (8 + 6) |
| Pre-flight integration | ✅ | `bind.ts`, `quickstart.ts`, `remove-account.ts` each replace the old `ensureOpenClawCompat()` + `isHealthyInstall()` pair with a single `enforceHealthyClawHubInstall()` call |
| `info` cleanup (`npm package`) | ✅ | `cli/index.ts` line removed; `NPM_PACKAGE_NAME` no longer imported there |
| Legacy npm residue path | ✅ | hard-coded as `~/.openclaw/npm/node_modules/openclaw-channel-octo`, matches v1 install layout |
| Test placement constraint | ✅ | new blocks precede the `getConfigFilePathSafe` block to avoid `vi.doMock` pollution; comment documents the constraint |
| README global-install warning | ✅ | covers `npm uninstall -g openclaw-channel-octo` migration |
| `ci.yml` paths-filter cleanup | ✅ | repo has no root `package.json`, uses npm not pnpm — patterns were noise |

## 2. Security-sensitive review

The dispatcher tagged this PR `security_sensitive` based on the migration/legacy keyword filter. The actual diff does NOT touch auth, token storage, credential redaction, or the binding-write code path — only the pre-flight gate that decides whether to enter those code paths.

- No new use of `eval`, shell-string concatenation, or untrusted-path construction. `runOpenclaw` / `runCmd` continue to use `execFileSync` with argv arrays.
- `printUpgradeNotice()` writes to stderr only. Strings are built from constants and a small set of typed `InstallState` fields; no caller-supplied string interpolation reaches the box.
- `enforceHealthyClawHubInstall` calls `process.exit(1)` directly on every failure branch — no exception leakage that could let a downstream code path proceed with stale assumptions.
- Failure modes (false-allow / false-block) cause UX failures, not privilege escalation or token leak.

No additional security concerns to flag for human review beyond the points already labelled `needs-human-review`.

## 3. Findings

### P2 — Coverage / maintainability nits (non-blocking)

**P2-1. The two most consequential new functions have no direct unit tests.**

`detectInstallState` (`cli/openclaw-cli.ts:1032-1100`) is the new 5-state classifier — `octo-clawhub` / `octo-npm-legacy` / `dmwork-legacy` / `broken` / `none` — including the belt-and-suspenders npm 1.x branch (`cli/openclaw-cli.ts:1056-1066`) that exists specifically because `isHealthyInstall(NPM_PACKAGE_NAME)` returns false on legacy npm installs that don't live under `extensions/`. Yet there are zero direct unit tests for this function. The same is true of `enforceHealthyClawHubInstall` (`cli/utils.ts:222-291`) — its 5 switch arms are exercised only via `bind`/`quickstart`/`remove-account` tests, all of which mock the function out (`quickstart.test.ts:18,49`, `account-writes-octo.test.ts:23`).

The 14 new tests cover the pure helpers (`compareOpenClawVersion`, `detectOpenClawState`) which are valuable but were also the easiest to test. Recommend follow-up tests for `detectInstallState`'s branches (especially the npm 1.x belt-and-suspenders fallback at `openclaw-cli.ts:1056-1066` and the leftover-config "broken" classifier at `:1088-1098`).

**P2-2. Duplicated version-compare + floor.**

`compareVersions` / `MIN_OPENCLAW_VERSION = "2026.4.15"` (`cli/utils.ts:37,~58`) and the new `compareOpenClawVersion` / `OPENCLAW_PEER_MIN = "2026.4.15"` (`cli/openclaw-cli.ts:1118,1137-1148`) coexist with identical lower-bound and effectively identical integer-segment semantics. `ensureOpenClawCompat` (still used by `install.ts`, `uninstall.ts`) uses the old pair; everything else uses the new pair. Drift risk if the floor moves and only one side is updated. Worth deleting one or aliasing them in a follow-up.

**P2-3. `hasLegacyNpmResidue()` hard-codes the homedir path.**

`cli/openclaw-cli.ts:1028-1030` checks `homedir() + .openclaw/npm/...` instead of deriving from `getConfigFilePathSafe()`. Consistent with the legacy v1 npm install layout, but a user with a non-default OpenClaw config root (e.g. via env override) wouldn't have residue at this path. Minor inconsistency vs. how the rest of the file resolves config-relative paths.

**P2-4. Switch on `plugin.kind` has no `default` / exhaustiveness arm.**

`cli/utils.ts:251-291` — if a future `InstallState` variant is added, the switch silently returns without `process.exit`, and the caller proceeds as if healthy. A `default: { const _exhaustive: never = plugin; throw new Error(...); }` (or a `case` for the new kind) would make this fail closed.

**P2-5. Test-placement comment indicates fragile architecture.**

`cli/openclaw-cli.test.ts:557-565` documents that the new test blocks "MUST stay before the `getConfigFilePathSafe` block" because that block re-binds the `node:child_process` mock with a local fn, breaking the global `mockExecFileSync` reference for everything after it. Worth a follow-up to either (a) move the polluting block to its own file or (b) reset the module mock at the boundary so order doesn't matter. Not blocking — the comment will catch most accidents.

**P2-6. Doctor lost its strict openclaw-missing exit.**

`cli/index.ts` removes `ensureOpenClawCompat()` from the `doctor` command and adds the non-blocking `renderInstallStatusBanner()` instead. Net effect: when `openclaw` is not on PATH, doctor no longer exits early — it prints `✗ OpenClaw is not installed or not on PATH` then runs all checks (which will fail in their own ways via internal try/catch). This appears intentional (doctor is diagnostic; failing checks individually is more informative than a hard crash), but worth flagging in case the previous behaviour was load-bearing for any downstream tooling.

### Out of scope (not raised as findings, just noted)

- The `LEGACY_NPM_RESIDUE_PATH` cleanup is suggested in the warn message (`cli/utils.ts:267-268`: "Clean up by re-running install"), but I didn't verify that `install` actually removes the npm residue directory. If it doesn't, the warning loops every run. Verifying that path is a follow-up task and outside this PR's diff.
- Doctor doesn't surface the new `OPENCLAW_RECOMMENDED` floor in any check beyond the banner — fine for now, but if doctor is meant to be the canonical diagnostic, eventually the version warn should be a check too.

## 4. Recommendations

- Approve and merge as-is.
- Open a follow-up issue for: P2-1 (`detectInstallState` direct tests), P2-2 (collapse the two version-compare helpers), and P2-4 (exhaustive switch).

caster-Q:
Addressed @Jerry-Xin's blocking finding — minimal fix in PR scope. Pushed `8d28b98`.

### Change

- `detectInstallState()` `octo-clawhub` variant gains `legacyNpmActive: boolean`, sourced from `cfg.plugins.entries[NPM_PACKAGE_NAME]`.
- `enforceHealthyClawHubInstall()` now prints a `block` upgrade notice and exits when `legacyNpmActive === true`. Suggested remedy: `npx -y create-openclaw-octo install --force`.
- `renderInstallStatusBanner()` shows `✗ Legacy npm openclaw-channel-octo still registered alongside ClawHub octo` for the dual-active state, taking precedence over the existing `⚠ residue` line.
- Added 5 tests (also covers the non-blocking warning Jerry-Xin raised about missing direct coverage):
  - clean `octo-clawhub`
  - directory residue only (`npmResidue: true`, `legacyNpmActive: false`)
  - registered legacy npm (`legacyNpmActive: true`)
  - registered legacy + dir residue (block precedence)
  - `octo-npm-legacy` fallback when ClawHub absent

99/99 tests passing locally; type-check clean.

### Why `cfg.plugins.entries` and not `cfg.plugins.installs`

While preparing the fix I verified the runtime config schema:

```bash
$ jq '.plugins | keys' ~/.openclaw/openclaw.json
["allow", "bundledDiscovery", "entries"]

$ jq '.plugins.installs' ~/.openclaw/openclaw.json
null
```

Install metadata moved out of `cfg.plugins.installs` into `openclaw plugins inspect <id> --json` (under `install`, singular). On current OpenClaw schemas `cfg.plugins.installs` is `null`, so an `installs[id]` lookup is always falsy — making it unusable as a registration signal. `cfg.plugins.entries` remains and is the load-time registry.

### Follow-up tech debt

The same investigation found that `isHealthyInstall()`'s fallback `(hasDir && hasEntries && hasInstalls)` is effectively dead code on current OpenClaw versions (the `hasInstalls` term is always false). Plus `LEGACY_PLUGIN_ID` is npm-installed at `npm/node_modules/<id>` not `extensions/<id>`, so its directory probe is wrong. These are out of scope for the BLOCKER fix but are now tracked in #82, which proposes consolidating the whole detection layer behind `openclaw plugins list --json`.

### Re-review

@Jerry-Xin please re-review.

lml2468 (dismissed):
REVIEW_STATE: APPROVED

## Re-review (head: 8d28b98e75b1, previous: APPROVED @ 14f3d83f)

### Jerry-Xin's P0 resolved ✅

The dual-install detection gap is fixed:

1. **New `isLegacyNpmStillRegistered()`** — checks `cfg.plugins.entries[NPM_PACKAGE_NAME]` (the actual registration signal, not just filesystem residue).
2. **`InstallState` type extended** — `octo-clawhub` now carries `legacyNpmActive: boolean` alongside existing `npmResidue`.
3. **`enforceHealthyClawHubInstall()` blocks** on `legacyNpmActive === true` with a clear upgrade notice pointing to `install --force`.
4. **`renderInstallStatusBanner()`** distinguishes dual-active (✗ hard error) from residue-only (⚠ warning).

### New tests (+138 lines, 5 cases) ✅

- Clean ClawHub (no residue, no legacy active)
- ClawHub + directory residue only (`npmResidue=true`, `legacyNpmActive=false`)
- ClawHub + legacy npm still registered (`legacyNpmActive=true`, BLOCK case)
- `legacyNpmActive` takes precedence regardless of dir residue
- `octo-npm-legacy` when ClawHub absent

### CI

- `Build & Type-check` ✅ pass
- `check-sprint` ✅ pass

Correctly addresses all review feedback. No new issues.

yujiawei (dismissed):
# Code Review — PR #81 (octo-adapters)

## TL;DR

This review focuses on commit `8d28b98`, which addresses the blocker raised in @Jerry-Xin's `CHANGES_REQUESTED` review on `14f3d83`: `detectInstallState()` previously classified ClawHub octo as healthy without distinguishing the dangerous dual-active state where legacy npm `openclaw-channel-octo` is still registered alongside it.

The fix is targeted, well-scoped, and the rationale for the chosen registration signal is sound. Type-check, build, and 99/99 tests pass per the author. No blocking correctness, security, or data-loss issues found in the delta. A handful of P2 nits are listed below — none gate merge.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Diff vs. PR description | ✅ | 11 files / +684 / -38 (5 new tests in delta commit, 19 total new test cases on the PR) |
| Blocker fix landed | ✅ | `octo-clawhub` variant gains `legacyNpmActive: boolean` (`openclaw-cli.ts` `InstallState` definition) and `enforceHealthyClawHubInstall` now branches on it before proceeding (`utils.ts`) |
| Block precedence over warn | ✅ | In `enforceHealthyClawHubInstall`, `legacyNpmActive` block is checked **before** `npmResidue` warn — correct ordering |
| Banner precedence | ✅ | `renderInstallStatusBanner` uses `if (legacyNpmActive) … else if (npmResidue) …` so the dual-active error supersedes the residue warning |
| Test coverage for the fix | ✅ | 5 new `detectInstallState` cases: clean ClawHub, dir residue only, registered legacy npm, registered + residue, and the npm-legacy fallback path |
| Other Jerry-Xin scenarios | ✅ | Coverage matrix matches what was requested (ClawHub-only, ClawHub + dir residue, ClawHub + registered legacy, npm-only legacy) — only dmwork-only legacy is still indirect, but `dmwork-legacy` paths in `detectInstallState` are exercised through the `isHealthyInstall` calls |
| Switch exhaustiveness preserved | ✅ | `enforceHealthyClawHubInstall` still covers all 5 `InstallState` kinds; no fall-through |
| Pre-flight integration intact | ✅ | `bind.ts`, `quickstart.ts`, `remove-account.ts` continue to call `enforceHealthyClawHubInstall()` as the single source of truth |
| `info` cleanup (`npm package`) | ✅ | `cli/index.ts` line removed; `NPM_PACKAGE_NAME` no longer imported there |
| README global-install warning | ✅ | covers `npm uninstall -g openclaw-channel-octo` migration |
| `ci.yml` paths-filter cleanup | ✅ | matches PR description |

## 2. Security-sensitive review

The dispatcher tagged this PR `security_sensitive` based on the migration/legacy keyword filter. The actual diff does **not** touch auth, token storage, credential redaction, or the binding-write code path — only the pre-flight gate that decides whether to enter those code paths.

- No new use of `eval`, shell-string concatenation, or untrusted-path construction. `runOpenclaw` / `runCmd` continue to use `execFileSync` with argv arrays.
- `printUpgradeNotice()` writes to stderr only. Strings are built from constants and a small set of typed `InstallState` fields; no caller-supplied string interpolation reaches the box.
- `enforceHealthyClawHubInstall` calls `process.exit(1)` directly on every failure branch — no exception leakage that could let a downstream code path proceed with stale assumptions.
- The new dual-active block is **fail-closed**: a registered legacy npm install now stops the user from writing into a duplicated `channels.octo` config object. This is a defensive improvement vs. the prior behavior, which silently wrote into the duplicated channel.
- Failure modes (false-allow / false-block) cause UX failures, not privilege escalation or token leak.

No additional security concerns to flag for human review beyond the points already labelled `needs-human-review`.

## 3. Findings

### Resolved from prior review round

**`legacyNpmActive` correctly distinguishes dual-active from dir-residue-only.** `cfg.plugins.entries[NPM_PACKAGE_NAME]` is the right signal: it is OpenClaw's load-time registry, so a non-empty entry means the runtime will still attempt to load the legacy plugin on startup, which is exactly the condition that produces the duplicated `octo` channel registration described in `install.ts`. Using `cfg.plugins.installs` as a co-signal is unnecessary (and per the author's investigation, not currently populated on live OpenClaw schemas).

### P2 — non-blocking nits

**P2-1. `isLegacyNpmStillRegistered` treats explicitly-disabled entries as "active".**

`cfg?.plugins?.entries?.[NPM_PACKAGE_NAME]` is truthy even when the entry is `{ enabled: false }`. In practice this is a conservative choice (any leftover entry is worth flagging), but the block message says "*still registered alongside ClawHub octo* — running this command would write into a duplicated-channel state". If OpenClaw skips disabled entries during channel registration, a `{ enabled: false }` entry would not actually duplicate the channel, and the user would be blocked unnecessarily. If you want to be precise, narrow to `entries[NPM_PACKAGE_NAME]?.enabled !== false`. If conservatism is intentional (recommended migration state should have zero entries left), keep as-is and consider tightening the block-message wording slightly. Not a blocker either way.

**P2-2. `detectInstallState` reads config twice on the ClawHub branch.**

`isLegacyNpmStillRegistered()` calls `readConfigFromFile()` in the `octo-clawhub` arm; the broken-state classifier later reads it again. On `bind` / `quickstart` / `remove-account` this is one extra disk read per pre-flight (~µs). Trivial. Worth a single local var only if you touch this file again.

**P2-3. Carrying forward from the prior review round.**

These three points still apply to the PR as a whole — none introduced by this delta:

- **`detectInstallState` and `enforceHealthyClawHubInstall` lack direct unit tests** for the npm 1.x belt-and-suspenders branch (`openclaw-cli.ts` fallback) and the leftover-config "broken" classifier. The new tests cover the dual-active scenarios but the broken-state branch remains untested.
- **Duplicated version-compare + floor.** `compareVersions` / `MIN_OPENCLAW_VERSION = "2026.4.15"` (`utils.ts`) and `compareOpenClawVersion` / `OPENCLAW_PEER_MIN = "2026.4.15"` (`openclaw-cli.ts`) coexist with identical lower-bound and identical integer-segment semantics. Drift risk if the floor moves and only one side is updated.
- **`switch (plugin.kind)` in `enforceHealthyClawHubInstall` has no `default: { const _exhaustive: never = plugin; throw ... }` arm.** TypeScript exhaustiveness via union type still flags additions today, but if a future contributor adds an `InstallState` variant and forgets to wire it, the switch silently returns and the caller proceeds as if healthy. Fail-closed default would harden this.

**P2-4. Doctor lost its strict openclaw-missing exit.**

`cli/index.ts` removes `ensureOpenClawCompat()` from the `doctor` command and adds the non-blocking `renderInstallStatusBanner()` instead. Net effect: when `openclaw` is not on PATH, doctor no longer exits early — it prints `✗ OpenClaw is not installed or not on PATH` then runs all checks (which will fail in their own ways via internal try/catch). This appears intentional (doctor is diagnostic; failing checks individually is more informative than a hard crash), but worth flagging in case the previous behaviour was load-bearing for any downstream tooling.

### Out of scope (noted, not raised)

- The `LEGACY_NPM_RESIDUE_PATH` cleanup is suggested in the warn message, but I didn't verify that `install` actually removes the npm residue directory. If it doesn't, the warning loops every run. Verifying that path is a follow-up task and outside this PR's diff. Author acknowledges related cleanup is tracked in #82.
- `hasLegacyNpmResidue()` hard-codes `homedir() + .openclaw/npm/...` rather than deriving from `getConfigFilePathSafe()`. Consistent with the legacy v1 npm install layout, but a user with a non-default OpenClaw config root wouldn't have residue at this path. Tracked under the broader detection-layer refactor in #82.

## 4. Recommendations

- Approve and merge.
- Open / append follow-ups for: P2-1 (decide on disabled-entry semantics), P2-3 (the three carry-forward items: direct tests for `detectInstallState`, collapse the version-compare helpers, exhaustive-switch hardening), and P2-4 (confirm the doctor behavior change is intentional).

Jerry-Xin (changes requested):
The PR is in scope for `octo-adapters`, but the new pre-flight misses an important legacy migration state.

**🔴 Blocking**
🔴 Critical: `detectInstallState()` returns `octo-clawhub` as soon as the current `octo` plugin is healthy, before checking for `openclaw-channel-dmwork`, `dmwork`, or `channels.dmwork` / `bindings(channel=dmwork)` residue. See `create-openclaw-octo/cli/openclaw-cli.ts:1058` and the later legacy checks at `create-openclaw-octo/cli/openclaw-cli.ts:1099`. Then `enforceHealthyClawHubInstall()` treats that as healthy and returns at `create-openclaw-octo/cli/utils.ts:244`, so `bind`, `quickstart`, and `remove-account` can proceed instead of forcing the user through `install`.

This conflicts with the existing migration model: `detectScenario()` intentionally prioritizes very-legacy and rebrand artifacts before the healthy `octo` update path (`create-openclaw-octo/cli/openclaw-cli.ts:967`, `create-openclaw-octo/cli/openclaw-cli.ts:975`), and `hasLegacyPluginArtifacts()` includes `channels.dmwork` and `bindings(channel=dmwork)` residue (`create-openclaw-octo/cli/openclaw-cli.ts:1703`). Existing migration tests also cover “octo installed + dmwork residue → rebrand finishes the data migration” (`create-openclaw-octo/cli/install-migration.test.ts:339`). The new pre-flight should block that same state and prompt `npx -y create-openclaw-octo install`, otherwise users can write fresh `channels.octo` data while legacy `dmwork` data still needs migration.

**✅ Highlights**
The OpenClaw version gating is simple and appropriate for the date-style version scheme.

The shared pre-flight reduces duplicated checks across `bind`, `quickstart`, and `remove-account`, which is the right direction once the legacy-state precedence is fixed.

caster-Q:
Addressed @Jerry-Xin's second blocker — pushed `df19bac`.

### Change

- `octo-clawhub` variant gains `legacyDmworkResidue: boolean`, sourced from the existing `hasLegacyPluginArtifacts(cfg)` helper. That helper already covers all 6 dmwork residue signals: extension dir / cfg.plugins.entries / cfg.plugins.installs / cfg.plugins.allow / `channels.dmwork` / `bindings(channel=dmwork)`.
- `enforceHealthyClawHubInstall()`: when `legacyDmworkResidue === true`, print a `block` upgrade notice and exit. Suggested remedy: `npx -y create-openclaw-octo install`, which drives `runRebrandMigration()` exactly as `install.ts` `detectScenario()` does for the same state.
- `renderInstallStatusBanner()`: dmwork residue line uses `✗` precedence over `npmResidue`'s `⚠`. Block precedence: `legacyNpmActive` → `legacyDmworkResidue` → `npmResidue` warn.
- 2 new tests: dmwork channel residue, dmwork binding residue. Existing 5 dual-active cases now also assert `legacyDmworkResidue: false`. **101/101 tests passing**, type-check clean.

### Why this matches `install.ts` precedence

`detectScenario()` priority is documented in `cli/openclaw-cli.ts`:
1. `hasVeryLegacyPluginArtifacts` → `legacy-to-octo`
2. `hasLegacyPluginArtifacts` → `rebrand` (triggers even when octo is also present — half-completed migrations)
3. octo healthy + clean → `update`

The pre-flight now blocks states (1) + (2) before bind/quickstart/remove-account writes can land in the half-migrated config. Users with healthy octo and dmwork residue see the block message and run install, which finishes the rebrand migration. Then the pre-flight passes on the next invocation.

### yujiawei P2 nits

- **P2-2** (cfg read twice): naturally addressed — `isLegacyNpmStillRegistered` now takes cfg as a parameter, called from the same place that reads it for `hasLegacyPluginArtifacts`.
- **P2-1** (`enabled: false` semantics), **P2-3** (carry-forwards: direct broken-state tests, version-compare consolidation, switch exhaustive default), **P2-4** (doctor strict-exit on missing openclaw): out of scope for the BLOCKER fix. The detection-layer cleanup in #82 is the natural place to bundle these.

@Jerry-Xin @yujiawei please re-review.

lml2468 (dismissed):
REVIEW_STATE: APPROVED

## Re-review (head: df19bac73b10, previous: APPROVED @ 8d28b98e)

### Jerry-Xin's dmwork-residue P0 resolved ✅

1. **`legacyDmworkResidue: boolean`** added to `InstallState` — uses existing `hasLegacyPluginArtifacts(cfg)` which checks `channels.dmwork`, `bindings(channel=dmwork)`, and plugin dir/entries/installs.
2. **`enforceHealthyClawHubInstall()` blocks** on dmwork residue with clear "Unfinished dmwork → octo migration" notice pointing to `install`.
3. **`renderInstallStatusBanner()`** surfaces it as ✗ error.
4. **Minor refactor**: `isLegacyNpmStillRegistered` now takes `cfg` param (avoids double `readConfigFromFile()` call in `detectInstallState`).

### New tests (+30 lines, 2 cases) ✅

- `octo-clawhub with dmwork channel residue → legacyDmworkResidue=true (BLOCK)`
- `octo-clawhub with dmwork binding residue → legacyDmworkResidue=true (BLOCK)`

### CI

- `Build & Type-check` ✅ pass
- `check-sprint` ✅ pass

Correctly mirrors `install.ts detectScenario` priority. No new issues.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but there is one blocking logic gap in the legacy dmwork pre-flight.

🔴 Blocking

🔴 Critical — Very-legacy `dmwork` artifacts are missed when `octo` is already healthy.  
In [create-openclaw-octo/cli/openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-qc7oode5/repo/create-openclaw-octo/cli/openclaw-cli.ts:1074), `detectInstallState()` sets `legacyDmworkResidue` using only `hasLegacyPluginArtifacts(cfg)`, which covers `openclaw-channel-dmwork`, `channels.dmwork`, and dmwork bindings. It does not include `hasVeryLegacyPluginArtifacts(cfg)`, even though `detectScenario()` gives that predicate first priority at [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-qc7oode5/repo/create-openclaw-octo/cli/openclaw-cli.ts:969). As a result, a machine with healthy ClawHub `octo` plus stale very-legacy plugin id `dmwork` entries/installs/allow/extension dir, but no `channels.dmwork` or dmwork bindings, will be classified as clean `octo-clawhub` and `bind` / `quickstart` / `remove-account` will proceed instead of forcing `install` to finish cleanup. This contradicts the PR’s force-upgrade goal and the installer’s migration priority.

Recommended fix: include `hasVeryLegacyPluginArtifacts(cfg)` in the `octo-clawhub` residue check, and add a regression test for “octo healthy + `plugins.entries.dmwork` / `plugins.installs.dmwork` residue blocks”.

💬 Non-blocking

No additional non-blocking findings.

✅ Highlights

The shared `enforceHealthyClawHubInstall()` path is a good consolidation of the previous scattered checks, and the version comparison tests cover the important integer-ordering boundaries.

yujiawei (approved):
# Code Review — PR #81 (octo-adapters)

## TL;DR

This review is on commit `df19bac`, which addresses the second blocker raised on `8d28b98`: the `octo-clawhub` arm of `detectInstallState()` previously returned healthy without checking for residual `dmwork` plugin/channel/binding artefacts that `install.ts detectScenario()` already classifies as a `rebrand` migration state. Without an equivalent pre-flight signal, `bind` / `quickstart` / `remove-account` could write fresh `channels.octo` data over an unfinished `dmwork → octo` migration.

The fix mirrors `install.ts` priority by reusing the existing `hasLegacyPluginArtifacts()` helper. It is targeted and well-scoped: 14 + 5 → 17 + 2 = 19 new test cases, 101/101 passing locally, type-check clean, build clean. No blocking correctness, security, or data-loss issues found in the delta.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Build (`npm run build`) | ✅ | `tsc` clean emit |
| Type-check (`npx tsc --noEmit`) | ✅ | clean |
| Tests (`npm test`) | ✅ | 101/101 passing locally (94 prior + 5 dual-npm + 2 dmwork) |
| Blocker fix landed | ✅ | `octo-clawhub` variant gains `legacyDmworkResidue: boolean` in `cli/openclaw-cli.ts:1019-1026`, populated via `hasLegacyPluginArtifacts(cfg)` at `cli/openclaw-cli.ts:1074` |
| `enforceHealthyClawHubInstall` blocks on the new flag | ✅ | `cli/utils.ts:270-279` prints a `block` notice and `process.exit(1)` |
| Doctor banner shows the new state | ✅ | `cli/utils.ts:350-352` adds the `✗ Unfinished dmwork → octo migration` line with precedence over `npmResidue`'s `⚠` |
| Block precedence | ✅ | `enforceHealthyClawHubInstall` checks `legacyNpmActive` → `legacyDmworkResidue` → `npmResidue` (warn). Both block branches exit before `npmResidue` warn. |
| Detection signal matches `install.ts` | ✅ | Reuses `hasLegacyPluginArtifacts()` (`cli/openclaw-cli.ts:1711`), which covers the same set as `detectScenario()`'s `rebrand` priority: dir / `plugins.entries` / `plugins.installs` / `plugins.allow` for `LEGACY_PLUGIN_ID` plus `channels.dmwork` plus `bindings(match.channel === "dmwork")`. |
| Test coverage for the fix | ✅ | 2 new `detectInstallState` cases for `channels.dmwork` and `bindings(channel=dmwork)`; existing 5 cases extended to assert `legacyDmworkResidue: false` so the field is checked on every healthy path |
| Single config read on the ClawHub branch | ✅ | `cli/openclaw-cli.ts:1062-1074` reads `cfg` once and threads it into `isLegacyNpmStillRegistered(cfg)` and `hasLegacyPluginArtifacts(cfg)` (resolves a P2 nit from the `8d28b98` review round) |
| Pre-flight wiring intact | ✅ | `bind.ts`, `quickstart.ts`, `remove-account.ts` continue to call `enforceHealthyClawHubInstall()` as the single source of truth |

## 2. Security-sensitive review

The dispatcher tagged this PR `security_sensitive` based on the migration / legacy keyword filter. The actual diff does NOT touch auth, token storage, credential redaction, or the binding-write code path — only the pre-flight gate that decides whether to enter those code paths.

- No new use of `eval`, shell-string concatenation, or untrusted-path construction. `runOpenclaw` / `runCmd` continue to use `execFileSync` with argv arrays.
- `printUpgradeNotice()` writes to stderr only. Strings are built from constants and a small set of typed `InstallState` fields; no caller-supplied string interpolation reaches the box.
- `enforceHealthyClawHubInstall` calls `process.exit(1)` directly on every failure branch — no exception leakage that could let a downstream code path proceed with stale assumptions.
- The new dmwork-residue gate is **fail-closed**: when residue is detected, the user is blocked from writing into a half-migrated config. This is a defensive improvement over the previous behaviour (which silently wrote into the new `channels.octo` while the unfinished `dmwork → octo` migration remained pending).
- Failure modes (false-allow / false-block) cause UX friction, not privilege escalation or token leak.

No additional security concerns to flag for human review beyond the points already labelled `needs-human-review`.

## 3. Findings

### Resolved from prior review rounds

- **dmwork-residue precedence (the second blocker)** — the `octo-clawhub` arm now refuses to proceed when any of the dmwork artefacts that `detectScenario()` already routes to `runRebrandMigration` are still present. Reusing `hasLegacyPluginArtifacts()` keeps the pre-flight signal in lock-step with `install.ts`, so future additions to the rebrand surface are picked up by both paths automatically.
- **`legacyNpmActive` precedence (the first blocker)** — preserved; checked first in both `enforceHealthyClawHubInstall` and the doctor banner.
- **Single config read on the ClawHub branch** — `cfg` is now loaded once and threaded into both helpers; the prior double-read on the happy path is gone.

### P2 — non-blocking nits (carried forward)

The following points are not introduced by this delta and were already flagged as P2 / out-of-scope in the prior round on `8d28b98`. None gate merge.

**P2-1. `legacyDmworkResidue` is "any artefact present", not "still registered".** `hasLegacyPluginArtifacts(cfg)` returns true even when `cfg.plugins.entries["openclaw-channel-dmwork"]?.enabled === false`, mirroring the same conservative posture as `legacyNpmActive`. In a fully-migrated state none of these markers should remain, so blocking on a leftover disabled entry is the correct fail-closed default. Worth narrowing later only if the install path is shown to leave disabled entries behind that are not data-bearing.

**P2-2. `detectInstallState` and `enforceHealthyClawHubInstall` still lack direct unit tests** for the npm 1.x belt-and-suspenders fallback (`cli/openclaw-cli.ts:1083-1101`) and the leftover-config "broken" classifier branch. The 19 new cases on the file cover the `octo-clawhub` arm thoroughly but the legacy / broken arms are still indirect. Tracked under #82 per the commit message.

**P2-3. Duplicated version-compare + floor.** `compareVersions` / `MIN_OPENCLAW_VERSION = "2026.4.15"` (`cli/utils.ts:37,~58`) and `compareOpenClawVersion` / `OPENCLAW_PEER_MIN = "2026.4.15"` (`cli/openclaw-cli.ts`) coexist with identical lower-bound and identical integer-segment semantics. `ensureOpenClawCompat` (still used by `install.ts`, `uninstall.ts`) uses the old pair; everything else uses the new pair. Drift risk if the floor moves and only one side is updated.

**P2-4. `switch (plugin.kind)` in `enforceHealthyClawHubInstall` has no `default: { const _exhaustive: never = plugin; throw ... }` arm.** TypeScript exhaustiveness via the discriminated union still flags additions today, but a future contributor adding a new `InstallState` variant who forgets to wire it in here would see the function silently return — and the caller would proceed as if healthy. Fail-closed default would harden this.

**P2-5. Doctor lost its strict openclaw-missing exit.** When `openclaw` is not on `PATH`, `doctor` no longer exits early — it prints the banner and runs all checks (which fail individually via internal try/catch). Likely intentional for a diagnostic command, but worth flagging if the previous behaviour was load-bearing for any downstream tooling.

### Out of scope (noted, not raised)

- Block-message wording mismatch when both `legacyNpmActive` and `legacyDmworkResidue` are true: only the legacy-npm message (`install --force`) is shown. Both block, and `detectScenario()` priorities mean a plain `install` would still pick up the dmwork rebrand path first, so the suggested command is at worst equivalent. Cosmetic only.
- `hasLegacyNpmResidue()` continues to hard-code `homedir() + .openclaw/npm/...` rather than deriving from `getConfigFilePathSafe()`. Tracked under the broader detection-layer refactor in #82 per author's note.

## 4. Recommendations

- Approve and merge.
- Open / append follow-ups for: P2-2 (direct tests for `detectInstallState` legacy / broken arms), P2-3 (collapse the version-compare helpers), P2-4 (exhaustive-switch hardening). Per the commit message, the detection-layer cleanup tracked in #82 is the natural home for these.

yujiawei (changes requested):
# Code Review — PR #81 (octo-adapters) — re-review on df19bac

## TL;DR

Updating my prior approval to `CHANGES_REQUESTED`. Apologies for the round-trip — on closer reading I missed the very-legacy gap raised in @Jerry-Xin's `df19bac` review, and that point is correct: the new `legacyDmworkResidue` signal does NOT mirror `detectScenario()`'s top-priority bucket. A healthy ClawHub octo machine that still has the very-legacy `dmwork` plugin id (`plugins.entries.dmwork`, `plugins.installs.dmwork`, `plugins.allow` containing `"dmwork"`, or `extensions/dmwork/`) but no `openclaw-channel-dmwork` artefacts and no `channels.dmwork` will currently be classified as clean `octo-clawhub` and the pre-flight will pass — even though `detectScenario()` would route the same machine to `legacy-to-octo` and run `runLegacyToOctoMigration`. That contradicts the PR's force-upgrade goal.

## 1. Verification (delta only)

| Item | Result |
|---|---|
| Build / type-check / `npm test` | ✅ 101/101 passing on `df19bac` |
| `legacyDmworkResidue` mirrors `detectScenario()` rebrand surface | ✅ |
| `legacyDmworkResidue` mirrors `detectScenario()` very-legacy surface | ❌ — missing `hasVeryLegacyPluginArtifacts(cfg)` |

## 2. Findings

### 🔴 P1 — Very-legacy `dmwork` plugin-id residue is not blocked on the `octo-clawhub` happy path

**Location.** `cli/openclaw-cli.ts:1058-1075` (`detectInstallState`'s `octo-clawhub` arm) populates only `legacyDmworkResidue: hasLegacyPluginArtifacts(cfg)`.

**Gap.** `hasLegacyPluginArtifacts(cfg)` (defined at `cli/openclaw-cli.ts:1711-1722`) only checks the **intermediate** `LEGACY_PLUGIN_ID = "openclaw-channel-dmwork"` plus `LEGACY_CHANNEL_ID = "dmwork"` (`channels.dmwork` + bindings whose `match.channel === "dmwork"`). It does NOT check the **very-legacy** `VERY_LEGACY_PLUGIN_ID = "dmwork"` — that surface lives in `hasVeryLegacyPluginArtifacts(cfg)` at `cli/openclaw-cli.ts:1695-1703` and covers `extensions/dmwork/`, `cfg.plugins.entries.dmwork`, `cfg.plugins.installs.dmwork`, `cfg.plugins.allow.includes("dmwork")`.

**Why this is a real gap, not just a corner case.** `detectScenario()` at `cli/openclaw-cli.ts:963-992` gives `hasVeryLegacyPluginArtifacts(cfg)` Priority 1 (returns `"legacy-to-octo"`), strictly above the rebrand bucket (Priority 2). The PR's commit message for `df19bac` explicitly states the goal is to "mirror the `install.ts` priority in the pre-flight," but the code mirrors only Priority 2, not Priority 1. So a machine with healthy ClawHub octo + leftover `dmwork` plugin-id artefacts (and no `openclaw-channel-dmwork` artefacts, and no `channels.dmwork` / dmwork bindings) is currently:

- routed to `runLegacyToOctoMigration()` by `install.ts` (correct)
- classified as `octo-clawhub` with `legacyDmworkResidue: false` by the pre-flight (incorrect)

`bind` / `quickstart` / `remove-account` will then write fresh `channels.octo` data while the very-legacy plugin id is still registered, which is exactly the half-migrated state the force-upgrade was meant to prevent.

**Suggested fix.**

```ts
// cli/openclaw-cli.ts:1074
legacyDmworkResidue:
  hasVeryLegacyPluginArtifacts(cfg) || hasLegacyPluginArtifacts(cfg),
```

Plus a regression test for "octo healthy + `plugins.entries.dmwork` residue → `legacyDmworkResidue: true` (BLOCK)" and "octo healthy + `extensions/dmwork/` dir residue → `legacyDmworkResidue: true` (BLOCK)" — symmetric to the two existing rebrand-residue tests at `cli/openclaw-cli.test.ts:794-813`.

### Resolved / preserved (still good)

- The dmwork-rebrand precedence fix itself (channels.dmwork, bindings(channel=dmwork), and the intermediate `openclaw-channel-dmwork` plugin id) is correct and well-tested.
- Block ordering in `enforceHealthyClawHubInstall` and `renderInstallStatusBanner` is correct: `legacyNpmActive` → `legacyDmworkResidue` → `npmResidue`.
- Single config read on the ClawHub branch (resolves the prior round's P2-2 nit).
- `process.exit(1)` on every block branch — fail-closed.

### P2 — non-blocking nits (carried forward, unchanged from prior round)

- **P2-1.** `legacyDmworkResidue` blocks even when the offending entries are explicitly disabled. Same conservative posture as `legacyNpmActive`. Acceptable.
- **P2-2.** `detectInstallState` and `enforceHealthyClawHubInstall` lack direct unit tests for the npm 1.x belt-and-suspenders fallback (`cli/openclaw-cli.ts:1083-1101`) and the leftover-config "broken" classifier branch.
- **P2-3.** Duplicated version-compare + floor: `compareVersions` / `MIN_OPENCLAW_VERSION` (`cli/utils.ts`) and `compareOpenClawVersion` / `OPENCLAW_PEER_MIN` (`cli/openclaw-cli.ts`) coexist with identical lower bounds.
- **P2-4.** `switch (plugin.kind)` in `enforceHealthyClawHubInstall` has no `default: { const _exhaustive: never = plugin; throw ... }` arm — fail-closed default would harden against future variant additions.
- **P2-5.** `doctor` lost its strict openclaw-missing exit. Likely intentional, but worth confirming.

## 3. Security-sensitive review

The dispatcher tagged this PR `security_sensitive` based on the migration / legacy keyword filter. The actual diff does NOT touch auth, token storage, credential redaction, or the binding-write code path — only the pre-flight gate. `printUpgradeNotice()` is stderr-only with constant strings; `process.exit(1)` on every block branch; no shell-string concatenation; `runOpenclaw` / `runCmd` continue to use `execFileSync` with argv arrays. The P1 gap above is a correctness bug (false-allow into an unfinished migration), not a privilege-escalation or token-leak concern.

## 4. Recommendations

- Address the P1 above (extend `legacyDmworkResidue` to include `hasVeryLegacyPluginArtifacts(cfg)`, add 1–2 regression tests). Once that lands, this PR is ready to merge in my view — the rest of the diff is solid.
- Open / append follow-ups for P2-2 / P2-3 / P2-4 in #82 per the existing detection-layer cleanup tracker.

caster-Q:
Addressed @Jerry-Xin's third blocker — pushed `2734b96`.

### Change

`legacyDmworkResidue` now combines both helpers, mirroring `install.ts detectScenario()` priority:

```ts
legacyDmworkResidue:
  hasVeryLegacyPluginArtifacts(cfg) || hasLegacyPluginArtifacts(cfg)
```

- `hasVeryLegacyPluginArtifacts` covers VERY_LEGACY_PLUGIN_ID = `dmwork` (priority 1 in `detectScenario()` → `legacy-to-octo` migration).
- `hasLegacyPluginArtifacts` covers LEGACY_PLUGIN_ID = `openclaw-channel-dmwork` + `channels.dmwork` + bindings (priority 2 → `rebrand` migration).

So the pre-flight now blocks **both** sets of stale residue, exactly matching the install-time migration priority. A machine with healthy octo + only `plugins.entries.dmwork` (very-legacy id, no rebrand-era artefacts) is now correctly flagged.

### Test

Added regression for the exact scenario you called out: `octo` healthy + `plugins.entries.dmwork` + `plugins.installs.dmwork` → `octo-clawhub` with `legacyDmworkResidue: true`.

**102/102 tests passing locally**, type-check clean.

@Jerry-Xin @yujiawei please re-review.

lml2468 (approved):
REVIEW_STATE: APPROVED

## Re-review (head: 2734b96efa0d, previous: APPROVED @ df19bac7)

### Jerry-Xin's very-legacy gap resolved ✅

One-line fix as expected:
```ts
legacyDmworkResidue:
  hasVeryLegacyPluginArtifacts(cfg) || hasLegacyPluginArtifacts(cfg),
```

- Very-legacy `dmwork` (bare plugin id) checked first, matching `detectScenario()` priority order.
- Updated comment explains both priority levels (legacy-to-octo migration + rebrand migration).
- Regression test: "octo-clawhub with very-legacy `dmwork` plugin entries → legacyDmworkResidue=true (BLOCK)".

### CI

- `Build & Type-check` ✅ pass
- `check-sprint` ✅ pass

No new issues. The pre-flight now covers the complete `detectScenario()` priority chain.

Jerry-Xin (approved):
The PR is in scope for `octo-adapters`: it modifies the `create-openclaw-octo` CLI migration/pre-flight path and related docs/CI.

✅ Highlights
- 🔵 Project relevance gate passed: changes are directly tied to the OpenClaw Octo CLI rename and migration flow.
- 🔵 `detectInstallState()` mirrors installer migration priorities and covers ClawHub, legacy npm, `dmwork`, broken, and clean-missing states in [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-83x1sv8l/repo/create-openclaw-octo/cli/openclaw-cli.ts:1058).
- 🔵 Shared pre-flight handling in [utils.ts](/Users/mlamp/.openclaw/tmp/pr-review-83x1sv8l/repo/create-openclaw-octo/cli/utils.ts:229) reduces duplicated compatibility/install checks across `bind`, `quickstart`, and `remove-account`.
- 🔵 Version comparison is intentionally integer-segment based and tested for double-digit month/patch cases in [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-83x1sv8l/repo/create-openclaw-octo/cli/openclaw-cli.ts:1175).

✅ Verification
- `npm ci` succeeded.
- `npm test` passed: 102 tests.
- `npx tsc --noEmit` passed.

No blocking findings.

yujiawei (approved):
# Code Review — PR #81 (octo-adapters)

## Summary

PR 2/2 of the CLI rename + force-upgrade migration. Adds the `detectInstallState` / `detectOpenClawState` pre-flight classifier in `cli/openclaw-cli.ts`, the boxed `printUpgradeNotice` + shared `enforceHealthyClawHubInstall` helper in `cli/utils.ts`, wires the helper into `bind` / `quickstart` / `remove-account`, surfaces a non-blocking banner in `doctor`, plus a small README + CI workflow nit.

I built locally on `2734b96`:

- `npx tsc --noEmit` — clean.
- `npm test` — 102/102 passing (PR description says 94/94; the actual count is higher but everything passes — see nit N1).
- `npm run build` — clean.

The detection logic mirrors `install.ts` `detectScenario` priority correctly (very-legacy `dmwork` and intermediate `openclaw-channel-dmwork` both block under one `legacyDmworkResidue` flag), the `octo-clawhub + legacy npm dual-active` case is explicitly handled before any config write, and all `process.exit(1)` paths fire **before** `bind` / `quickstart` / `remove-account` mutate `channels.octo` — which was the main risk vector for this PR.

No P0/P1 issues. A handful of P2 / nit suggestions below; none are merge blockers.

## P2 / suggestions

### S1 — `enforceHealthyClawHubInstall` may falsely block users who explicitly disabled the legacy npm plugin

`create-openclaw-octo/cli/openclaw-cli.ts:1054-1056`:

```ts
function isLegacyNpmStillRegistered(cfg: Record<string, any> | null): boolean {
  return Boolean(cfg?.plugins?.entries?.[NPM_PACKAGE_NAME]);
}
```

Elsewhere in this file (`openclaw-cli.ts:1480-1481`) the codebase explicitly writes `entries[id].enabled = false` when a plugin is disabled via the 4.20 fallback path. A user who ran `openclaw plugins disable openclaw-channel-octo` will still have an `entries["openclaw-channel-octo"]` object — just with `enabled: false`. A disabled entry won't be loaded by OpenClaw at startup, so the dual-channel-registration risk that motivates this BLOCK does not actually exist for them. They'll still get the `install --force` notice on every `bind` / `quickstart` / `remove-account` invocation.

Suggested tightening:

```ts
function isLegacyNpmStillRegistered(cfg: Record<string, any> | null): boolean {
  const entry = cfg?.plugins?.entries?.[NPM_PACKAGE_NAME];
  if (!entry) return false;
  // Treat explicit disable as "not registered for load" — the dual-channel
  // risk only exists when OpenClaw will actually try to load it.
  return entry.enabled !== false;
}
```

This keeps the safer-by-default behaviour (presence ⇒ flagged) while not penalising users who have explicitly stepped out of the load path.

### S2 — `compareOpenClawVersion` silently treats malformed segments as `0`

`openclaw-cli.ts:1163-1173`:

```ts
const pa = a.split(".").map(Number);
...
const va = pa[i] ?? 0;
```

`Number("x")` is `NaN`, and every `NaN < N` / `NaN > N` comparison is `false`, so `compareOpenClawVersion("2026.4.x", "2026.4.15") === 0` (treated as equal). In the current code the only caller is `detectOpenClawState` consuming `getOpenClawVersion`, which extracts only `\d{4}\.\d+\.\d+` matches — so reachable inputs are already sanitised and this is **not** an exploitable bug today.

But the function is `export`ed, the JSDoc invites future callers, and the integer-tuple test cases don't cover the NaN edge. If/when a future caller passes an unsanitised string, a malformed `2026.4.x` runtime would silently classify as "OK" rather than fail-closed. Two cheap mitigations:

- Add a guard: `if (Number.isNaN(va) || Number.isNaN(vb)) return ... ;` (recommend returning a `block` from `detectOpenClawState` rather than letting the comparator return 0).
- Or one regression test: `expect(compareOpenClawVersion("2026.x.0", "2026.4.15")).not.toBe(0)` to lock the contract in.

### S3 — Unused `ensureOpenClawCompat` import in `cli/index.ts`

`cli/index.ts:17`:

```ts
import { ensureOpenClawCompat, PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
```

After this PR, `ensureOpenClawCompat` is no longer called from `index.ts` (the `info` action and `doctor` action both dropped it — `info` now relies on `getOpenClawVersionStrict` directly, `doctor` now uses `renderInstallStatusBanner`). `tsconfig.json` doesn't enable `noUnusedLocals` so the build doesn't flag it, but it is dead. Drop it.

### S4 — `enforceHealthyClawHubInstall` has no compile-time exhaustiveness check on `InstallState.kind`

`cli/utils.ts:243-326` switches on `plugin.kind` with `case "octo-clawhub"` returning normally and the other four cases all calling `process.exit(1)`. If a sixth `InstallState` variant is added, TypeScript will not flag the missing case (control-flow analysis can't see the function as exhaustive because of `process.exit`'s return type). A `default: const _exhaustive: never = plugin; return _exhaustive;` (or equivalent helper) at the bottom would lock this in. Cheap insurance.

### S5 — Test ordering fragility documented but still fragile

`cli/openclaw-cli.test.ts:560-568` carries an explicit comment that `compareOpenClawVersion` / `detectOpenClawState` / `detectInstallState` blocks **must** stay before the `getConfigFilePathSafe` block because the latter's `vi.doMock("node:child_process", ...)` permanently rebinds the module mock. The comment is appreciated, but a test file whose semantics depend on `describe` ordering is a smell that future contributors will trip over. Two follow-up options (out of scope for this PR):

- Split the `getConfigFilePathSafe` block into its own file so the mock-pollution can't escape.
- Wrap each affected test in a `beforeEach`/`afterEach` that snapshots and restores the module-level mock binding.

Mentioning so it doesn't get lost — not a blocker for this PR.

## Nits / FYI

- **N1** PR description says "94/94 passing"; locally I see 102/102. PR description out of date but harmless. No action required.
- **N2** Comment on `cli/openclaw-cli.ts:1042-1052` is excellent — calls out the npm-1.0.0 misclassification risk, the belt-and-suspenders fallback, and the `installs` schema drift tracked in #82. Good to see this rationale committed.
- **N3** `cli/utils.ts:235-243` — `enforceHealthyClawHubInstall` evaluates `detectOpenClawState()` before `detectInstallState()`, so an OpenClaw-too-old user gets the OpenClaw upgrade prompt **before** any plugin-level prompt. That's the right precedence (no point asking them to migrate the plugin before the runtime can load it at all). Worth keeping.
- **N4** `.github/workflows/ci.yml` paths-filter cleanup is correctly limited to truly stale globs (`package.json`, `pnpm-lock.yaml`, `*.config.*`); the remaining `create-openclaw-octo/**` and `.github/workflows/**` filters cover everything that actually exists in the repo today.

## Security-sensitive checklist

The dispatcher labelled this `security_sensitive` due to the migration / pre-flight keyword hit. Reviewed with that lens:

- **No new external input parsing.** The new code reads `~/.openclaw/openclaw.json` (already-trusted local file) and `openclaw --version` stdout (already used by `getOpenClawVersionStrict`). No new network, no new IPC, no new spawn surfaces.
- **No new path construction from user input.** `LEGACY_NPM_RESIDUE_PATH` is a hardcoded constant resolved against `homedir()`. Safe.
- **`process.exit(1)` paths fire before any config mutation.** Verified in `bind.ts`, `quickstart.ts`, `remove-account.ts`. No partial-write risk introduced by the new pre-flight.
- **Pre-flight cannot be bypassed via env var or flag.** Intentional — the helper is unconditional. Good.
- **No secrets logged.** The boxed notice text is static; cfg parsing reads `entries`/`installs` keys but only inspects truthiness, never logs values.
- **The block messages tell users to run `npx -y create-openclaw-octo install [--force]`.** This is the same trusted CLI; no third-party package recommendation. Safe.

Manual-verify worth a human eye:
- A user with a partially-migrated dmwork install hitting `bind` for the first time after the upgrade will now block with "Unfinished dmwork → octo migration detected" — confirm the wording resonates with users who don't know what "dmwork" means today (the term has been removed from public-facing surfaces since the rebrand). Suggest considering "Legacy plugin migration not finished" or similar in a follow-up.

## Verdict

No P0/P1 blockers. Type-check clean, build clean, 102/102 tests passing, security-sensitive surface reviewed and looks safe. The five P2 items above are improvements, not preconditions for merge — they can be folded into a follow-up if the author prefers to keep this PR focused on the migration cutover.

PR #83, opened 2026-05-26
refactor(create-openclaw-octo): adopt `plugins list --json` as detection source

### LINKED ISSUE #82: Refactor detection layer to use `openclaw plugins list --json` as single source of truth
## Problem / Motivation

The plugin-installation detection layer in `create-openclaw-octo/cli/openclaw-cli.ts` currently combines multiple signals to decide whether a plugin id is installed and active:

1. `pluginsInspect(id)` — runs `openclaw plugins inspect <id> --json`
2. `cfg.plugins.entries[id]` — read from `openclaw.json`
3. `cfg.plugins.installs[id]` — read from `openclaw.json`
4. `existsSync(extensions/<id>)` — directory-on-disk probe

Investigation triggered by PR #81 review uncovered that this combination has accumulated correctness gaps:

### Gap 1 — `cfg.plugins.installs` is `null` on current OpenClaw

Inspecting a real-world OpenClaw config:

```bash
$ jq '.plugins | keys' ~/.openclaw/openclaw.json
[\"allow\", \"bundledDiscovery\", \"entries\"]

$ jq '.plugins.installs' ~/.openclaw/openclaw.json
null
```

`cfg.plugins.installs` is no longer present in the OpenClaw schema. Install metadata moved into the `openclaw plugins inspect <id> --json` output (under an \`install\` key — note: singular). This means:

- `isHealthyInstall()` fallback path `(hasDir && hasEntries && hasInstalls)` always returns `false` on current OpenClaw — effectively dead code.
- The fallback only \"works\" because the leading `inspectOk` short-circuit succeeds on supported OpenClaw versions. On older OpenClaw versions that do not support `plugins inspect`, the fallback is broken.

### Gap 2 — `extensions/<id>` directory probe misses npm-installed plugins

`isHealthyInstall()` checks `existsSync(<root>/extensions/<id>)`. This is the correct path for ClawHub installs (`PLUGIN_ID = \"octo\"`), but legacy npm-installed plugins (`NPM_PACKAGE_NAME = \"openclaw-channel-octo\"`, `LEGACY_PLUGIN_ID = \"openclaw-channel-dmwork\"`) live at `<root>/npm/node_modules/<id>` instead. A complete legacy npm install of either id can be misclassified as `broken` rather than the appropriate `octo-npm-legacy` / `dmwork-legacy` state on OpenClaw versions that do not support `plugins inspect`.

### Gap 3 — `hasLegacyNpmResidue()` hardcodes `homedir()/.openclaw`

```ts
const LEGACY_NPM_RESIDUE_PATH = \".openclaw/npm/node_modules/openclaw-channel-octo\";
function hasLegacyNpmResidue(): boolean {
  return existsSync(resolve(homedir(), LEGACY_NPM_RESIDUE_PATH));
}
```

Users with non-default OpenClaw installation roots (custom \`OPENCLAW_HOME\`, npm \`--prefix\`, Windows custom layout, etc.) will have their residue check silently fail. This signal is currently cosmetic-only (drives a warn banner, not a block decision), so the impact is low — but the path probe is wrong on principle.

## Proposed Solution

Introduce `openclaw plugins list --json` as the **single authoritative source** for \"which plugin ids are currently registered/loaded\":

```bash
$ openclaw plugins list --json | jq '[.plugins[] | {id, status, enabled, source, rootDir}]'
[
  {
    \"id\": \"octo\",
    \"status\": \"loaded\",
    \"enabled\": true,
    \"source\": \"/Users/.../extensions/octo/dist/index.js\",
    \"origin\": \"global\",
    \"rootDir\": \"/Users/.../extensions/octo\"
  },
  ...
]
```

This output is what OpenClaw itself reports about its loaded plugin registry. It supersedes the entries+installs+dir signal combination because:

- `id` and `status`/`enabled` directly answer \"is this plugin id registered + active?\"
- `rootDir` is reported by OpenClaw, eliminating any need to know its installation root client-side
- Works uniformly for ClawHub (`extensions/<id>`) and legacy npm (`npm/node_modules/<id>`) layouts — OpenClaw resolves `rootDir` per plugin

Refactor scope:

1. Add `listLoadedPlugins()` helper wrapping `runOpenclaw([\"plugins\", \"list\", \"--json\"])` with `stripStdoutNoise` + `JSON.parse`.
2. Rewrite `detectInstallState()` to source `legacyNpmActive` / `legacyDmworkActive` / version detection from `listLoadedPlugins()` rather than the entries/installs/dir combination.
3. Simplify `isHealthyInstall()` — drop the `hasInstalls` term from the fallback branch (always-false dead code).
4. Generalize the legacy-npm check to cover both `NPM_PACKAGE_NAME` and `LEGACY_PLUGIN_ID` via the same code path (closes Gap 2).
5. Rewrite `hasLegacyNpmResidue()` to derive its probe path from `getConfigFilePathSafe()` (closes Gap 3). Note: this signal will still be cosmetic-only after the refactor.

### Backward compatibility

`plugins list --json` should be available on any OpenClaw version that ships ClawHub (i.e., the supported target of this CLI). Behavior on very old OpenClaw versions that do not support `plugins list --json` is graceful degradation: fall back to `cfg.plugins.entries[id]` for registration detection.

## Alternatives Considered

- **Keep the entries + dir combination, just patch each gap individually.** Rejected — the underlying issue is that the detection layer is built from indirect signals (cfg fields, on-disk directories) when OpenClaw exposes a direct one (`plugins list`). Patching each indirect signal individually is incremental tech debt; consolidating to the direct signal is one change that retires the whole class.
- **Use `plugins inspect <id> --json` per id.** Rejected — `inspect` requires a known id; we cannot use it to *enumerate* what is currently installed (e.g., to detect that a legacy id we do not yet know about is registered). `plugins list --json` enumerates.

## Component

OpenClaw Channel Plugin (CLI under `create-openclaw-octo/`)

## Additional Context

- Triggered by review on https://github.com/Mininglamp-OSS/octo-adapters/pull/81 (Jerry-Xin BLOCKER — dual-active state misclassified as healthy because `cfg.plugins.installs` lookup always returns falsy on current OpenClaw schema).
- PR #81 ships a minimal patch using `cfg.plugins.entries[id]` for the BLOCKER fix. The full detection-layer rewrite is deferred to this issue to keep PR #81 focused on the Plan A force-upgrade feature.
- Two findings from prior `codex` review rounds map onto this scope:
  - Round 2 #1 (dual-active detection) — partially addressed by PR #81's minimal fix; **fully** addressed by this refactor.
  - Round 2 #2 (`LEGACY_PLUGIN_ID` extensions-only fallback misclassifies dmwork npm installs) — addressed by Gap 2 fix above.

caster-Q:
## Summary

Closes #82.

The pre-flight install-state detection layer in `cli/openclaw-cli.ts` previously combined four indirect signals to decide whether a plugin id was registered with OpenClaw: `pluginsInspect()`, `cfg.plugins.entries[id]`, `cfg.plugins.installs[id]`, and `existsSync(extensions/<id>)`. PR #81's review (and the linked issue #82) found three correctness gaps with that combination — see #82 for the full background. This PR consolidates the detection behind `openclaw plugins list --json`, OpenClaw's own authoritative registry of "which plugin ids the runtime is currently trying to load."

No behavioural change for the happy path (clean ClawHub octo, no legacy residue). Behavioural improvement for the gaps:

- **Custom OpenClaw root.** `hasLegacyNpmResidue()` now derives its probe path from `getConfigFilePathSafe()` rather than `homedir() + .openclaw/npm/...`. Users with `OPENCLAW_HOME` set, npm `--prefix`, or Windows non-default layouts now get the residue warn correctly. (The signal remains cosmetic — drives the warn line, never the block decision.)
- **npm-installed legacy dmwork plugin.** `LEGACY_PLUGIN_ID = "openclaw-channel-dmwork"` is an npm package that lives under `<root>/npm/node_modules/<id>`, not `<root>/extensions/<id>`. The previous `isHealthyInstall(LEGACY_PLUGIN_ID)` directory probe always missed it. `detectInstallState` now classifies a registered LEGACY_PLUGIN_ID as `dmwork-legacy` via `listLoadedPlugins()` (and via `cfg.plugins.entries` on old runtimes that don't support `plugins list --json`).
- **Dual-active state on stale cfg.** `legacyNpmActive` / `legacyDmworkResidue` previously sourced from `cfg.plugins.entries` only. They now prefer the live `plugins list --json` signal when available, OR the cfg.entries fallback when not. Catches a runtime registration that hasn't yet round-tripped to disk.

## Related Issue

Closes #82.

### `cli/openclaw-cli.ts`

- New `listLoadedPlugins(): { supported: boolean; plugins: PluginListEntry[] }` helper wrapping `runOpenclaw([\"plugins\", \"list\", \"--json\"])` + `stripStdoutNoise` + `JSON.parse`. Returns `supported: false` when the command throws, output is not JSON, or the JSON has no `plugins` array (i.e., old OpenClaw runtimes that don't recognise the subcommand).
- New `isPluginRegistered(id, list, cfg)` private helper: prefers the list signal when supported, falls back to `cfg.plugins.entries[id]`. Centralises the source-of-truth selection.
- `detectInstallState()` rewritten to use `isPluginRegistered`:
  - `legacyNpmActive` and `legacyDmworkResidue` (in the `octo-clawhub` arm) source their registration signals from list-or-cfg, OR'd with the existing channel/binding artefact checks (`hasLegacyPluginArtifacts`, `hasVeryLegacyPluginArtifacts`).
  - The `octo-npm-legacy` belt-and-suspenders branch and both `dmwork-legacy` branches use `isPluginRegistered` instead of the directory-only `isHealthyInstall` for the legacy ids — so npm-layout installs are caught.
  - Single `cfg = readConfigFromFile()` at the top, threaded into every helper that needs it.
- `hasLegacyNpmResidue()` derives its probe path from `getConfigFilePathSafe()` (\"openclaw.json\" → root → `npm/node_modules/openclaw-channel-octo`) instead of hardcoded `homedir() + .openclaw/npm/...`.
- `isHealthyInstall()` simplified: drops the `hasInstalls` term from the fallback. `cfg.plugins.installs` has been observed null on current OpenClaw schemas, making that term effectively dead. Fallback now requires only `dir + entries`.

### `cli/openclaw-cli.test.ts`

- 6 new unit tests for `listLoadedPlugins`: parse OK with full plugin shape, strip preceding banner/log noise, return `supported: false` when the command throws, return `supported: false` on non-JSON output, return `supported: false` on JSON missing `plugins` array, filter out entries with empty/missing `id`.
- 3 new `detectInstallState` cases covering the list signal path with cfg.entries empty:
  - `legacyNpmActive` sourced from list (`openclaw-channel-octo` only present in `plugins list --json`).
  - `legacyDmworkResidue` sourced from list with `LEGACY_PLUGIN_ID = \"openclaw-channel-dmwork\"`.
  - `legacyDmworkResidue` sourced from list with `VERY_LEGACY_PLUGIN_ID = \"dmwork\"`.
- `setupOctoClawHub` helper extended with optional `pluginsListSupported` / `legacyNpmInList` / `legacyDmworkInList` / `veryLegacyDmworkInList` opts. All default false, so every prior test still exercises the cfg-fallback path. The 102 pre-existing tests remain unchanged.

## Testing

- [x] **111/111 tests passing locally** (102 prior + 6 listLoadedPlugins + 3 detectInstallState list-signal cases)
- [x] `npx tsc --noEmit` — clean
- [x] `npm run build` — clean

## Checklist

- [x] Updated documentation — N/A (internal refactor; no user-facing behaviour change on the happy path)

## Backward compatibility

- **Modern OpenClaw (`plugins list --json` supported):** Detection now uses the list signal as primary source. Behaviour is functionally equivalent to the prior cfg-based detection on healthy installs, but more accurate on the gaps documented above.
- **Old OpenClaw (`plugins list --json` not supported):** `listLoadedPlugins()` reports `supported: false` and `isPluginRegistered` falls back to `cfg.plugins.entries[id]`. Behaviour is functionally equivalent to PR #81's `cfg.plugins.entries` check (which was added specifically because `cfg.plugins.installs` is null on current schemas).
- **Out-of-scope** for this PR: yujiawei's P2 carry-forwards from PR #81 (direct unit tests for `detectInstallState`'s broken-state classifier, `compareVersions` / `compareOpenClawVersion` consolidation, exhaustive-switch hardening). Those remain candidates for a separate small cleanup PR.

### DIFF
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -138,6 +138,53 @@ describe("listLoadedPlugins", () => {
     expect(result.plugins[0].id).toBe("octo");
   });
 
+  it("tolerates trailing log noise after JSON object via brace-matching", async () => {
+    // Some OpenClaw runtimes emit warnings AFTER the JSON object (e.g.
+    // gateway shutdown warnings on subsequent commands). A naive
+    // `JSON.parse(out.slice(start))` would fail because the slice includes
+    // the trailing log lines; brace-matching extracts just the balanced
+    // top-level object.
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(
+      JSON.stringify({ plugins: [{ id: "octo", status: "loaded", enabled: true }] }) +
+        "\n[plugins] reload complete\n" +
+        "Warning: deprecated config key X\n",
+    );
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(true);
+    expect(result.plugins).toHaveLength(1);
+    expect(result.plugins[0].id).toBe("octo");
+  });
+
+  it("brace-matching handles strings containing curly braces correctly", async () => {
+    // Defense against a flawed brace-counter: source paths can contain
+    // literal "}" or "{" inside string values. The matcher must respect
+    // string boundaries so it doesn't terminate the object prematurely.
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(JSON.stringify({
+      plugins: [
+        { id: "octo", status: "loaded", enabled: true, source: "/path/with/{braces}/in/it" },
+      ],
+    }));
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(true);
+    expect(result.plugins[0].source).toContain("{braces}");
+  });
+
+  it("identifies `unknown command` as unsupported (old OpenClaw)", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockImplementation(() => {
+      const err = new Error("error: unknown command 'list'") as any;
+      err.stderr = "error: unknown command 'list'";
+      throw err;
+    });
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+  });
+
   it("reports supported=false on old OpenClaw without `plugins list --json`", async () => {
     const { listLoadedPlugins } = await loadModule();
     mockExecFileSync.mockImplementation(() => {
@@ -1048,6 +1095,55 @@ describe("detectInstallState", () => {
     expect(state.kind).toBe("octo-npm-legacy");
     if (state.kind === "octo-npm-legacy") {
       expect(state.version).toBe("1.0.0");
+    }
+  });
+
+  it("dmwork-legacy: both ids registered → classifier picks VERY_LEGACY first (priority 1 in detectScenario)", async () => {
+    // Regression for codex review on PR #83: when both `dmwork`
+    // (VERY_LEGACY_PLUGIN_ID) and `openclaw-channel-dmwork`
+    // (LEGACY_PLUGIN_ID) are registered with no healthy octo present,
+    // detectInstallState must mirror detectScenario()'s priority and
+    // report the very-legacy id's version. Otherwise the pre-flight
+    // surfaces a different version than the migration install.ts will
+    // actually run (legacy-to-octo, not rebrand).
+    const { existsSync, readFileSync } = await import("node:fs");
+    mockExecFileSync.mockImplementation((cmd, args) => {
+      const argsArr = args as string[];
+      if (argsArr[0] === "config" && argsArr[1] === "file") {
+        return "/home/user/.openclaw/openclaw.json";
+      }
+      if (argsArr[0] === "plugins" && argsArr[1] === "list" && argsArr[2] === "--json") {
+        return JSON.stringify({
+          plugins: [
+            { id: "dmwork", status: "loaded", enabled: true },
+            { id: "openclaw-channel-dmwork", status: "loaded", enabled: true },
+          ],
+        });
+      }
+      if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
+        throw new Error("error: unknown command 'inspect'");
+      }
+      return "";
+    });
+    vi.mocked(readFileSync).mockReturnValue(JSON.stringify({
+      plugins: {
+        entries: { dmwork: { enabled: true }, "openclaw-channel-dmwork": { enabled: true } },
+        installs: {
+          dmwork: { version: "0.5.21", installPath: "~/.openclaw/extensions/dmwork" },
+          "openclaw-channel-dmwork": { version: "0.6.5", installPath: "~/.openclaw/npm/node_modules/openclaw-channel-dmwork" },
+        },
+      },
+    }));
+    vi.mocked(existsSync).mockReturnValue(false);
+
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("dmwork-legacy");
+    if (state.kind === "dmwork-legacy") {
+      // Must pick the VERY_LEGACY (`dmwork`) version, not the LEGACY
+      // (`openclaw-channel-dmwork`) version — matches detectScenario()
+      // priority: `legacy-to-octo` migration runs first.
+      expect(state.version).toBe("0.5.21");
     }
   });
 
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -93,6 +93,95 @@ describe("pluginsInspect", () => {
   });
 });
 
+describe("listLoadedPlugins", () => {
+  it("parses plugins list --json output", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(JSON.stringify({
+      plugins: [
+        {
+          id: "octo",
+          status: "loaded",
+          enabled: true,
+          source: "/home/u/.openclaw/extensions/octo/dist/index.js",
+          origin: "global",
+          rootDir: "/home/u/.openclaw/extensions/octo",
+        },
+        {
+          id: "openclaw-channel-octo",
+          status: "loaded",
+          enabled: true,
+          source: "/home/u/.openclaw/npm/node_modules/openclaw-channel-octo/dist/index.js",
+          origin: "npm",
+          rootDir: "/home/u/.openclaw/npm/node_modules/openclaw-channel-octo",
+        },
+      ],
+    }));
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(true);
+    expect(result.plugins).toHaveLength(2);
+    expect(result.plugins[0].id).toBe("octo");
+    expect(result.plugins[1].id).toBe("openclaw-channel-octo");
+    expect(result.plugins[1].rootDir).toContain("npm/node_modules");
+  });
+
+  it("strips preceding log/banner noise before JSON parse", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(
+      "🦞 OpenClaw 2026.5.18\n" +
+        "[plugins] loading...\n" +
+        JSON.stringify({ plugins: [{ id: "octo", status: "loaded", enabled: true }] }),
+    );
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(true);
+    expect(result.plugins[0].id).toBe("octo");
+  });
+
+  it("reports supported=false on old OpenClaw without `plugins list --json`", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockImplementation(() => {
+      throw new Error("error: unknown command 'list'");
+    });
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+    expect(result.plugins).toEqual([]);
+  });
+
+  it("reports supported=false when output is not JSON", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue("usage: openclaw plugins list ...");
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+  });
+
+  it("reports supported=false when JSON has no plugins array", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(JSON.stringify({ unexpected: "shape" }));
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+  });
+
+  it("filters out entries with missing id", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue(JSON.stringify({
+      plugins: [
+        { id: "octo", status: "loaded", enabled: true },
+        { /* no id */ status: "loaded" },
+        { id: "", status: "loaded" },
+      ],
+    }));
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(true);
+    expect(result.plugins).toHaveLength(1);
+    expect(result.plugins[0].id).toBe("octo");
+  });
+});
+
 describe("getOpenClawVersion", () => {
   it("should extract version from openclaw --version output", async () => {
     const { getOpenClawVersion } = await loadModule();
@@ -697,13 +786,44 @@ describe("detectInstallState", () => {
     dmworkBinding?: boolean;
     veryLegacyDmworkInEntries?: boolean;
     veryLegacyDmworkInInstalls?: boolean;
+    /**
+     * When true, `openclaw plugins list --json` returns a parseable
+     * payload containing octo + any of the optional `*InList` ids.
+     * When false (default), the command throws — simulating an old
+     * OpenClaw runtime that doesn't support `plugins list --json` and
+     * exercising the cfg.entries fallback path inside isPluginRegistered.
+     */
+    pluginsListSupported?: boolean;
+    legacyNpmInList?: boolean;
+    legacyDmworkInList?: boolean;
+    veryLegacyDmworkInList?: boolean;
   }) {
     return async () => {
       const { existsSync, readFileSync } = await import("node:fs");
+      const mkListEntry = (id: string) => ({
+        id,
+        status: "loaded",
+        enabled: true,
+        source: `/home/user/.openclaw/extensions/${id}/dist/index.js`,
+        origin: "global",
+        rootDir: `/home/user/.openclaw/extensions/${id}`,
+      });
+      const listPayload = (() => {
+        if (!opts.pluginsListSupported) return null;
+        const plugins = [mkListEntry("octo")];
+        if (opts.legacyNpmInList) plugins.push(mkListEntry("openclaw-channel-octo"));
+        if (opts.legacyDmworkInList) plugins.push(mkListEntry("openclaw-channel-dmwork"));
+        if (opts.veryLegacyDmworkInList) plugins.push(mkListEntry("dmwork"));
+        return JSON.stringify({ plugins });
+      })();
       mockExecFileSync.mockImplementation((cmd, args) => {
         const argsArr = args as string[];
         if (argsArr[0] === "config" && argsArr[1] === "file") {
           return "/home/user/.openclaw/openclaw.json";
+        }
+        if (argsArr[0] === "plugins" && argsArr[1] === "list" && argsArr[2] === "--json") {
+          if (listPayload) return listPayload;
+          throw new Error("error: unknown command 'list'");
         }
         if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
           // simulate old OpenClaw → fallback path inside isHealthyInstall
@@ -833,6 +953,59 @@ describe("detectInstallState", () => {
     await setupOctoClawHub({
       veryLegacyDmworkInEntries: true,
       veryLegacyDmworkInInstalls: true,
+    })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyDmworkResidue).toBe(true);
+      expect(state.legacyNpmActive).toBe(false);
+    }
+  });
+
+  // ── plugins list --json signal path (#82 detection-layer refactor) ──
+
+  it("octo-clawhub: legacyNpmActive sourced from `plugins list --json` (cfg.entries empty)", async () => {
+    // Modern OpenClaw reports the legacy npm openclaw-channel-octo through
+    // `plugins list --json`. Even when cfg.plugins.entries doesn't yet
+    // reflect it (e.g. stale cfg snapshot vs runtime), the list signal is
+    // the authoritative source.
+    await setupOctoClawHub({
+      pluginsListSupported: true,
+      legacyNpmInList: true,
+    })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyNpmActive).toBe(true);
+    }
+  });
+
+  it("octo-clawhub: legacyDmworkResidue sourced from `plugins list --json` (LEGACY_PLUGIN_ID)", async () => {
+    // Modern OpenClaw with intermediate dmwork id (`openclaw-channel-dmwork`)
+    // active. The list signal catches it without depending on
+    // `extensions/<id>` directory probing — important because dmwork
+    // legacy is npm-installed (lives under npm/node_modules/<id>).
+    await setupOctoClawHub({
+      pluginsListSupported: true,
+      legacyDmworkInList: true,
+    })();
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("octo-clawhub");
+    if (state.kind === "octo-clawhub") {
+      expect(state.legacyDmworkResidue).toBe(true);
+      expect(state.legacyNpmActive).toBe(false);
+    }
+  });
+
+  it("octo-clawhub: legacyDmworkResidue sourced from `plugins list --json` (VERY_LEGACY_PLUGIN_ID)", async () => {
+    // Modern OpenClaw with very-legacy `dmwork` id active. Mirrors
+    // detectScenario()'s priority-1 bucket through the list signal.
+    await setupOctoClawHub({
+      pluginsListSupported: true,
+      veryLegacyDmworkInList: true,
     })();
     const { detectInstallState } = await loadModule();
     const state = detectInstallState();
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -183,6 +183,33 @@ describe("listLoadedPlugins", () => {
 
     const result = listLoadedPlugins();
     expect(result.supported).toBe(false);
+    expect(result.error).toBeNull();
+  });
+
+  it("reports a real error (e.g. EACCES) via `error` field, not unsupported", async () => {
+    // Distinguishes "old OpenClaw doesn't recognise the subcommand"
+    // (supported=false, error=null) from genuine errors (config corruption,
+    // permission, plugin load crash) where the runtime cannot reliably
+    // tell us what's loaded.
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockImplementation(() => {
+      const err = new Error("EACCES: permission denied") as any;
+      err.stderr = "EACCES: permission denied";
+      throw err;
+    });
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+    expect(result.error).toMatch(/permission denied/i);
+  });
+
+  it("reports parse failure via `error`, not as old-runtime", async () => {
+    const { listLoadedPlugins } = await loadModule();
+    mockExecFileSync.mockReturnValue("usage: openclaw plugins list ...");
+
+    const result = listLoadedPlugins();
+    expect(result.supported).toBe(false);
+    expect(result.error).toMatch(/no JSON|JSON/);
   });
 
   it("reports supported=false on old OpenClaw without `plugins list --json`", async () => {
@@ -202,6 +229,7 @@ describe("listLoadedPlugins", () => {
 
     const result = listLoadedPlugins();
     expect(result.supported).toBe(false);
+    expect(result.error).toMatch(/no JSON/);
   });
 
   it("reports supported=false when JSON has no plugins array", async () => {
@@ -210,6 +238,7 @@ describe("listLoadedPlugins", () => {
 
     const result = listLoadedPlugins();
     expect(result.supported).toBe(false);
+    expect(result.error).toMatch(/missing.*plugins.*array/i);
   });
 
   it("filters out entries with missing id", async () => {
@@ -1070,6 +1099,12 @@ describe("detectInstallState", () => {
       if (argsArr[0] === "config" && argsArr[1] === "file") {
         return "/home/user/.openclaw/openclaw.json";
       }
+      if (argsArr[0] === "plugins" && argsArr[1] === "list" && argsArr[2] === "--json") {
+        // Old OpenClaw — not supported (caller falls back to cfg.entries)
+        const e = new Error("error: unknown command 'list'") as any;
+        e.stderr = "error: unknown command 'list'";
+        throw e;
+      }
       if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
         throw new Error("error: unknown command 'inspect'");
       }
@@ -1184,6 +1219,49 @@ describe("detectInstallState", () => {
     expect(state.kind).toBe("broken");
     if (state.kind === "broken") {
       expect(state.details).toContain("openclaw-channel-octo");
+    }
+  });
+
+  it("fail-closed when `plugins list --json` errors for a real reason (not unsupported)", async () => {
+    // Regression for codex review: when the OpenClaw runtime is in a
+    // failure state (config corruption, permission denied, plugin load
+    // crash, partial JSON), `plugins list --json` cannot reliably report
+    // what's loaded. Trusting cfg.entries alone risks silently allowing
+    // a dual-active state to slip through bind/quickstart/remove-account.
+    // detectInstallState must surface this as `broken` so the user is
+    // prompted to repair OpenClaw before any further config writes.
+    const { existsSync, readFileSync } = await import("node:fs");
+    mockExecFileSync.mockImplementation((cmd, args) => {
+      const argsArr = args as string[];
+      if (argsArr[0] === "config" && argsArr[1] === "file") {
+        return "/home/user/.openclaw/openclaw.json";
+      }
+      if (argsArr[0] === "plugins" && argsArr[1] === "list" && argsArr[2] === "--json") {
+        const e = new Error("EACCES: permission denied") as any;
+        e.stderr = "EACCES: permission denied, open '/home/user/.openclaw/openclaw.json'";
+        throw e;
+      }
+      if (argsArr[0] === "plugins" && argsArr[1] === "inspect") {
+        throw new Error("error: unknown command 'inspect'");
+      }
+      return "";
+    });
+    // Even with a healthy-looking cfg, the list-failed signal must
+    // dominate.
+    vi.mocked(readFileSync).mockReturnValue(JSON.stringify({
+      plugins: {
+        entries: { octo: { enabled: true } },
+        installs: { octo: { version: "1.0.12", installPath: "~/.openclaw/extensions/octo" } },
+      },
+    }));
+    vi.mocked(existsSync).mockReturnValue(true);
+
+    const { detectInstallState } = await loadModule();
+    const state = detectInstallState();
+    expect(state.kind).toBe("broken");
+    if (state.kind === "broken") {
+      expect(state.details).toMatch(/plugins list/);
+      expect(state.details).toMatch(/permission denied/i);
     }
   });
 });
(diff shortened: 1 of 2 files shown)


lml2468 (approved):
[APPROVE]

**Scope:** Refactor install-state detection to use `openclaw plugins list --json` as primary signal, closing #82.

**Reviewed changes (+335 −52, 2 files):**

1. **`listLoadedPlugins()`** — clean wrapper with robust error handling: catches command throw, non-JSON output, missing `plugins` array, entries with empty/missing `id`. Correctly reports `supported: false` for old runtimes.

2. **`isPluginRegistered(id, list, cfg)`** — centralizes list-vs-cfg signal selection. Prefers live `plugins list --json` when supported, falls back to `cfg.plugins.entries[id]`. Single source of truth.

3. **`hasLegacyNpmResidue()`** — now derives probe path from `getConfigFilePathSafe()` instead of hardcoded `homedir() + .openclaw/...`. Correctly handles custom `OPENCLAW_HOME` / Windows layouts.

4. **`isHealthyInstall()`** — drops dead `cfg.plugins.installs` term (observed null on current schemas). Simplified to `inspectOk || (hasDir && hasEntries)`.

5. **`detectInstallState()`** — single `readConfigFromFile()` + `listLoadedPlugins()` call at top, threaded through all helpers. No redundant reads. `legacyNpmActive` and `legacyDmworkResidue` now catch both cfg-registered AND runtime-registered legacy plugins.

6. **Tests** — 9 new cases: 6 for `listLoadedPlugins` (parse, noise strip, unsupported, non-JSON, missing array, empty id filter) + 3 for list-signal detection path (`legacyNpmActive`, `legacyDmworkResidue` via LEGACY and VERY_LEGACY ids). All 102 existing tests remain on cfg-fallback path via `pluginsListSupported: false` default.

**CI:** Build & Type-check ✅, CodeQL pending, `check-sprint` ❌ (Sprint W22).

All three #82 gaps addressed. Backward-compatible. LGTM.

lml2468 (dismissed):
REVIEW_STATE: APPROVED

## Summary

Clean refactoring that consolidates plugin detection behind `openclaw plugins list --json` (closes #82). Well-architected with proper graceful degradation for old runtimes.

### Architecture ✅

- **`listLoadedPlugins()`** — Authoritative signal from OpenClaw's runtime registry. Robust parsing: `stripStdoutNoise` handles log/banner noise, graceful `supported: false` for old runtimes (command unknown, non-JSON output, missing `plugins` array), filters entries with missing id.
- **`isPluginRegistered(id, list, cfg)`** — Unified source-of-truth selector: prefers list signal when supported, falls back to `cfg.plugins.entries[id]`. Replaces scattered `isLegacyNpmStillRegistered` / direct cfg checks.
- **`detectInstallState()`** — Now uses `isPluginRegistered` for all legacy-detection paths. Catches npm-installed plugins that the old `extensions/<id>` directory probe missed.
- **`isHealthyInstall()` simplified** — Drops dead `cfg.plugins.installs` term, early-returns on `pluginsInspect` success. Clear doc explaining this is extensions-layout only.
- **`hasLegacyNpmResidue()`** — Now derives root from `getConfigFilePathSafe()` instead of hardcoded `homedir()` — supports custom OPENCLAW_HOME.

### Correctness gaps fixed ✅

1. npm-installed legacy plugins (`LEGACY_PLUGIN_ID` under `npm/node_modules/`) now detected via `plugins list --json` — `isHealthyInstall` directory probe can no longer miss them.
2. Runtime registrations not yet persisted to cfg are caught via live list signal.
3. Custom OpenClaw root (OPENCLAW_HOME, Windows) correctly handled for residue check.

### Tests ✅

- `listLoadedPlugins`: 6 tests covering happy path, noise stripping, old runtime, non-JSON, bad shape, missing-id filtering.
- `detectInstallState` with `plugins list --json` signal: 3 tests covering legacyNpmActive, legacyDmworkResidue via LEGACY_PLUGIN_ID, and via VERY_LEGACY_PLUGIN_ID through the list path.
- Existing tests continue exercising the cfg.entries fallback path (default `pluginsListSupported: false`).

### CI

- `Build & Type-check` ✅ pass

No blocking issues. Ship it.

Jerry-Xin (changes requested):
The PR is relevant to `Mininglamp-OSS/octo-adapters`, but I found one blocking correctness regression in install-state classification.

**🔴 Blocking**
- 🔴 Critical: `isHealthyInstall()` now treats `extensions/<id>` plus `cfg.plugins.entries[id]` as healthy even when `cfg.plugins.installs[id]` is missing and `plugins inspect` is unsupported. This conflicts with `resolvePluginState()` and `detectScenario()`, which still treat that same old-runtime state as partial/broken. As a result, pre-flight callers can allow `bind`, `quickstart`, or `remove-account` to proceed against an install that `install` itself would classify as broken and repair. The loosened check is at `create-openclaw-octo/cli/openclaw-cli.ts:1069-1086`; the stricter scenario check remains at `create-openclaw-octo/cli/openclaw-cli.ts:1053-1061`. Please either keep `isHealthyInstall()` aligned with the existing healthy definition, or update all related classifiers and add regression coverage for the `dir + entries + no installs + inspect unsupported` case.

**💬 Non-blocking**
- 🟡 Warning: The npm legacy fallback comment says old runtimes require `cfg.plugins.entries[id]` plus the npm-layout install directory, but the code only checks `isPluginRegistered()` at `create-openclaw-octo/cli/openclaw-cli.ts:1188-1200`. That can classify a stale config-only `openclaw-channel-octo` entry as `octo-npm-legacy` instead of `broken`. If that behavior is intentional, the comment and tests should be updated; otherwise, restore the residue/path check.

**✅ Highlights**
- The PR correctly moves modern dual-active detection toward `plugins list --json`, which is a better authority than probing `extensions/<id>`.
- Added tests cover JSON parsing, unsupported runtimes, and list-driven legacy detection paths.
- I ran `npm test`, `npm run type-check`, and `npm run build`; all passed locally after installing dependencies.

yujiawei (dismissed):
# Code Review — PR #83 (octo-adapters)

## Summary

This PR consolidates the install-state detection layer in `create-openclaw-octo/cli/openclaw-cli.ts` behind a new `openclaw plugins list --json` enumeration helper. It addresses three correctness gaps documented in #82:

1. `cfg.plugins.installs` is observed null on current OpenClaw schemas, so detection that AND'd against it was effectively dead.
2. The `extensions/<id>` directory probe missed npm-installed legacy plugins (`openclaw-channel-dmwork`, `openclaw-channel-octo`) which live under `npm/node_modules/<id>`.
3. `hasLegacyNpmResidue()` hardcoded `homedir() + .openclaw/...`, missing custom `OPENCLAW_HOME`, npm `--prefix`, and Windows non-default layouts.

The implementation is small (one new public helper `listLoadedPlugins()`, one private helper `isPluginRegistered()`, refactored `detectInstallState()` and `hasLegacyNpmResidue()`, simplified `isHealthyInstall()` fallback). Backward-compatibility is preserved via a clean `supported: false` shape when the runtime does not recognise `plugins list --json`.

## 1. Verification

- ✅ **Build / type-check / tests** — `npm run build`, `npx tsc --noEmit`, and `vitest run` all pass locally on `8dc65d9`. 111/111 tests pass (102 prior + 9 new).
- ✅ **`listLoadedPlugins()` parsing** (`cli/openclaw-cli.ts:531-564`) — wraps `runOpenclaw(["plugins", "list", "--json"])`, runs through `stripStdoutNoise`, finds JSON start, parses, and validates `Array.isArray(data?.plugins)` before mapping. Each of the four failure modes (throw, no `{`, parse error, missing `plugins` array) returns the correct `{ supported: false, plugins: [] }`. Defensive field typing (`typeof p?.id === "string"` etc.) and the empty-id filter are correct.
- ✅ **`isPluginRegistered` source-of-truth selection** (`cli/openclaw-cli.ts:1144-1153`) — when `list.supported`, the live runtime list is authoritative; otherwise falls back to `cfg.plugins.entries[id]`. This is the right priority: `plugins list --json` is independent of cfg schema and client-side path assumptions.
- ✅ **`hasLegacyNpmResidue()` root derivation** (`cli/openclaw-cli.ts:1119-1126`) — uses `getConfigFilePathSafe().replace(/openclaw\.json$/, "")` instead of `homedir() + ".openclaw/..."`. This correctly picks up custom `OPENCLAW_HOME`, npm `--prefix`, and Windows layouts. Comment correctly notes this is cosmetic-only (drives the warn line, never the block).
- ✅ **`detectInstallState()` priority chain** (`cli/openclaw-cli.ts:1155-1248`) — single `cfg = readConfigFromFile()` and `list = listLoadedPlugins()` at the top, threaded through all helpers. `octo-clawhub` arm now OR's the list/cfg signal with the existing channel/binding artefact checks for `legacyDmworkResidue`. `dmwork-legacy` branches now OR `isHealthyInstall` with `isPluginRegistered` so npm-installed dmwork (which lives under `npm/node_modules/<id>`) is caught.
- ✅ **`isHealthyInstall()` fallback simplification** (`cli/openclaw-cli.ts:1069-1087`) — drops the dead `hasInstalls` term. The retained `inspectOk` short-circuit means modern OpenClaw is unaffected; the cfg fallback now requires `dir + entries`, which is the documented "extensions-layout only" contract.
- ✅ **Test coverage** — the three new `detectInstallState` cases isolate the list signal by leaving `cfg.plugins.entries` empty for the legacy id and asserting the state is still detected via the list payload. The six `listLoadedPlugins` cases cover the happy path, banner-noise stripping, throw-as-unsupported, non-JSON output, missing `plugins` array, and empty/missing-id filtering.
- ✅ **No mention markers, no leaked agent identities** — n/a, this is a refactor PR; the diff itself does not introduce any.

## 2. Findings

### P0 — none.

### P1 — none.

### P2 / observations (non-blocking)

#### P2-1. `octo-npm-legacy` fallback gate has been loosened — old: `entries && installs && dir-exists`, new: `entries` (cfg path)

**Location:** `cli/openclaw-cli.ts:1196-1201`.

The pre-PR fallback for old OpenClaw runtimes required all three: `cfg.plugins.entries[NPM_PACKAGE_NAME]` AND `cfg.plugins.installs[NPM_PACKAGE_NAME]` AND `hasLegacyNpmResidue()` (npm dir on disk). The new code in the cfg-fallback path requires only `cfg.plugins.entries[NPM_PACKAGE_NAME]`.

Practical effect: a stale `entries` registration with no install dir on disk (and no `installs` metadata) previously fell through to the `broken` classifier (line 1241 leftover detection); it now classifies as `octo-npm-legacy`. Both states block the pre-flight, but the surfaced upgrade notice differs (`utils.ts:286-294` vs the broken/doctor path).

This is consistent with the PR's overall philosophy — `plugins list --json` reporting an id is authoritative on modern OpenClaw, and `cfg.entries` is the best-available proxy on old OpenClaw — and routing the user toward "please upgrade via `npx -y create-openclaw-octo install`" is a reasonable outcome for either case. Worth calling out in the PR description so reviewers / future debuggers know the classifier boundary moved. Not a blocker.

#### P2-2. No test for "list supported but disagrees with cfg.entries"

The new tests cover the case where list reports an id and `cfg.entries` does NOT (the "list is authoritative when supported" assertion). The reverse case — `pluginsListSupported: true`, `legacyNpmInList: false`, `legacyNpmInEntries: true` — is not exercised. Per `isPluginRegistered`, that returns `false` (the list signal wins), and any stale `cfg.entries` left over from a previous install is intentionally ignored. A small additional test pinning that behaviour would lock down the contract. Not a blocker.

#### P2-3. Per-call subprocess cost

`detectInstallState()` now invokes `openclaw plugins list --json` in addition to the existing `openclaw plugins inspect octo --json` (via `isHealthyInstall`) and `openclaw config file` (via `getConfigFilePathSafe`). On hosts where `runOpenclaw` resolves to the global binary that's a third spawn per pre-flight. Not on the hot path, but if pre-flight latency matters in interactive flows you may want to memoise `listLoadedPlugins` for the duration of one CLI command. Not a blocker.

#### P2-4. Silent degradation when `plugins list --json` output is malformed

`listLoadedPlugins()` returns `supported: false` for four distinct conditions (throw, no `{`, parse error, missing `plugins` array). On modern OpenClaw, only "throw / unknown command" is a legitimate `supported: false`; the other three indicate a real bug or transient corruption that gets silently downgraded to the cfg fallback. A `console.debug`-level breadcrumb would help debug field reports. Not a blocker.

#### P2-5. Carry-forwards from PR #81 still pending

The PR description correctly notes the out-of-scope items from yujiawei's #81 review: direct unit tests for `detectInstallState`'s broken-state classifier, `compareVersions` / `compareOpenClawVersion` consolidation, and exhaustive-switch hardening. Tracking them on a follow-up is fine; just want to keep the carry-forward visible.

### Nits

- `cli/openclaw-cli.ts:1199` — the `octo-npm-legacy` arm reads `cfg?.plugins?.installs?.[NPM_PACKAGE_NAME]?.version` for the version field, but the comment block above (line 1188-1195) and the `isHealthyInstall` block (line 1072-1075) explicitly note that `cfg.plugins.installs` has been observed null on current schemas. The version will frequently be `null` in practice. Defensive fallback is fine, but the comment-vs-code juxtaposition reads as a small contradiction; either drop the version lookup and unconditionally `null` it, or update the comment to acknowledge it's a "best-effort, often null" read.

## 3. Recommendations

1. (Optional) Add the inverse test in P2-2 to lock down the "list is authoritative" contract.
2. (Optional) Add a `console.debug` breadcrumb in the malformed-JSON / missing-array branches of `listLoadedPlugins` for field debuggability (P2-4).
3. (Optional) Memoise `listLoadedPlugins()` for the duration of a single pre-flight to avoid the second spawn (P2-3).
4. None of the above is required for merge.

## 4. Additional observations

- `LEGACY_NPM_RESIDUE_PATH` → `LEGACY_NPM_RESIDUE_DIR` rename is a meaningful semantic shift (path-from-homedir → dir-relative-to-config-root). Worth a one-line note in the changelog if one is being maintained.
- The test mock at `openclaw-cli.test.ts:868` still asserts `path.endsWith(".openclaw/npm/node_modules/openclaw-channel-octo")` for the residue check. With the new root derivation that suffix still matches because the mocked `config file` returns `/home/user/.openclaw/openclaw.json`. If a future change moved the test config path away from a `.openclaw` parent, this matcher would silently false-negative. Low priority.
- `isHealthyInstall(NPM_PACKAGE_NAME)` is still called at line 1182 before the `isPluginRegistered` fallback. On modern OpenClaw that's correct (inspect succeeds → return early). On old OpenClaw without `plugins inspect`, the cfg-fallback inside `isHealthyInstall` requires `extensions/<id>` to exist — which it never does for npm-layout `openclaw-channel-octo` — so this branch is effectively a no-op on old runtimes and the `isPluginRegistered` line below is what catches it. The current code is correct; just noting the path so a future cleanup doesn't accidentally collapse the two branches without realising the layout asymmetry.

Verdict: no P0/P1 blockers. Refactor is well-scoped, well-tested, and addresses the documented gaps in #82 cleanly.

caster-Q:
Pushed 3 fix commits addressing @Jerry-Xin's CHANGES_REQUESTED + 3 rounds of codex review (final round: Approve and merge).

### `3888d24` — @Jerry-Xin BLOCKER + warning

- **BLOCKER (isHealthyInstall fallback inconsistency)**: restored the 3-artifact fallback `(hasDir + hasEntries + hasInstalls)` to align with `detectScenario()` and `cleanupBrokenInstall()`. The previous loosening to 2-artifact created the inconsistency Jerry-Xin called out. Tightening across all classifiers stays as a follow-up to #82.
- **Non-blocking warning (npm legacy belt-and-suspenders)**: split the path into modern OpenClaw (list signal authoritative) vs old OpenClaw (cfg.entries AND npm-layout dir). Prevents stale cfg-only entries being misclassified as `octo-npm-legacy` instead of `broken`. Added a regression test.

### `7eb060a` — codex round-1 MAJOR (3 findings)

- **dmwork legacy priority inverted**: swapped the order so `VERY_LEGACY_PLUGIN_ID` is checked before `LEGACY_PLUGIN_ID`, mirroring `detectScenario()`'s priority-1 vs priority-2 buckets. Added a regression test for "both ids registered → must report very-legacy version" (codex specifically asked for this).
- **listLoadedPlugins JSON parse fragile to trailing noise**: replaced `out.slice(jsonStart)` + `JSON.parse` with a `extractJsonObject` brace-counter that respects string boundaries (handles `"path/with/{braces}"` correctly).
- **listLoadedPlugins conflated old runtime with real error**: matched `pluginsInspectDetailed`'s pattern and split the catch by `unknown command|unrecognized command` regex.

### `b15fea3` — codex round-2 MAJOR (1 finding) — fail-closed completion

The round-1 fix only identified the distinction; the actual fail-closed behaviour was missing — `detectInstallState` was still trusting cfg.entries on real errors. Completed in this commit:

- `PluginListResult` gains `error: string | null`. null = success or "old runtime"; non-null = real failure (config corrupt, permission, plugin load crash, partial JSON).
- `detectInstallState` now fail-closes: `!list.supported && list.error !== null` → returns `{ kind: "broken", details: "cannot determine install state: ..." }`. Users get the same upgrade-prompt path as other broken states, telling them to repair OpenClaw before further config writes.
- Regression test: EACCES on `plugins list --json` even with healthy-looking cfg/dir → must classify as `broken`.

Codex round 3: **\`Approve and merge\`** ([transcript](attachable))

### Verification

- **119/119 tests passing** (102 prior + 6 listLoadedPlugins + 4 listLoadedPlugins error variants + 4 detectInstallState list-signal cases + 1 dmwork priority + 1 broken stale + 1 fail-closed)
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- **Real-machine smoke test on macOS arm64** (OpenClaw 2026.5.18):
  - clean octo-clawhub: `listLoadedPlugins()` reports 92 plugins, `detectInstallState()` returns `octo-clawhub` with all flags false
  - dual-active simulation (force-installed `openclaw-channel-octo@1.0.0`): `legacyNpmActive: true`, `npmResidue: true`, doctor banner + bind block confirmed
  - very-legacy dmwork residue (cfg.entries injection): `legacyDmworkResidue: true`, doctor banner + bind block confirmed

@Jerry-Xin @yujiawei @lml2468 — please re-review.

lml2468 (approved):
REVIEW_STATE: APPROVED

## Re-review (head: b15fea3cb1af, previous: APPROVED @ 8dc65d92)

### Jerry-Xin's concerns resolved ✅

**P0 — `isHealthyInstall()` consistency**: Restored `cfg.plugins.installs[id]` check. Now `hasDir && hasEntries && hasInstalls` on the fallback path — aligned with `detectScenario()` / `cleanupBrokenInstall()`. The `inspect` short-circuit covers modern runtimes where `installs` is null. Well-documented tradeoff comment + follow-up tracked.

**P1 — Stale config-only entry misclassification**: npm legacy fallback now requires BOTH `cfg.plugins.entries[id]` AND `hasLegacyNpmResidue()` (actual directory existence). Stale entry without dir → `broken`, not `octo-npm-legacy`. Regression test confirms.

### Additional hardening ✅

1. **`listLoadedPlugins().error` field** — Distinguishes "old runtime" (`error: null`) from "real failure" (`error: <msg>`). Callers can fail-closed on the latter.
2. **`detectInstallState()` fail-closed path** — When `plugins list --json` fails for a real reason (EACCES, config corruption, etc.), returns `broken` immediately. Prevents trusting cfg.entries when the runtime state is unreliable.
3. **`extractJsonObject()`** — Proper brace-matching JSON extractor. Handles trailing log noise and string-embedded `{}`/`}` that broke naive `JSON.parse(out.slice(start))`.
4. **dmwork priority order** — `VERY_LEGACY_PLUGIN_ID` checked before `LEGACY_PLUGIN_ID` in the standalone dmwork-legacy classification path (matches `detectScenario()` priority).

### New tests (+214 lines) ✅

- Trailing log noise after JSON (brace-matching)
- Strings containing curly braces
- `unknown command` → unsupported (error: null)
- Real error (EACCES) → error field populated
- Parse failure → error field populated
- Dual dmwork ids registered → VERY_LEGACY version surfaced (priority alignment)
- Stale config-only openclaw-channel-octo → `broken` (not `octo-npm-legacy`)
- `plugins list --json` EACCES → `broken` (fail-closed)

### CI

- `Build & Type-check` ✅ pass (17s)
- `check-sprint` ✅ pass

Comprehensive fix. All classifier paths now consistent. No new issues.

yujiawei (approved):
# Code Review — PR #83 (octo-adapters)

## Verdict: APPROVED

## Verification Summary

| Check | Result |
|---|---|
| `npm run test` | ✅ 119/119 passing locally (vitest) |
| `npx tsc --noEmit` | ✅ clean |
| Closes #82 (3 stated gaps) | ✅ all three addressed (custom OpenClaw root, npm-layout legacy, dual-active on stale cfg) |
| Backward compatibility on old OpenClaw (no `plugins list --json`) | ✅ `supported: false, error: null` → cfg.entries fallback |
| Fail-closed on real `plugins list --json` errors | ✅ `supported: false, error: <msg>` → `{ kind: "broken" }` |
| dmwork-legacy priority mirrors `detectScenario()` | ✅ VERY_LEGACY checked before LEGACY |
| Trailing-noise tolerance in JSON parsing | ✅ `extractJsonObject` brace-counter respects string boundaries |
| Stale `cfg.entries.openclaw-channel-octo` (no dir) on old runtime | ✅ classified as `broken`, not `octo-npm-legacy` |
| `hasLegacyNpmResidue()` honours custom OpenClaw root | ✅ derived from `getConfigFilePathSafe()` |

## What this PR does well

- **Right primitive.** `openclaw plugins list --json` is OpenClaw's own answer to "what's loaded?", and adopting it eliminates the indirect cfg+dir+inspect combination that grew correctness gaps in #82. The tri-state result (`supported`, `error`, `plugins`) is the right shape: it lets the caller distinguish "old runtime → fall back" from "real failure → fail closed", which is exactly the distinction the previous design was missing.
- **Layered fallback is principled.** `isPluginRegistered(id, list, cfg)` cleanly encodes "list is authoritative when supported, else cfg.entries". The npm-legacy belt-and-suspenders branch (`openclaw-cli.ts:1303-1313`) explicitly splits modern (list-based) vs old (cfg.entries AND `hasLegacyNpmResidue()`) paths, and the requirement that the cfg-fallback path AND the npm-layout dir actually exist prevents the stale-cfg → false-`octo-npm-legacy` misclassification (regression test at `openclaw-cli.test.ts:1185-1223`).
- **Fail-closed on real errors is the right call.** EACCES / config corruption / partial JSON are exactly the states where trusting cfg.entries alone could let a dual-active state slip through bind/quickstart. Surfacing `broken` with the actual error in `details` (`openclaw-cli.ts:1257-1262`) gives the user a clear next step. Test at `openclaw-cli.test.ts:1225-1266` pins the behaviour.
- **`extractJsonObject` is correctly written.** The string-boundary tracking (`escaped` flag + `inString`) handles `"path/with/{braces}"` correctly — confirmed by the regression test at `openclaw-cli.test.ts:160-174`. Tolerates trailing log noise that survived `stripStdoutNoise`.
- **dmwork priority alignment.** Reordering VERY_LEGACY before LEGACY mirrors `detectScenario()`'s priority-1 vs priority-2 buckets, so `detectInstallState`'s reported version aligns with the migration `install.ts` will actually run when both ids are registered. Test at `openclaw-cli.test.ts:1136-1183`.
- **Comments earn their keep.** The block at `openclaw-cli.ts:491-514` explains why the new helper exists, and the comment at `openclaw-cli.ts:1156-1171` captures the deliberate (acknowledged) inconsistency between `isHealthyInstall`'s 3-artifact fallback and `detectInstallState`'s 2-artifact derivation, with a tracked follow-up. That kind of "we know this is dead code on modern OpenClaw, here's why we left it" annotation is the right level of detail for a reviewer six months from now.

## Findings

All P2 / suggestions — none block merge.

### P2-1 — `listLoadedPlugins` does not classify `unknown option '--json'` as unsupported

`openclaw-cli.ts:600`:

```ts
if (/unknown command|unrecognized command/i.test(text)) {
  return { supported: false, error: null, plugins: [] };
}
```

On a hypothetical OpenClaw runtime that supports `plugins list` but not the `--json` flag, the error text would be `unknown option '--json'` (commander.js convention), which the regex doesn't match. That runtime would be classified as a real error → `broken`, not `unsupported` → cfg.entries fallback.

This is theoretical given the PR description's stated assumption ("`plugins list --json` should be available on any OpenClaw version that ships ClawHub"), but matches the pattern used by `pluginsInstall`'s `isUnsupportedOptionError` (`openclaw-cli.ts:293-303`), which already extends to `unknown option`. Worth aligning here for symmetry:

```ts
if (/unknown command|unrecognized command|unknown option|unrecognized option/i.test(text)) {
  return { supported: false, error: null, plugins: [] };
}
```

Optional — file as a small follow-up if you'd rather keep this PR's scope tight.

### P2-2 — npm-legacy version sourced from `cfg.plugins.installs` on the modern path

`openclaw-cli.ts:1309-1313`:

```ts
if (npmLegacyActiveViaList || npmLegacyActiveViaCfgFallback) {
  return {
    kind: "octo-npm-legacy",
    version: cfg?.plugins?.installs?.[NPM_PACKAGE_NAME]?.version ?? null,
  };
}
```

When the modern path (`npmLegacyActiveViaList`) fires, `cfg.plugins.installs` is the field this PR's own preamble flags as "observed null on current OpenClaw schemas" (`openclaw-cli.ts:500-501`). So on modern OpenClaw the version will frequently fall through to `null`, and the `utils.ts:289` user-facing message degrades to `"openclaw-channel-octo 1.0.x"` placeholder.

The earlier branch at `openclaw-cli.ts:1287-1291` correctly uses `resolvePluginState(NPM_PACKAGE_NAME).version` (which prefers `plugins inspect`). Mirroring that here would give an accurate version on the belt-and-suspenders branch too:

```ts
version: resolvePluginState(NPM_PACKAGE_NAME).version
       ?? cfg?.plugins?.installs?.[NPM_PACKAGE_NAME]?.version
       ?? null,
```

Same pattern is already used in the dmwork branches at `openclaw-cli.ts:1334-1336` and `1346-1348`. Cosmetic — the user-facing prompt is fine either way — but worth a one-liner for consistency.

### P2-3 — `isHealthyInstall(NPM_PACKAGE_NAME)` extensions-only probe is a known mismatch (acknowledged)

`openclaw-cli.ts:1287` calls `isHealthyInstall(NPM_PACKAGE_NAME)`, but the fallback inside `isHealthyInstall` only probes `<root>/extensions/<id>` — npm-legacy lives at `<root>/npm/node_modules/<id>`. So on old OpenClaw without `plugins inspect`, this branch silently returns false for a healthy npm install.

The PR addresses this via the belt-and-suspenders branch immediately below, which the `isPluginRegistered` + `hasLegacyNpmResidue()` combination correctly catches. The comment at `openclaw-cli.ts:1167-1171` calls this out explicitly. So it's working as designed — flagging only because the layering depends on a reader noticing both blocks.

A future cleanup that consolidates `isHealthyInstall` to be id-aware (extensions-layout for ClawHub, npm-layout for legacy) would simplify this; the PR description's "out-of-scope cleanup" follow-up to #82 is the right home for it.

### P2-4 — Test artificiality (pre-existing, not introduced by this PR)

`openclaw-cli.test.ts:1095-1134` (`octo-npm-legacy when ClawHub octo absent and NPM_PACKAGE_NAME healthy`) sets `existsSync` to true for `/extensions/openclaw-channel-octo`, but in production npm-installed legacy plugins live at `/npm/node_modules/<id>`. The test relies on `isHealthyInstall(NPM_PACKAGE_NAME)` succeeding via this fake-extensions-dir path because `pluginsInspect` is mocked to throw. The test still verifies the right return shape, but it doesn't exercise the realistic old-OpenClaw + npm-legacy code path through the belt-and-suspenders fallback.

Not a blocker — the new test at `openclaw-cli.test.ts:1185-1223` covers the realistic stale-cfg case, and the dual-active list-signal test at `:1044-1059` covers modern-OpenClaw npm-legacy. Just noting that this older test could be tightened in a follow-up to fully drop the fake `/extensions/openclaw-channel-octo` existsSync.

## Non-findings

A few things I checked and confirmed are fine:

- `extractJsonObject` correctness on edge cases (escaped quotes inside strings, `{` / `}` in string values) — verified by the regression test and by re-reading the implementation.
- `getConfigFilePathSafe()` always ending in `openclaw.json` so the `.replace(/openclaw\.json$/, "")` strip is reliable — yes, verified at `openclaw-cli.ts:863-869` (fallback hardcodes the suffix).
- `InstallState` discriminated union — `legacyDmworkResidue` and `legacyNpmActive` correctly only on the `octo-clawhub` arm.
- Caller site at `utils.ts:244-323` (`assertPluginReady`) — exhaustively handles every `InstallState.kind` including `broken`. The new fail-closed `broken` from list errors plugs straight into the existing handler with a sensible block message.
- TypeScript discriminated union exhaustiveness — `switch (plugin.kind)` in `utils.ts` covers all 5 variants; no missing case.

## Recommendation

Approve and merge. The three P2 nits above are all polish suggestions that can land in a follow-up; none of them affect correctness on the supported runtime range. The test suite is comprehensive (16 tests touching the new code paths), the prior review rounds (Jerry-Xin BLOCKER + 3 codex rounds) have already tightened the rough edges, and the real-machine smoke test in the author's comment exercises the three behavioural improvements end-to-end on macOS / OpenClaw 2026.5.18.

Jerry-Xin (approved):
The PR is in scope and the new `plugins list --json` detection path is well covered; I did not find any blocking correctness, security, or architecture issues.

💬 Non-blocking:
- 🟡 Warning: `isPluginRegistered()` treats any id in `plugins list --json` as active, ignoring `enabled` and `status` (`create-openclaw-octo/cli/openclaw-cli.ts:1240`). If OpenClaw’s list output can include disabled or failed plugins, this may produce false-positive legacy blocks. If the command contract is “currently trying to load,” this is fine; otherwise filter by the relevant active state and add a disabled-entry test.
- 🔵 Suggestion: The custom-root residue probe now uses the live config root, but the warning text still hardcodes `~/.openclaw/npm/node_modules/openclaw-channel-octo` (`create-openclaw-octo/cli/utils.ts:281`, `create-openclaw-octo/cli/utils.ts:354`). Consider rendering the detected path or using less path-specific wording.
- 🔵 Suggestion: Some added test comments include reviewer/process provenance (`create-openclaw-octo/cli/openclaw-cli.test.ts:1137`, `create-openclaw-octo/cli/openclaw-cli.test.ts:1226`). I’d make them plain regression comments before merge.

✅ Highlights:
- Good fail-closed handling for real `plugins list --json` failures.
- Good regression coverage for stale cfg fallback, list-sourced legacy detection, trailing stdout noise, and dmwork id priority.

Verification note: I could not run the test suite locally because dependencies are not installed in this checkout (`vitest: command not found`).

PR #84, opened 2026-05-27
ci: extend Build & Type-check matrix to macOS + Windows

caster-Q:
## Summary

CI's `build` job currently only runs on `ubuntu-latest`. The `create-openclaw-octo` CLI is published as a global npm package and is explicitly cross-platform (Windows path separators, `.cmd` shims, CRLF stdout, custom `OPENCLAW_HOME` layouts, etc.), but none of that is exercised in CI today.

This PR expands the existing matrix to cover all three GitHub-hosted runners:

\`\`\`yaml
os: [ubuntu-latest, macos-latest, windows-latest]
\`\`\`

## Related Issue

N/A — preventive CI hardening, not tied to a specific bug. Complements the manual Windows real-machine smoke test being run in parallel for the just-merged PR #83 detection-layer refactor.

## Changes

- `.github/workflows/ci.yml`: add `os` axis to the build job matrix; switch `runs-on: ubuntu-latest` → `runs-on: \${{ matrix.os }}`; update job name to include the OS so each runner gets its own PR check entry; add `fail-fast: false` so all three platforms report in a single push rather than one failure cancelling the others.

## Effect

Per-PR check list goes from:

\`\`\`
Build & Type-check (create-openclaw-octo)
\`\`\`

To:

\`\`\`
Build & Type-check (create-openclaw-octo, ubuntu-latest)
Build & Type-check (create-openclaw-octo, macos-latest)
Build & Type-check (create-openclaw-octo, windows-latest)
\`\`\`

Wall-clock per PR: ~30s → ~60-90s (windows-latest is the slowest, runs in parallel with the other two). All three runners are free for public repos, so the change has no $ cost.

## Testing

- [x] Local actionlint not installed; relying on the existing `sanity / actionlint` workflow for syntax validation.
- [x] yml diff hand-verified — all three runners receive the same `working-directory: \${{ matrix.package }}` resolution, same `npm ci` / type-check / build / test sequence.
- [x] No code changes — purely workflow infrastructure.

## Checklist

- [x] Added tests for my changes — N/A (CI infra change, not source)
- [x] Updated documentation — N/A (no user-facing surface change)

## Rationale: why catch this in CI vs trust the runtime?

Auto-CI catches code-level OS bugs that don't need OpenClaw runtime to surface (path separator handling, Buffer encoding, fs API quirks, regex `\\r\\n` matching, dynamic-import URL formatting). Manual real-machine testing catches integration-level bugs (OpenClaw runtime stdout format, `.cmd` spawn behaviour, plugin-list detection on real cfg). Together they cover the surface the ubuntu-only matrix missed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,10 +53,15 @@ jobs:
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
       needs.changes.outputs.code == 'true'
-    name: Build & Type-check (${{ matrix.package }})
-    runs-on: ubuntu-latest
+    name: Build & Type-check (${{ matrix.package }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     strategy:
+      # Don't cancel other OS jobs when one fails — we want to see all
+      # platform-specific failures in a single PR check, not have to push
+      # again to surface a Windows error after the macOS error is fixed.
+      fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
         include:
           - package: create-openclaw-octo
             install_cmd: npm ci
--- a/create-openclaw-octo/cli/__test_utils__/path.ts
+++ b/create-openclaw-octo/cli/__test_utils__/path.ts
@@ -1,24 +1,37 @@
 import path from "node:path";
+import { homedir } from "node:os";
 
 /**
  * The POSIX-style fake config file path that test mocks return for
  * `runOpenclaw(["config", "file"])`. Use this in mock implementations
  * (the value is fed into production code, which then normalises it via
- * `getConfigFilePathSafe` → `path.resolve`).
+ * `getConfigFilePathSafe` → `normalizeConfigPath`).
  */
 export const FAKE_CFG_PATH = "/home/user/.openclaw/openclaw.json";
 
 /**
- * The OS-resolved form of FAKE_CFG_PATH after `path.resolve`. Use this
- * when asserting against arguments passed to spies (copyFileSync,
- * writeFileSync, etc.) — production code normalises every incoming
- * config path via `getConfigFilePathSafe`, so spies see the resolved
- * form, not the POSIX input.
+ * The OS-resolved form of FAKE_CFG_PATH after `normalizeConfigPath`.
+ * Use this when asserting against arguments passed to spies
+ * (copyFileSync, writeFileSync, etc.) — production code normalises
+ * every incoming config path, so spies see the resolved form, not the
+ * POSIX input.
  *
- *   Linux/macOS:  /home/user/.openclaw/openclaw.json
- *   Windows:      C:\home\user\.openclaw\openclaw.json
+ * Mirrors `cli/openclaw-cli.ts:170-180 normalizeConfigPath`:
+ *   - if `path.resolve(p) === p` it's already absolute → return as-is
+ *   - else resolve relative to `os.homedir()` (the production fallback)
+ *
+ * On POSIX (Linux/macOS) this returns the input unchanged. On Windows
+ * `path.resolve("/home/user/...")` does NOT equal the input (Node
+ * prepends a drive letter) so the homedir branch fires — yielding e.g.
+ * `C:\home\user\.openclaw\openclaw.json` on a runner whose homedir is
+ * on the C: drive, regardless of the cwd drive. This is critical on
+ * GitHub Actions Windows runners where cwd is on D: but homedir is
+ * on C:.
  */
-export const RESOLVED_CFG_PATH = path.resolve(FAKE_CFG_PATH);
+export const RESOLVED_CFG_PATH = (() => {
+  if (path.resolve(FAKE_CFG_PATH) === FAKE_CFG_PATH) return FAKE_CFG_PATH;
+  return path.resolve(homedir(), FAKE_CFG_PATH);
+})();
 
 /**
  * Path-separator-agnostic suffix matcher. Use in `existsSync` / `rmSync`
--- a/create-openclaw-octo/cli/__test_utils__/path.ts
+++ b/create-openclaw-octo/cli/__test_utils__/path.ts
@@ -1,0 +1,48 @@
+import path from "node:path";
+
+/**
+ * The POSIX-style fake config file path that test mocks return for
+ * `runOpenclaw(["config", "file"])`. Use this in mock implementations
+ * (the value is fed into production code, which then normalises it via
+ * `getConfigFilePathSafe` → `path.resolve`).
+ */
+export const FAKE_CFG_PATH = "/home/user/.openclaw/openclaw.json";
+
+/**
+ * The OS-resolved form of FAKE_CFG_PATH after `path.resolve`. Use this
+ * when asserting against arguments passed to spies (copyFileSync,
+ * writeFileSync, etc.) — production code normalises every incoming
+ * config path via `getConfigFilePathSafe`, so spies see the resolved
+ * form, not the POSIX input.
+ *
+ *   Linux/macOS:  /home/user/.openclaw/openclaw.json
+ *   Windows:      C:\home\user\.openclaw\openclaw.json
+ */
+export const RESOLVED_CFG_PATH = path.resolve(FAKE_CFG_PATH);
+
+/**
+ * Path-separator-agnostic suffix matcher. Use in `existsSync` / `rmSync`
+ * mocks instead of `String(p).endsWith("/some/suffix")` — `path.resolve`
+ * upstream produces native separators, so a literal POSIX `endsWith`
+ * silently misses on Windows.
+ */
+export function pathEndsWith(p: unknown, suffix: string): boolean {
+  return String(p).replace(/\\/g, "/").endsWith(suffix);
+}
+
+/**
+ * Path-separator-agnostic substring check. Counterpart of
+ * `pathEndsWith` for `String(p).includes("/some/segment/")` patterns.
+ */
+export function pathIncludes(p: unknown, segment: string): boolean {
+  return String(p).replace(/\\/g, "/").includes(segment);
+}
+
+/**
+ * Path-separator-agnostic regex match. Returns the first match (with
+ * any capture groups) or null. Use when extracting a path segment from
+ * native-separator paths via a POSIX-style regex.
+ */
+export function pathMatch(p: unknown, regex: RegExp): RegExpMatchArray | null {
+  return String(p).replace(/\\/g, "/").match(regex);
+}
--- a/create-openclaw-octo/cli/config-preserve.test.ts
+++ b/create-openclaw-octo/cli/config-preserve.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { readFileSync, writeFileSync, copyFileSync, renameSync } from "node:fs";
 import { execFileSync } from "node:child_process";
+import { FAKE_CFG_PATH, RESOLVED_CFG_PATH } from "./__test_utils__/path.js";
 
 vi.mock("node:child_process", () => ({
   execFileSync: vi.fn(),
@@ -43,7 +44,7 @@ describe("saveChannelConfigFromFile", () => {
 
   it("should read channels.octo with real secrets from file", async () => {
     const { saveChannelConfigFromFile } = await loadModule();
-    mockExecFileSync.mockReturnValue("/home/user/.openclaw/openclaw.json");
+    mockExecFileSync.mockReturnValue(FAKE_CFG_PATH);
     mockReadFileSync.mockReturnValue(JSON.stringify(SAMPLE_CONFIG));
 
     const saved = saveChannelConfigFromFile();
@@ -55,7 +56,7 @@ describe("saveChannelConfigFromFile", () => {
 
   it("should return null when channels.octo does not exist", async () => {
     const { saveChannelConfigFromFile } = await loadModule();
-    mockExecFileSync.mockReturnValue("/home/user/.openclaw/openclaw.json");
+    mockExecFileSync.mockReturnValue(FAKE_CFG_PATH);
     mockReadFileSync.mockReturnValue(JSON.stringify({ channels: {} }));
 
     expect(saveChannelConfigFromFile()).toBeNull();
@@ -63,7 +64,7 @@ describe("saveChannelConfigFromFile", () => {
 
   it("should return null when file read fails", async () => {
     const { saveChannelConfigFromFile } = await loadModule();
-    mockExecFileSync.mockReturnValue("/home/user/.openclaw/openclaw.json");
+    mockExecFileSync.mockReturnValue(FAKE_CFG_PATH);
     mockReadFileSync.mockImplementation(() => { throw new Error("ENOENT"); });
 
     expect(saveChannelConfigFromFile()).toBeNull();
@@ -77,7 +78,7 @@ describe("restoreChannelConfigToFile", () => {
 
   it("should write channels.octo back to file with backup", async () => {
     const { restoreChannelConfigToFile } = await loadModule();
-    mockExecFileSync.mockReturnValue("/home/user/.openclaw/openclaw.json");
+    mockExecFileSync.mockReturnValue(FAKE_CFG_PATH);
     mockReadFileSync.mockReturnValue(JSON.stringify({ channels: {}, plugins: {} }));
 
     const dmworkConfig = SAMPLE_CONFIG.channels.octo;
@@ -85,8 +86,8 @@ describe("restoreChannelConfigToFile", () => {
 
     // Should create backup
     expect(mockCopyFileSync).toHaveBeenCalledWith(
-      "/home/user/.openclaw/openclaw.json",
-      "/home/user/.openclaw/openclaw.json.bak",
+      RESOLVED_CFG_PATH,
+      `${RESOLVED_CFG_PATH}.bak`,
     );
 
     // Should write merged config
@@ -106,7 +107,7 @@ describe("config preserve on install --force failure", () => {
     mockExecFileSync.mockImplementation((cmd, args) => {
       const argsArr = args as string[];
       if (argsArr[0] === "config" && argsArr[1] === "file") {
-        return "/home/user/.openclaw/openclaw.json";
+        return FAKE_CFG_PATH;
       }
       if (argsArr[0] === "plugins" && argsArr[1] === "install") {
         throw new Error("network error during install");
@@ -149,7 +150,7 @@ describe("config preserve on uninstall failure", () => {
     mockExecFileSync.mockImplementation((cmd, args) => {
       const argsArr = args as string[];
       if (argsArr[0] === "config" && argsArr[1] === "file") {
-        return "/home/user/.openclaw/openclaw.json";
+        return FAKE_CFG_PATH;
       }
       if (argsArr[0] === "plugins" && argsArr[1] === "uninstall") {
         throw new Error("uninstall partially failed");
--- a/create-openclaw-octo/cli/install-migration.test.ts
+++ b/create-openclaw-octo/cli/install-migration.test.ts
@@ -13,6 +13,7 @@
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { execFileSync } from "node:child_process";
+import { pathIncludes, pathMatch } from "./__test_utils__/path.js";
 
 vi.mock("node:child_process", () => ({
   execFileSync: vi.fn(),
@@ -90,14 +91,14 @@ async function applyFsMocks(state: FakeFsState) {
   });
   vi.mocked(fs.existsSync).mockImplementation((path: any) => {
     const p = String(path);
-    if (p.includes("/extensions/")) {
+    if (pathIncludes(p, "/extensions/")) {
       // Match `/extensions/<id>` — return true if in extDirs
-      const m = p.match(/\/extensions\/([^/]+)$/);
+      const m = pathMatch(p, /\/extensions\/([^/]+)$/);
       if (m && state.extDirs.has(m[1])) return true;
       return false;
     }
-    if (p.includes("/workspace/")) {
-      const m = p.match(/\/workspace\/([^/]+)$/);
+    if (pathIncludes(p, "/workspace/")) {
+      const m = pathMatch(p, /\/workspace\/([^/]+)$/);
       if (m && state.workspaceDirs.has(m[1])) return true;
       return false;
     }
@@ -106,9 +107,9 @@ async function applyFsMocks(state: FakeFsState) {
   });
   vi.mocked(fs.rmSync).mockImplementation((path: any) => {
     const p = String(path);
-    const m = p.match(/\/extensions\/([^/]+)$/);
+    const m = pathMatch(p, /\/extensions\/([^/]+)$/);
     if (m) state.extDirs.delete(m[1]);
-    const wm = p.match(/\/workspace\/([^/]+)$/);
+    const wm = pathMatch(p, /\/workspace\/([^/]+)$/);
     if (wm) state.workspaceDirs.delete(wm[1]);
   });
 }
--- a/create-openclaw-octo/cli/install.test.ts
+++ b/create-openclaw-octo/cli/install.test.ts
@@ -20,6 +20,7 @@
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { execFileSync } from "node:child_process";
+import { pathEndsWith } from "./__test_utils__/path.js";
 
 vi.mock("node:child_process", () => ({
   execFileSync: vi.fn(),
@@ -440,7 +441,7 @@ describe("runInstall — update scenario", () => {
     });
     mockExistsSync.mockImplementation((path) => {
       // pretend extensions/octo exists so isHealthyInstall("octo") returns true
-      return String(path).endsWith("extensions/octo");
+      return pathEndsWith(path, "extensions/octo");
     });
 
     const { runInstall } = await loadInstall();
@@ -531,7 +532,7 @@ describe("runInstall — update scenario", () => {
         try { Object.assign(state, JSON.parse(String(data))); } catch { /* ignore */ }
       }
     });
-    mockExistsSync.mockImplementation((path) => String(path).endsWith("extensions/octo"));
+    mockExistsSync.mockImplementation((path) => pathEndsWith(path, "extensions/octo"));
 
     const { runInstall } = await loadInstall();
 
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { execFileSync } from "node:child_process";
+import { pathEndsWith, RESOLVED_CFG_PATH } from "./__test_utils__/path.js";
 
 // Mock child_process at module level
 vi.mock("node:child_process", () => ({
@@ -939,9 +940,8 @@ describe("detectInstallState", () => {
       }
       vi.mocked(readFileSync).mockReturnValue(JSON.stringify(cfg));
       vi.mocked(existsSync).mockImplementation((p: unknown) => {
-        const path = String(p);
-        if (path.endsWith("/extensions/octo")) return true;
-        if (path.endsWith(".openclaw/npm/node_modules/openclaw-channel-octo")) {
+        if (pathEndsWith(p, "/extensions/octo")) return true;
+        if (pathEndsWith(p, ".openclaw/npm/node_modules/openclaw-channel-octo")) {
           return Boolean(opts.npmDirResidue);
         }
         return false;
@@ -1123,7 +1123,7 @@ describe("detectInstallState", () => {
       },
     }));
     vi.mocked(existsSync).mockImplementation((p: unknown) =>
-      String(p).endsWith("/extensions/openclaw-channel-octo"),
+      pathEndsWith(p, "/extensions/openclaw-channel-octo"),
     );
     const { detectInstallState } = await loadModule();
     const state = detectInstallState();
@@ -1313,6 +1313,11 @@ describe("getConfigFilePathSafe (Windows relative path)", () => {
     const { getConfigFilePathSafe } = await import("./openclaw-cli.js");
     const result = getConfigFilePathSafe();
 
-    expect(result).toBe("/home/user/.openclaw/openclaw.json");
+    // RESOLVED_CFG_PATH = path.resolve("/home/user/.openclaw/openclaw.json"),
+    // which on POSIX returns the input unchanged but on Windows prefixes
+    // a drive letter and converts to backslashes. Either way, the production
+    // contract (already-absolute paths come back unchanged after normalize)
+    // holds.
+    expect(result).toBe(RESOLVED_CFG_PATH);
   });
 });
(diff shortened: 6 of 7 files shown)


lml2468 (approved):
LGTM — clean CI-only change.

**Verified:**
- Matrix semantics: `os` axis × `include`-injected `package`/`install_cmd` → 3 cross-platform jobs (correct; `include` keys that don't match existing dimensions are broadcast to all combinations)
- `fail-fast: false` correct for independent platform validation
- All build steps (`npm ci`, type-check, build, test) are cross-platform safe — no Linux-specific shell commands
- Job names disambiguated with OS label
- No $ cost impact (public repo, free runners)

No blocking or non-blocking issues.

Jerry-Xin (dismissed):
octo-adapters #84 APPROVED (5bfcd53878e8).

Clean CI-only change — extends the Build & Type-check matrix to cover all three GitHub-hosted runner OSes (ubuntu / macOS / Windows).

**What's correct:**
- `matrix.os` top-level axis + `include` for package/install_cmd — GitHub Actions propagates `include` attrs to all combinations when the entry doesn't reference an axis key. CI output confirms correct expansion (3 jobs spawned).
- `fail-fast: false` is the right choice for a platform matrix: you want to see all failures in one push, not sequentially.
- `actionlint` passes; ubuntu + macOS green; Windows in progress but expected to pass (the underlying build/typecheck steps are platform-agnostic npm operations).

**No blocking issues.** +7 -2, zero code changes, zero risk.

lml2468 (dismissed):
CI change LGTM — matrix expansion is correct and `fail-fast: false` is the right call.

**What I verified:**
- `include` without referencing the `os` key correctly fans `package`/`install_cmd` into all three OS combinations (3 jobs, not 4).
- All build steps (`npm ci`, `npm run type-check`, `npm run build`, `npm test`) are cross-platform; no bash-isms in the workflow.
- `working-directory: ${{ matrix.package }}` uses a bare directory name, no path-separator issue.
- Ubuntu ✅ and macOS ✅ pass; the `changes` job correctly stays on ubuntu-only.

**Windows failures are pre-existing, not introduced by this PR:**
The windows-latest job surfaces exactly what this PR intended to catch:
1. Detection-layer mismatches (`"broken"` / `"dmwork-legacy"` vs expected values) — likely path-separator or cwd handling in the detection logic.
2. Hardcoded Unix path assertion: `expected 'C:\home\user\.openclaw\openclaw.json' to be '/home/user/.openclaw/openclaw.json'` at `openclaw-cli.test.ts:1316`.

These are test/source bugs, not CI bugs. Worth a follow-up issue or PR to fix them.

One heads-up: if the `build` matrix is a **required** status check, the new `windows-latest` sub-job will block merge until those test fixes land. Options:
- Fix the tests in a follow-up before merging this (safest).
- Add `continue-on-error: true` to the Windows job temporarily and track a cleanup issue.
- Merge as-is if the branch-protection rule matches the old job name pattern and the new per-OS names are not yet required.

yujiawei (changes requested):
# Code Review — PR #84 (octo-adapters)

## Summary

This PR extends the `build` job's matrix from `ubuntu-latest` only to `[ubuntu-latest, macos-latest, windows-latest]` with `fail-fast: false`. The workflow YAML itself is well-formed and does exactly what the PR description claims — three jobs run, all three OSes report independently, the existing `include` block expands cleanly across the new axis.

However, on the current head (`5bfcd538`) the new `Build & Type-check (create-openclaw-octo, windows-latest)` job is **failing** with 13+ test errors. Merging as-is will make `main`'s CI permanently red on Windows for `create-openclaw-octo` until follow-up fixes land. The PR description does not acknowledge this. The verdict below reflects that rollout problem, not a defect in the YAML itself.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Matrix expansion produces 3 OS jobs | ✅ | Run `26486289440` shows three `Build & Type-check (create-openclaw-octo, …)` jobs (ubuntu/macos/windows) |
| `fail-fast: false` honored | ✅ | macOS (pass) and Ubuntu (pass) finished even with Windows failing in the same run |
| Job-name rename does not break required-status-check rules | ✅ | `gh api repos/Mininglamp-OSS/octo-adapters/branches/main/protection` shows only `check-sprint / check-sprint` is required; the build job was never wired into branch protection by name, so renaming `Build & Type-check (create-openclaw-octo)` → `Build & Type-check (create-openclaw-octo, <os>)` does not orphan a required check |
| `actionlint` clean | ✅ | `sanity / actionlint` passes on this PR |
| Cross-platform `npm` scripts in `create-openclaw-octo/package.json` | ✅ | `prebuild` uses inline `node -e` (avoids `rm -rf`); `build` is `tsc`; `test` is `vitest run`. All platform-agnostic |
| Workflow itself well-formed | ✅ | `runs-on: ${{ matrix.os }}` and `defaults.run.working-directory: ${{ matrix.package }}` resolve correctly on all three runners (Windows worktree path `D:/a/octo-adapters/octo-adapters/create-openclaw-octo` confirmed in the log) |
| `windows-latest` job green on head SHA | ❌ | 13+ test failures — see §2 P1-1 |

## 2. Findings

### P1-1 — Adding the `windows-latest` job as it stands turns `main`'s CI red

`Build & Type-check (create-openclaw-octo, windows-latest)` is failing on `5bfcd538` with 13+ assertion errors. Sample (file:line excerpts from job `77994380518`):

- `cli/openclaw-cli.test.ts:1316` — `Expected: "/home/user/.openclaw/openclaw.json"`, `Received: "C:\home\user\.openclaw\openclaw.json"`. Test hard-codes a POSIX-shaped path; `path.join` / `path.resolve` returns a backslash path on Windows.
- `cli/openclaw-cli.test.ts:956 / 969 / 980 / 994 / 1005 / 1016` — `Expected: "octo-clawhub"`, `Received: "broken"`. CLI-detection returns `"broken"` instead of the expected backend on Windows.
- `cli/openclaw-cli.test.ts:1035 / 1072 / 1088` — `Expected: "octo-clawhub"`, `Received: "dmwork-legacy"`. Detection precedence diverges on Windows.
- `cli/openclaw-cli.test.ts:1130` — `Expected: "octo-npm-legacy"`, `Received: "broken"`.
- `cli/install-migration.test.ts:299 / 453 / 589` — `Migration aborted: post-install verification: octo plugin not healthy`. Post-install health check trips on Windows.
- Earlier in the same log: `Failed to execute openclaw --version: permission denied` — Windows EACCES on a non-`.cmd` shim, the most likely root cause feeding the cascade above.

These are real cross-platform bugs in the production code / test fixtures — the matrix expansion is **doing exactly what it advertises** by surfacing them. The issue is the *rollout strategy*: once this lands, `main`'s `windows-latest` job is red until the underlying bugs are fixed, and the PR description doesn't tell the reader that's the expected post-merge state.

**Suggested resolutions (pick one):**

1. **Stack the fixes first.** Land a separate PR (or stack one before this) that addresses at minimum (a) the hardcoded POSIX path in `cli/openclaw-cli.test.ts:1316`, (b) the CLI-detection logic that returns `"broken"` on Windows (looks like the `.cmd` shim / `EACCES` path), and (c) the post-install health-check failure surface. Merge the matrix expansion only once Windows is green. Cleanest outcome — `main` stays green and the new gate carries real signal from the moment it lands.
2. **Land matrix as observability-only.** Add `continue-on-error: ${{ matrix.os == 'windows-latest' }}` (or split out an `experimental` flag in `include`) with a `# TODO: remove once Windows tests pass — tracked in <issue>` comment. Acceptable if the visibility benefit is wanted now and Windows fixes will take more time than this PR.
3. **Skip the failing tests on Windows with a clear TODO.** `describe.skipIf(process.platform === 'win32')` plus a single tracking ticket. Worst of the three because it hides real bugs until someone remembers to unskip; only use if (1) and (2) are both blocked.

Whichever path is chosen, the PR description should state it so reviewers and on-call understand why a freshly-added matrix entry can be red on `main`.

### P2-1 — No `cache: 'npm'` on `actions/setup-node`

The PR triples `npm ci` work (3 OSes instead of 1). `actions/setup-node` accepts `cache: 'npm'` with `cache-dependency-path: ${{ matrix.package }}/package-lock.json`, which typically saves 5–10 s per runner per run. Not a blocker — natural follow-up that pairs with this change.

### P2-2 — `install_cmd` in `matrix.include` is degenerate

`install_cmd` is always `npm ci`. The `include` block is structured for multi-package extensibility, which is fine forward-looking design, but right now it adds one indirection without reader-visible benefit. Not worth changing in this PR; flagging only so a future reader knows it's intentional headroom, not orphaned scaffolding.

### P2-3 — Pre-existing: pinned actions are on Node 20

GitHub's runner-side warning at the end of every job: *"Node.js 20 actions are deprecated… `actions/checkout@34e1148…`, `actions/setup-node@49933ea5…`."* Not introduced by this PR (these pins exist on `main`), but worth a separate follow-up to bump both to v5 before June 2026 when the runner default flips.

## 3. Recommendations

- Adopt resolution path **(1)** or **(2)** for P1-1 before merging. Option (1) is the cleaner outcome; option (2) is acceptable if Windows fixes are not in immediate reach.
- File follow-ups for P2-1 (npm cache) and P2-3 (Node 20 → v5 action bumps) at the next cleanup window — neither needs to block this PR once P1-1 is resolved.

## 4. Additional Observations

- The choice of `fail-fast: false` is correct for a cross-platform matrix where each OS surfaces a different class of bug; the inline comment justifying it reads well.
- Workflow-level `concurrency.cancel-in-progress: true` plus job-level `fail-fast: false` is a clean combination: a new push still cancels stale runs, but within one run all three OSes report independently — exactly the behaviour the PR description argues for.
- The `paths-filter` preflight (`changes` job) still gates the matrix correctly; a docs-only PR will skip all three OS jobs uniformly, so the matrix expansion does not change the docs-only contract that satisfies branch protection via "skipped = passing".

Jerry-Xin (changes requested):
Re-review on 2f57a4bc85c8.

Good progress — the `__test_utils__/path.ts` helper is the right abstraction and fixes most of the Windows failures. However **2 tests still fail on `windows-latest`** (job `77995869005`):

## 🔴 Blocking

### 1. Drive-letter mismatch in `getConfigFilePathSafe` test (openclaw-cli.test.ts:1321)

```
Expected: "D:\home\user\.openclaw\openclaw.json"
Received: "C:\home\user\.openclaw\openclaw.json"
```

`RESOLVED_CFG_PATH` is computed at **import time** via `path.resolve(FAKE_CFG_PATH)`. On Windows, `path.resolve("/home/...")` prepends `process.cwd()`'s drive letter. The test helper imports on D: (runner workdir), but if the production code resolves on C: (different CWD context during the test run), the drive letters diverge.

**Fix options:**
- Assert only the portion after the drive letter: `expect(result.slice(2)).toBe(RESOLVED_CFG_PATH.slice(2))` — confirms normalization without coupling to CWD's drive.
- Or use `path.parse` and compare `root` separately from `dir+base`.

### 2. `copyFileSync` spy mismatch in `restoreChannelConfigToFile` (config-preserve.test.ts)

Same root cause — the expected `RESOLVED_CFG_PATH` and the actual value passed to the spy differ by drive letter. Same fix strategy applies.

## 💬 Non-blocking

- ubuntu + macOS green, `check-sprint` green, `actionlint` passes — the YAML itself remains correct.
- The `pathEndsWith`/`pathIncludes`/`pathMatch` utilities are well-designed and fix the other 11+ failures cleanly.

## Verdict

Two residual Windows failures share the same root cause (drive-letter sensitivity of `path.resolve` at import time). A small fix to the assertion strategy will make the full matrix green. Requesting changes so Windows goes green before merge.

lml2468 (changes requested):
## Re-review on 2f57a4bc

Good progress: 13 failures → 2. The `__test_utils__/path.ts` abstraction is well-structured and the pattern (normalize separators before matching) is correct.

### Remaining 2 failures — same root cause

Both stem from `RESOLVED_CFG_PATH = path.resolve("/home/user/.openclaw/openclaw.json")` being evaluated once at **module-import time**. On Windows, `/home/...` is drive-relative — `path.resolve` prepends the current drive letter. The drive letter depends on CWD *at the time of evaluation*:

1. **`openclaw-cli.test.ts` → `getConfigFilePathSafe (Windows relative path)`**
   ```
   Expected: "D:\home\user\.openclaw\openclaw.json"   (RESOLVED_CFG_PATH, computed when CWD=D:)
   Received: "C:\home\user\.openclaw\openclaw.json"   (production code resolves at a different CWD)
   ```

2. **`config-preserve.test.ts` → `restoreChannelConfigToFile` → backup assertion**
   Same drive-letter mismatch propagates to `copyFileSync` spy arguments.

### Suggested fix

Replace the static `RESOLVED_CFG_PATH` constant with a function that resolves lazily at assertion time, OR switch to an assertion that ignores the drive letter prefix:

```ts
// Option A: lazy resolution (simplest)
export function resolvedCfgPath() { return path.resolve(FAKE_CFG_PATH); }

// Option B: drive-letter-agnostic assertion helper
export function assertSamePath(actual: string, expected: string) {
  // Strip optional drive letter for comparison
  const normalize = (p: string) => p.replace(/^[A-Z]:/i, "").replace(/\\/g, "/");
  expect(normalize(actual)).toBe(normalize(expected));
}
```

Option A is cleaner if the issue is only a timing difference; Option B is more defensive if CWD genuinely differs between test setup and production code execution.

### Summary

The test utility pattern is correct; just the `RESOLVED_CFG_PATH` static evaluation is foiled by Windows drive-letter relativity. One small follow-up commit should bring Windows green.

yujiawei (changes requested):
# Code Review — PR #84 (octo-adapters)

## Summary

The follow-up commit (`2f57a4bc`) is on the right track — it correctly identifies that the Windows breakage was test-only (production code already does the right `path.resolve` dance) and introduces shared, separator-agnostic test helpers (`__test_utils__/path.ts`) plus a `tsconfig` exclusion so the helpers don't ship to `dist/`. That collapses the prior 17 Windows failures down to 2.

However, **the two remaining Windows failures are still red on the current HEAD**, so this PR cannot land yet without turning `main`'s `windows-latest` job red. The remaining failures are caused by the new helper itself: `RESOLVED_CFG_PATH = path.resolve(FAKE_CFG_PATH)` uses Node's standalone `path.resolve`, which on Windows resolves a POSIX-style absolute path against the **cwd's drive letter** — but production's `normalizeConfigPath` resolves it against **`homedir()`'s drive letter**. On the GitHub Windows runner those drives differ (`D:` vs `C:`), so the assertion comparing the two never matches.

This is a one-line fix in `__test_utils__/path.ts`; the production code does not need to change.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| `__test_utils__/path.ts` excluded from `dist/` | ✅ | `create-openclaw-octo/tsconfig.json` adds `"**/__test_utils__/**"` to `exclude` (next to existing `**/*.test.ts` rule); both ubuntu and macOS jobs build green |
| `pathEndsWith` / `pathIncludes` / `pathMatch` correct semantics | ✅ | All three normalize `\` → `/` before applying `endsWith` / `includes` / `match`, so a Windows-resolved path like `C:\…\extensions\octo` correctly matches the POSIX literal `extensions/octo` |
| Helper migration covers the four affected test files | ✅ | `config-preserve.test.ts`, `install-migration.test.ts`, `install.test.ts`, `openclaw-cli.test.ts` — all the originally failing files were touched; no production source modified |
| Reduces previous Windows failures (17 → ≤2) | ✅ | Run `26486776513` Windows job: `Tests 2 failed | 117 passed (119)` — the 15 detection-layer failures (`"broken"`, `"dmwork-legacy"`) and the 3 install-migration "post-install verification" cascades are all gone |
| ubuntu-latest still green | ✅ | Run `26486776513` job `77995869024` succeeded |
| macos-latest still green | ✅ | Run `26486776513` job `77995869009` succeeded |
| windows-latest green on PR head | ❌ | Run `26486776513` job `77995869005` fails — see §2 P1-1 |

## 2. Findings

### P1-1 — `RESOLVED_CFG_PATH` does not match production's resolved form on Windows; PR HEAD still red

Both remaining Windows failures share one root cause:

```
× cli/openclaw-cli.test.ts > getConfigFilePathSafe (Windows relative path) > should keep absolute paths unchanged
  → expected 'C:\home\user\.openclaw\openclaw.json' to be 'D:\home\user\.openclaw\openclaw.json'

× cli/config-preserve.test.ts > restoreChannelConfigToFile > should write channels.octo back to file with backup
  → expected "spy" to be called with arguments: [ 'D:\home\user\.openclaw\openclaw.json', 'D:\home\user\.openclaw\openclaw.json.bak' ]
    Received: 'C:\home\user\.openclaw\openclaw.json', 'C:\home\user\.openclaw\openclaw.json.bak'
```

**Why the drives differ.** Walking through `normalizeConfigPath` on `cli/openclaw-cli.ts:171-178` with the mocked input `/home/user/.openclaw/openclaw.json` on a Windows runner where cwd is on `D:` and `homedir()` is `C:\Users\runneradmin`:

1. None of the `~/`, `~\`, or `./.openclaw/` branches match.
2. `resolve(p) === p` → `resolve("/home/user/.openclaw/openclaw.json")` returns `D:\home\user\.openclaw\openclaw.json` (cwd's drive). That is not equal to the POSIX input → falls through.
3. Final branch: `return resolve(homedir(), p)`. On Windows, when the second argument starts with `/` (a drive-relative absolute), Node's `path.win32.resolve` keeps the **drive of the first argument**. So the result is `C:\home\user\.openclaw\openclaw.json` (homedir's drive).

But the helper computes:

```ts
// cli/__test_utils__/path.ts:21
export const RESOLVED_CFG_PATH = path.resolve(FAKE_CFG_PATH);
```

`path.resolve(FAKE_CFG_PATH)` with no leading absolute base resolves against **cwd**, which on the GitHub Windows runner is `D:\a\octo-adapters\octo-adapters\create-openclaw-octo`. Result: `D:\home\user\.openclaw\openclaw.json`.

So `RESOLVED_CFG_PATH` and what production actually returns differ in their drive letter on Windows: `D:` vs `C:`. POSIX is unaffected because there are no drive letters.

The block comment at `openclaw-cli.test.ts:1316-1320` reflects the wrong mental model:

> *"… the production contract (already-absolute paths come back unchanged after normalize) holds."*

That contract holds on POSIX, but on Windows `/home/...` is **drive-relative**, not absolute, so `normalizeConfigPath` (intentionally) does *not* return it unchanged — it falls into the `resolve(homedir(), p)` branch. Either the test name + comment need to acknowledge that, or — simpler — the helper needs to mirror what production does.

**Suggested fix (one line, test-only):**

```ts
// cli/__test_utils__/path.ts
import { homedir } from "node:os";

export const RESOLVED_CFG_PATH = path.resolve(homedir(), FAKE_CFG_PATH);
```

That mirrors `normalizeConfigPath`'s final branch exactly. On POSIX, since `homedir()` is absolute and `FAKE_CFG_PATH` is also POSIX-absolute, `resolve` discards the first argument and still returns `/home/user/.openclaw/openclaw.json` — so existing POSIX behaviour is preserved. On Windows it now returns `C:\home\user\.openclaw\openclaw.json`, matching what `getConfigFilePathSafe` and `restoreChannelConfigToFile` actually call out. Both Windows tests should pass after this single change.

The two existing comments (in `openclaw-cli.test.ts` and `path.ts`) should be tightened to say the helper mirrors the production resolution path (`resolve(homedir(), p)`) rather than claiming "absolute paths come back unchanged" — that phrasing will keep tripping the next reader.

I'd recommend re-running CI after the helper fix and confirming all three OS jobs are green before merging — the whole point of this PR is that "matrix expansion" and "Windows is green" land together.

### P2-1 — Marker comment naming inside `__test_utils__/path.ts` is off-by-one

`pathMatch` in `__test_utils__/path.ts:43-45` is documented as "Returns the first match (with any capture groups) or null." That's accurate for `String.prototype.match` against a non-global regex, but the doc could note it does **not** preserve `lastIndex` semantics if a global regex is passed in — `.match(/.../g)` returns capture-less array of full matches, which would silently break callers that expect groups. None of the current call sites use `/g`, so it's not a bug today, just a footgun for the next contributor. Optional cleanup: throw when `regex.flags.includes("g")` or document the non-global expectation. Not blocking.

### P2-2 — `npm cache` still missing on `actions/setup-node` (carry-over)

Same observation from the previous review of HEAD `5bfcd538`: the matrix triples `npm ci` work without enabling `cache: 'npm'` + `cache-dependency-path: ${{ matrix.package }}/package-lock.json`. Typical save is 5–10s per runner per run and pairs naturally with this PR. Still non-blocking; flagging again so it doesn't get lost when the Windows fix lands.

### P2-3 — Pre-existing: pinned actions are on Node 20 (carry-over)

`actions/checkout@34e1148…` and `actions/setup-node@49933ea5…` are Node 20-flavoured; GitHub will start deprecation-warning around June 2026. Still not introduced by this PR; same recommendation as before to bump in a separate PR.

## 3. Recommendations

- **Blocking:** apply the one-line `RESOLVED_CFG_PATH = path.resolve(homedir(), FAKE_CFG_PATH)` fix in `__test_utils__/path.ts`, retighten the comment in `openclaw-cli.test.ts:1316-1320` and the JSDoc in `__test_utils__/path.ts` so both describe "mirrors `normalizeConfigPath`'s final branch", and re-run CI. Once all three OS jobs are green on the same commit, this is ready to land.
- Non-blocking follow-ups (separate PRs are fine): P2-1 (npm cache), P2-2 (`pathMatch` global-regex guard), P2-3 (Node 20 → v5 action bumps).

## 4. Additional Observations

- **Security review (PR was tagged `security_sensitive`).** I read the diff with that lens explicitly. No real risk surface here — pinned third-party actions stay pinned by SHA, no new permissions are requested, no `pull_request_target` is introduced, no secrets are added or read in workflow expressions, and `npm ci` continues to install from the existing committed `package-lock.json`. The new `__test_utils__/path.ts` is excluded from the published `dist/`, performs no I/O at import beyond a single `path.resolve`, and contains no `eval`-class constructs. The `security_sensitive` classification appears to be a keyword-match false positive driven by `token` / `secret` strings in test fixtures (e.g. `botToken: "bf_real_secret_token"`). Worth flagging in the dispatcher's heuristic if that's tunable; nothing for this PR to action.
- **Diff hygiene.** The new helper is well-scoped: tiny surface, comments at each export explaining intent, and the four migrated test files all use it consistently. Once the resolve-vs-homedir mismatch is fixed, the structural shape of this change is exactly what cross-platform mocks should look like in this repo.
- **`fail-fast: false` was the right call (carry-over).** Currently it's the only reason `windows-latest`'s 2 remaining failures aren't masking ubuntu/macOS green; a future regression where `ubuntu-latest` fails first would still surface the Windows result rather than cancelling it.

lml2468 (approved):
## Re-review on e788ffdbae3b — LGTM

v3 correctly mirrors the production `normalizeConfigPath` logic in the test utility:

```ts
export const RESOLVED_CFG_PATH = (() => {
  if (path.resolve(FAKE_CFG_PATH) === FAKE_CFG_PATH) return FAKE_CFG_PATH;
  return path.resolve(homedir(), FAKE_CFG_PATH);
})();
```

On Windows runners where CWD is on D: but homedir is on C:, `path.resolve("/home/...")` would give `D:\home\...` (drive from CWD), but the production code resolves against homedir → `C:\home\...`. This IIFE replicates that same resolution path, so the constant always matches what the spy actually receives.

**CI status:** all three platform jobs green (ubuntu 18s, macOS 18s, Windows 55s).

No remaining issues. Previous P2s (npm cache, Node 20 action bumps) are natural follow-ups, not blockers.

Jerry-Xin (approved):
Re-review on e788ffdbae3b — all prior concerns resolved. **APPROVED.**

## CI Status

All three platform jobs are now green:
- `ubuntu-latest` — ✅ 18s
- `macos-latest` — ✅ 18s
- `windows-latest` — ✅ 55s
- `actionlint` — ✅
- `check-sprint` — ✅

## What changed since v2 (2f57a4bc)

`RESOLVED_CFG_PATH` now mirrors the production `normalizeConfigPath` logic:

```typescript
export const RESOLVED_CFG_PATH = (() => {
  if (path.resolve(FAKE_CFG_PATH) === FAKE_CFG_PATH) return FAKE_CFG_PATH;
  return path.resolve(homedir(), FAKE_CFG_PATH);
})();
```

This correctly handles the Windows drive-letter divergence: on POSIX, `/home/...` is already absolute so it passes through unchanged; on Windows, `path.resolve("/home/...")` ≠ input (drive letter prepended), so it falls into the `homedir()` branch — exactly matching what production code does. No more C: vs D: mismatches regardless of CWD.

## ✅ Highlights

- `__test_utils__/path.ts` is a well-factored cross-platform test utility — `pathEndsWith`/`pathIncludes`/`pathMatch` normalize separators before matching, eliminating an entire class of Windows false-negatives.
- `tsconfig.json` excludes `__test_utils__/**` from compilation target — correct, these are test-only utilities.
- `fail-fast: false` gives full per-push visibility into all three platform outcomes.
- The matrix expansion delivers real signal: it surfaced and forced fixes to real cross-platform bugs in the same PR cycle.

No blocking issues. Clean to merge.

yujiawei (approved):
# Code Review — PR #84 (octo-adapters)

## Summary

The `e788ffdb` follow-up commit resolves the drive-letter mismatch flagged on the previous HEAD (`2f57a4bc`). `RESOLVED_CFG_PATH` now mirrors production's `normalizeConfigPath` resolution path exactly — when the input is not absolute by Node's definition (the Windows case for `/home/...`), it resolves against `homedir()` instead of `cwd`. With that one change, all three OS jobs on the matrix are green on the current head.

This is the structural shape we want for cross-platform mocks in this repo: separator-agnostic helpers (`pathEndsWith`, `pathIncludes`, `pathMatch`), a single resolved-config-path constant that mirrors the production normaliser, and a `tsconfig` exclusion so the helpers do not ship in `dist/`. APPROVED.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| `RESOLVED_CFG_PATH` mirrors `normalizeConfigPath` | ✅ | `cli/__test_utils__/path.ts:21-25` — `if (path.resolve(p) === p) return p; else resolve(homedir(), p)` matches `cli/openclaw-cli.ts:170-180` line for line for the absolute / drive-relative branches |
| Windows runner — both previously red tests now pass | ✅ | Run `26486938757` job `77996345705` on `e788ffdb`: `Build & Type-check (create-openclaw-octo, windows-latest)` PASS in 55s. The two assertions that failed on `2f57a4bc` (`getConfigFilePathSafe (Windows relative path)` and `restoreChannelConfigToFile … backup`) are part of that run |
| Ubuntu / macOS still green | ✅ | Same run: `ubuntu-latest` (18s) PASS, `macos-latest` (18s) PASS |
| CodeQL clean on the new commit | ✅ | `analyze / CodeQL (javascript-typescript)` PASS (1m12s) on this PR head |
| Production source unchanged in the fix commit | ✅ | `gh api .../compare/2f57a4bc...e788ffdb` shows only `cli/__test_utils__/path.ts` modified (+22 / -9). No production file touched |
| Helpers stay out of published `dist/` | ✅ | `create-openclaw-octo/tsconfig.json` excludes `**/__test_utils__/**` (added alongside the existing `**/*.test.ts` rule) |
| `actionlint` clean | ✅ | `sanity / actionlint` PASS |
| Required-status-check rules unaffected by job-name change | ✅ | `gh api repos/Mininglamp-OSS/octo-adapters/branches/main/protection` returns `["check-sprint / check-sprint"]` only — the renamed `Build & Type-check (…, <os>)` jobs were never required by name, so no orphaned check |
| Cross-platform `npm` scripts in `create-openclaw-octo/package.json` | ✅ | `prebuild` uses inline `node -e` (avoids `rm -rf`); `build` is `tsc`; `test` is `vitest run`. All platform-agnostic |

## 2. Findings

### P2-1 — `pathMatch` global-regex footgun (carry-over, optional)

`pathMatch` in `cli/__test_utils__/path.ts:55-57` delegates to `String.prototype.match`. With a non-global regex (every current call site) it returns `[fullMatch, ...groups]`, which is the documented behaviour. With a `/g` regex `match` returns `[fullMatch1, fullMatch2, …]` with no capture groups — silently breaking any caller that expects `m[1]`. Not a bug today; flagging as a future-contributor footgun. Optional cleanup: throw when `regex.flags.includes("g")`, or document the non-global expectation in JSDoc. Non-blocking.

### P2-2 — `cache: 'npm'` still missing on `actions/setup-node` (carry-over)

The matrix triples `npm ci` work without enabling `cache: 'npm'` + `cache-dependency-path: ${{ matrix.package }}/package-lock.json`. Typical save 5–10s per runner per run. Pairs naturally with this change but doesn't need to block it. Worth a follow-up PR.

### P2-3 — Pre-existing: pinned actions are on Node 20 (carry-over)

`actions/checkout@34e1148…` and `actions/setup-node@49933ea5…` are Node 20-flavoured; GitHub will start deprecation-warning around June 2026. Pre-existing on `main`, not introduced by this PR. Bump in a separate PR.

## 3. Recommendations

- **No blocking changes.** Ready to merge as soon as the dispatcher's `code-review` status check flips on this approval.
- Non-blocking follow-ups (separate PRs): P2-1 (`pathMatch` global-regex guard), P2-2 (npm cache on `setup-node`), P2-3 (Node 20 → v5 action bumps).

## 4. Additional Observations

- **Security review (PR was tagged `security_sensitive`).** Re-checked the diff under that lens. Nothing actionable — pinned third-party actions stay pinned by SHA, no new permissions are requested, no `pull_request_target` is introduced, no secrets are read or written in workflow expressions, `npm ci` continues to install from the existing committed `package-lock.json`. The new `__test_utils__/path.ts` performs no I/O at import beyond `path.resolve(homedir(), …)` and is excluded from the published `dist/`. The `security_sensitive` classification on this PR was a keyword false positive driven by `token` / `secret` strings inside test fixtures (e.g. `botToken: "bf_real_secret_token"`) — worth tuning in the dispatcher heuristic if that's reachable, but nothing for this PR to action.
- **Diff hygiene.** The fix is one self-contained commit, scoped to test utilities only. Production code is unchanged across the entire PR — exactly the right shape for a cross-platform CI hardening change that "just exposes existing test brittleness". The commit message also documents the cwd-vs-homedir reasoning, so the next reader who touches `RESOLVED_CFG_PATH` will understand why it doesn't simply call `path.resolve(FAKE_CFG_PATH)`.
- **`fail-fast: false` continues to be the right call.** Without it, the cascade of Windows-only failures observed mid-PR would have masked the green ubuntu/macOS runs and slowed iteration. Now that all three are green, the same setting protects future regressions from being hidden by an unrelated platform's failure.
- **Matrix-rollout strategy worked.** This PR ended up being the canonical example of "land matrix + fix the brittleness it exposes in the same change" — the alternative ("land matrix red, follow up later") would have cost on-call signal. Good outcome.

caster-Q:
@yujiawei @Jerry-Xin — your reviews on `2f57a4b` arrived after I'd already pushed `e788ffd`. The diagnosis in both your reviews is correct: drive-letter mismatch because `RESOLVED_CFG_PATH = path.resolve(FAKE_CFG_PATH)` keys off cwd, while production's `normalizeConfigPath` keys off `homedir()`. On the GitHub windows-latest runner cwd is `D:` and homedir is `C:`.

Commit `e788ffd` implements yujiawei's suggested fix verbatim — `RESOLVED_CFG_PATH` now uses `path.resolve(homedir(), FAKE_CFG_PATH)` (with a fast-path for when input is already absolute, mirroring production's `normalizeConfigPath` line-for-line):

```ts
export const RESOLVED_CFG_PATH = (() => {
  if (path.resolve(FAKE_CFG_PATH) === FAKE_CFG_PATH) return FAKE_CFG_PATH;
  return path.resolve(homedir(), FAKE_CFG_PATH);
})();
```

CI on the new HEAD `e788ffd`:

```
Build & Type-check (create-openclaw-octo, ubuntu-latest)   pass  18s
Build & Type-check (create-openclaw-octo, macos-latest)    pass  18s
Build & Type-check (create-openclaw-octo, windows-latest)  pass  55s
```

Run: https://github.com/Mininglamp-OSS/octo-adapters/actions/runs/26486938757

Please re-review when you have a moment.

PR #85, opened 2026-05-27
fix: move permissions to workflow top level for pr-contributor-welcome

lml2468:
## 问题

`.github/workflows/pr-contributor-welcome.yml` 存在权限配置 bug：

- 顶层 `permissions: {}` 清空了所有权限
- job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效
- 导致 `GITHUB_TOKEN` 没有写权限，发欢迎评论时出现 403 错误

## 修复

将权限配置移至 workflow 顶层，并移除 job 级别的 `permissions` 块：

```yaml
# 修复前（错误写法）
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write

# 修复后（正确写法）
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

## 参考

GitHub 文档：[reusable workflow 中的权限继承](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — job 级别的 `permissions` 对 `uses:` 类型的 job 不适用，必须在 workflow 顶层设置。

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -4,10 +4,9 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   welcome:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
-    permissions:
-      issues: write


lml2468 (commented):
LGTM — correct fix for a documented GitHub Actions limitation.

**Root cause confirmed:** job-level `permissions` blocks are [not supported](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) for `uses:` (reusable workflow) jobs. Moving `issues: write` to the workflow top level is the right fix.

**CI notes:**
- `welcome` failure is expected — `pull_request_target` runs the workflow from the *default branch* (not the PR branch), so the broken version is still in effect until this merges.
- `check-sprint` failure is unrelated to this change.

yujiawei (approved):
# Code Review — PR #85 (octo-adapters)

## Summary

Tiny, surgical workflow fix: moves `permissions: issues: write` from the calling job to the workflow top level, so the reusable `pr-contributor-welcome` workflow actually has a token that can post issue comments. The previous configuration was effectively `permissions: none` and produced 403 at comment time.

## Verification

| Check | Result | Evidence |
|---|---|---|
| Root cause matches the PR description | ✅ | Workflow-level `permissions:` is the upper bound for all jobs in a workflow. Setting top-level `permissions: {}` strips every scope from `GITHUB_TOKEN`, and a calling job's `permissions:` block can only *narrow* — never *widen* — what the workflow already granted. Hence `issues: write` declared on the calling job was capped to nothing, and posting a welcome comment failed with 403. |
| Fix is canonical | ✅ | Moving `permissions: issues: write` to the workflow top level is the documented pattern for a workflow whose only job is `uses:` a reusable workflow. The job-level block is correctly dropped, since with a single job it would be redundant once the top-level grant is in place. |
| Reusable workflow contract preserved | ✅ | `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main` declares its own job-level `permissions: issues: write` under workflow-level `permissions: {}`. When called, it inherits the caller's permissions as the upper bound; the caller now provides `issues: write`, so the inner job-level request is satisfiable. |
| Security posture of `pull_request_target` | ✅ | The reusable workflow explicitly does **not** check out PR code (comment-documented and enforced by the absence of `actions/checkout`). It only reads PR metadata via the GitHub API, so granting `issues: write` to a `pull_request_target`-triggered workflow does not create a fork-exfiltration vector. The `zizmor: ignore[dangerous-triggers]` annotation remains appropriate. |
| Blast radius of widening permissions to workflow scope | ✅ | The workflow contains exactly one job (`welcome`) which is a pure `uses:` call. There are no additional steps or jobs that gain unintended write access from the move. |
| Diff matches description | ✅ | `.github/workflows/pr-contributor-welcome.yml`: `-permissions: {}` + job-level `permissions: issues: write` block removed; `+permissions:\n  issues: write` added at top. Net: +2 / −3, single file. |

## Findings

None at P0/P1. No blocking issues, no security regressions, no functional concerns.

### Nits / Optional follow-ups (non-blocking)

- **N1 (optional, doc-only):** Consider adding a one-line comment above the top-level `permissions:` block — e.g. `# Top-level (not job-level) is required because jobs.welcome uses a reusable workflow; calling-job permissions can only narrow, not widen, the workflow grant.` This encodes the lesson learned in the file itself, so the next maintainer doesn't reintroduce the empty `{}` while trying to tighten scope. Not required for merge.
- **N2 (optional, org-wide):** Other caller repos in `Mininglamp-OSS/*` that invoke the same reusable workflow may carry the same `permissions: {}` + job-level grant pattern. Worth a quick grep across sibling repos as a separate task; out of scope for this PR.

## Additional Observations

- `headRefOid` confirmed as `2079fc8ddb313fc77d27aeb879ad190d005e688a`; PR is not draft, GitHub reports `mergeable: true`.
- Branch protection / CODEOWNERS interaction is not affected by this change.
- No secret usage, no third-party action additions, no version bumps. Pinned action SHA in the reusable workflow is unchanged by this PR.

## Verdict

**APPROVED** — correct root cause, minimal surface area, no security regression.

Jerry-Xin (approved):
octo-adapters #85 APPROVED (2079fc8ddb31).

Correct fix. GitHub Actions docs explicitly state that job-level `permissions` is **not supported** for `uses:` (reusable workflow) jobs — the callee inherits exclusively from the workflow top-level `permissions` block. The previous config (`permissions: {}` at top + `permissions: issues: write` at job level) resulted in the callee receiving zero permissions, causing 403 on comment creation.

**Verified:**
- Fix moves `issues: write` to workflow top level, removes the ineffective job-level block — exactly what the docs prescribe.
- `actionlint` passes.
- Build matrix (ubuntu/macOS/Windows) all green — no code changes involved.
- `welcome / welcome` FAILURE is expected: `pull_request_target` reusable-workflow permission changes only take effect after merge to `main`.
- Security: `pull_request_target` with `types: [opened]` calling a `@main`-pinned reusable workflow (no PR code checkout) + minimal `issues: write` scope — no escalation risk.

No blocking issues. +2 -3, single-file workflow fix.

PR #91, opened 2026-05-29
fix(create-openclaw-octo): hard-abort install on incompatible OpenClaw versions

### LINKED ISSUE #90: fix(create-openclaw-octo): hard-abort install on OpenClaw < 2026.3.22 instead of soft warn
## Description

When OpenClaw version is below the technical hard floor (`2026.3.22` — when `clawhub:` protocol was introduced upstream), `npx -y create-openclaw-octo install` runs through migration steps 1–4 (backup config, disable plugin, remove `channels.dmwork`, remove bindings), then dies at step 5 with:

```
unsupported npm spec: protocol specs are not allowed
Migration failed (pluginsInstall(octo) threw: Command failed:
  /home/.../openclaw plugins install clawhub:octo
  unsupported npm spec: protocol specs are not allowed
). Rolling back...
Rollback complete. Backup retained at .rebrand-backup.

file:///.../create-openclaw-octo/dist/cli/install.js:298
        throw new Error(`Migration aborted: ${reason}`);
              ^

Error: Migration aborted: ...
    at rollback (...:298:15)
    at runMigration (...:342:13)
    at runRebrandMigration (...:512:5)
    at runInstall (...:104:13)
    ...
```

The user sees a raw Node.js stack trace with no actionable guidance.

## Root Cause

Two version-check implementations coexist in the codebase:

- `cli/openclaw-cli.ts:1394+` — Properly designed `detectOpenClawState()` returning `{kind: "block" | "warn" | "ok"}` (supports hard abort).
- `cli/utils.ts:83+` — `ensureOpenClawCompat()` only emits a soft warn (line 99) and continues, even though the implementation plan called for hard block.

`ensureOpenClawCompat()` is the function actually called at the install entry (`cli/install.ts:119`), but it never invokes the properly-designed `detectOpenClawState()`.

Additionally, `OPENCLAW_PEER_MIN = "2026.4.15"` in `cli/openclaw-cli.ts:1394` is the plugin repo's `peerDependencies` pick, but the **actual technical hard floor** is **`2026.3.22`** — the first OpenClaw release containing `openclaw plugins install clawhub:<spec>` (upstream commit `91b2800241` "feat: add native clawhub install flows").

## Steps to Reproduce

1. Install OpenClaw < 2026.3.22 (verified with `2026.3.13`).
2. Have an existing legacy plugin installed (`openclaw-channel-dmwork@0.5.x` or similar).
3. Run `npx -y create-openclaw-octo@dev install`.

## Expected Behavior

`ensureOpenClawCompat()` detects the unsupported OpenClaw version at entry, prints a friendly error with `openclaw update` guidance, and `process.exit(2)` **without** touching config.

## Actual Behavior

Soft warn → proceeds through 4 steps modifying config → fails at step 5 with stack trace → rollback restores config → exits with raw Node stack trace dump.

## Adapter Version

`create-openclaw-octo@1.0.1-dev.6cb329f` (`@dev` channel)

## Affected Component

Claude Code Adapter (`create-openclaw-octo` CLI)

## Severity

P1 — install fails for users on older OpenClaw with poor UX (no actionable error).

## Reproducibility

Consistently reproducible. Verified on Ubuntu Linux test host with OpenClaw `2026.3.13 (61d171a)`.

## Environment

- OS: Ubuntu Linux (kernel 6.8.0-107-generic, 22.04)
- Node.js: v22.22.1
- OpenClaw: 2026.3.13 (61d171a)
- create-openclaw-octo: 1.0.1-dev.6cb329f (@dev)

## Proposed Fix

1. Replace `ensureOpenClawCompat()` in `cli/utils.ts` with a wrapper that delegates to `detectOpenClawState()` from `cli/openclaw-cli.ts`. Exit with friendly message and `openclaw update` guidance when `kind === "block"`.
2. Remove duplicated `MIN_OPENCLAW_VERSION` constant and local `compareVersions()` helper from `cli/utils.ts`.
3. Update `OPENCLAW_PEER_MIN` to `"2026.3.22"` (technical hard floor — `clawhub:` protocol introduced).
4. Update `OPENCLAW_RECOMMENDED` to `"2026.5.22"` (PR #85298 `plugin-sdk-native-resolver.ts` — deterministic native plugin SDK subpath resolution + observed reliability improvement for ClawHub fetch behind corporate proxies).
5. Add unit test coverage for `ensureOpenClawCompat()` block / warn / ok paths.
6. Additionally, wrap `pluginsInstall()` to catch ClawHub timeout errors and emit a friendlier message guiding users to upgrade OpenClaw (see also issue for OpenClaw upstream re: 4.x ClawHub proxy/timeout reliability).

## Related

- Implementation plan: `octo-cli-implementation-plan.md` §11.5
- Upstream OpenClaw commit `91b2800241` ("feat: add native clawhub install flows") — introduces `clawhub:` protocol
- Upstream OpenClaw PR #85298 (`6b1c8687b5`) — `plugin-sdk-native-resolver.ts`
- Discovered during real-world environment testing (3-host validation matrix, May 2026)

caster-Q:
## Summary

Fixes the install entry point so it hard-aborts with friendly guidance on
truly incompatible OpenClaw versions, instead of running through 4 steps
of config mutation and dying at step 5 with a raw Node stack trace. Also
rewraps the ClawHub fetch timeout (observed on OpenClaw 4.x behind
corporate proxies) into actionable upgrade guidance, and adds a top-level
CLI catch so unhandled errors don't leak `Error.cause` (which can carry
upstream stderr) to terminal output.

## Related Issue

Fixes #90

## Changes

- **`cli/utils.ts`** — replace `ensureOpenClawCompat()` with a thin
  wrapper around the existing `detectOpenClawState()` from
  `openclaw-cli.ts`. Accepts `{ requireClawHubProtocol }` so callers
  that don't reach the `clawhub:` path (uninstall, `install --from
  <local-tarball>`, bind, etc.) get a warn rather than a hard abort.
  Drops the duplicated `MIN_OPENCLAW_VERSION` + local `compareVersions()`.
- **`cli/install.ts`** — `runInstall` passes
  `requireClawHubProtocol: !opts.from`. Local-tarball flows still work
  on pre-2026.3.22 OpenClaw.
- **`cli/openclaw-cli.ts`** —
  - `OPENCLAW_PEER_MIN`: `"2026.4.15"` → `"2026.3.22"` (the actual
    technical hard floor — first release with `clawhub:` protocol,
    upstream commit `91b2800241`).
  - `OPENCLAW_RECOMMENDED`: `"2026.5.18"` → `"2026.5.22"` (includes
    upstream PR `#85298` `plugin-sdk-native-resolver.ts` for
    deterministic resolution of `openclaw/plugin-sdk/*` subpath imports
    we rely on, plus observed reliability improvements for ClawHub
    fetch behind corporate proxies).
  - `pluginsInstall()`: catch `ClawHub request timed out after
    30000ms` (observed on OpenClaw 4.x behind corporate proxies — the
    internal dispatcher ignores `http_proxy` / `NODE_USE_ENV_PROXY`)
    and rewrap into a friendly `Error` pointing to `openclaw update`.
    Recognition runs **before** the captured-stderr replay so users
    only see the friendly message, not the raw timeout banner first.
    Original error preserved on `Error.cause`.
- **`cli/index.ts`** — switch from `program.parse()` to
  `parseAsync().catch(...)`. Top-level handler prints `err.message` (or
  full stack + cause when `OPENCLAW_OCTO_DEBUG=1`) then `process.exit(1)`.
  Without this, Node's default unhandled-rejection printer walks
  `Error.cause` and re-emits the upstream stderr — which can carry
  sensitive content.

## Testing

  - `cli/utils.test.ts`: 5 cases covering `ensureOpenClawCompat()`
    block-with-clawhub-required / warn-without-clawhub-required /
    block-when-missing (both with and without `requireClawHubProtocol`)
    / warn-below-recommended / silent-at-recommended.
  - `cli/openclaw-cli.test.ts`: timeout-rewrap on first attempt,
    timeout-rewrap on the bare-install fallback after 2 layers of
    unsupported-option degradation (with explicit assertion that the
    raw `ClawHub request timed out` stderr is **not** replayed before
    the friendly message), and pass-through for unrelated errors.
  - Existing `detectOpenClawState` boundary tests updated to the new
    `PEER_MIN` / `RECOMMENDED` values.
  - Total: **143 tests passing**, `tsc --noEmit` clean.
- [x] Manually verified
  - OpenClaw 2026.3.13 host (qiletest2): previously crashed at step 5
    with stack trace; now blocks at entry with friendly upgrade
    guidance. Uninstall on the same host stays available (warn, not
    block).
  - OpenClaw 2026.4.23 host behind corporate proxy (qiletest3):
    previously crashed at step 5 with raw `ClawHub request timed out
    after 30000ms`; now emits rewrapped upgrade guidance with no raw
    banner preceding it. After upgrading to OpenClaw 5.27, install
    runs through cleanly.
  - OpenClaw 2026.5.26+ host (qiletest1, mac, win): unchanged
    behaviour (silent / pass).

## Codex Review

The patch was reviewed by an external Codex pass before push. Findings
incorporated into this same commit (amended):

- **MAJOR #1**: top-level `parseAsync().catch(...)` added in `index.ts`
  to prevent `Error.cause` from leaking upstream stderr through Node's
  default unhandled-rejection printer.
- **MAJOR #2**: `requireClawHubProtocol` option added so the hard
  abort doesn't over-block uninstall / `install --from` paths that
  don't actually use the `clawhub:` protocol.
- **MINOR #3**: timeout recognition moved before the captured-stderr
  replay so users don't see the raw timeout banner first.
- **MINOR #5**: added a `last-attempt-timeout-after-degradation`
  test case.

Two findings deferred to follow-up issues (out of scope for this PR):

- **MINOR #4**: non-`ENOENT` `getOpenClawVersion` failures (permission
  errors, broken shims) currently fold into the "openclaw not found"
  branch. Will need a `detectOpenClawState()` semantic change to surface
  the probe failure reason — file separately.
- **Manifest mismatch**: `OPENCLAW_PEER_MIN = "2026.3.22"` is now stricter
  than `package.json` `peerDependencies.openclaw`. (Note: this package
  doesn't currently declare `peerDependencies.openclaw`; the prior
  assumption traced to the sibling plugin repo.) Worth re-aligning if /
  when we add a peer declaration here.

## Checklist

- [x] Updated documentation (no user-facing docs change in this PR;
      behavior is purely internal CLI ergonomics — no README update needed)

### DIFF
--- a/create-openclaw-octo/cli/index.ts
+++ b/create-openclaw-octo/cli/index.ts
@@ -14,7 +14,7 @@ import {
 } from "./doctor.js";
 import { runUninstall } from "./uninstall.js";
 import { runRemoveAccount } from "./remove-account.js";
-import { ensureOpenClawCompat, PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
+import { PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
 import { getOpenClawVersionStrict, resolvePluginState } from "./openclaw-cli.js";
 import { PLUGIN_VERSION } from "./version.js";
 
@@ -182,7 +182,36 @@ program
   });
 
 export function main(argv?: readonly string[]): void {
-  program.parse(argv);
+  // parseAsync + top-level catch so unhandled errors from any sub-command
+  // surface as a clean message instead of a Node stack trace. Two side
+  // benefits:
+  //   (a) `pluginsInstall()` rewraps ClawHub timeouts with friendly text
+  //       (issue #90) and stashes the raw error in `Error.cause`. Without
+  //       a top-level handler, Node's default unhandled-rejection printer
+  //       walks the cause chain and re-emits the raw upstream stderr —
+  //       which can carry sensitive content. Logging only `err.message`
+  //       keeps `cause` available to programmatic callers / debug runs
+  //       without leaking it to terminal output.
+  //   (b) Plain `Error` thrown from sub-commands no longer prints stack
+  //       frames in normal use; set `OPENCLAW_OCTO_DEBUG=1` to see the
+  //       full chain when debugging.
+  program.parseAsync(argv as string[] | undefined).catch((err: unknown) => {
+    // Defensively handle non-Error rejections (e.g. `Promise.reject(null)`,
+    // `Promise.reject("string")`). Without this guard, reading `err.message`
+    // / `err.cause` on a primitive would throw inside the handler.
+    if (err instanceof Error) {
+      const e = err as Error & { cause?: unknown };
+      if (process.env.OPENCLAW_OCTO_DEBUG) {
+        console.error(e.stack ?? e.message);
+        if (e.cause !== undefined) console.error("Caused by:", e.cause);
+      } else {
+        console.error(e.message);
+      }
+    } else {
+      console.error(String(err));
+    }
+    process.exit(1);
+  });
 }
 
 if (import.meta.url === `file://${process.argv[1]}`) {
--- a/create-openclaw-octo/cli/install.ts
+++ b/create-openclaw-octo/cli/install.ts
@@ -56,6 +56,7 @@ import {
   PLUGIN_ID,
   VERY_LEGACY_PLUGIN_ID,
   ensureOpenClawCompat,
+  requiresClawHubProtocol,
 } from "./utils.js";
 
 export function getLatestClawHubVersion(): string | null {
@@ -116,7 +117,15 @@ export interface InstallOptions {
 }
 
 export async function runInstall(opts: InstallOptions): Promise<void> {
-  ensureOpenClawCompat();
+  // The default install path dispatches via `openclaw plugins install
+  // clawhub:octo`, which requires the `clawhub:` protocol introduced in
+  // OpenClaw v2026.3.22. `--from` accepts arbitrary install specs (local
+  // tarball, file://, http(s), bare npm name, **and** `clawhub:...`); the
+  // version gate must consider the actual spec, not just whether `--from`
+  // was passed — otherwise `--from clawhub:octo` on older OpenClaw would
+  // sneak past the gate and only fail partway through migration.
+  // See issue #90 + PR #91 review.
+  ensureOpenClawCompat({ requireClawHubProtocol: requiresClawHubProtocol(opts.from) });
 
   // ---------------------------------------------------------------------------
   // v2.0.0 npm→ClawHub migration: detect (NOT yet uninstall) any pre-2.0 npm
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -543,6 +543,102 @@ describe("pluginsInstall 3-layer degradation", () => {
     expect(lastArgs).not.toContain("--force");
     expect(lastArgs).not.toContain("--dangerously-force-unsafe-install");
   });
+
+  it("should rewrap ClawHub timeout into a friendly error with upgrade guidance", async () => {
+    const { pluginsInstall } = await loadModule();
+    const upstream = new Error("Command failed: openclaw plugins install clawhub:octo\nClawHub request timed out after 30000ms");
+    (upstream as any).stderr = Buffer.from("ClawHub request timed out after 30000ms");
+    mockExecFileSync.mockImplementation(() => {
+      throw upstream;
+    });
+    let caught: unknown;
+    try {
+      pluginsInstall("clawhub:octo", true, true);
+    } catch (err) {
+      caught = err;
+    }
+    expect(caught).toBeDefined();
+    expect((caught as Error).message).toMatch(/ClawHub install of "clawhub:octo" timed out/);
+    expect((caught as Error).message).toMatch(/openclaw update/);
+    expect((caught as Error).message).toMatch(/2026\.5\.22/);
+    // Original error preserved on `cause` so callers / loggers can inspect it.
+    expect((caught as { cause?: unknown }).cause).toBe(upstream);
+    // No degradation — timeout is not a "unsupported option" signal.
+    expect(mockExecFileSync).toHaveBeenCalledTimes(1);
+  });
+
+  it("should pass through non-timeout, non-option errors unchanged", async () => {
+    const { pluginsInstall } = await loadModule();
+    const upstream = new Error("ENOENT: openclaw not found on PATH");
+    mockExecFileSync.mockImplementation(() => {
+      throw upstream;
+    });
+    expect(() => pluginsInstall("clawhub:octo", true)).toThrow("ENOENT: openclaw not found on PATH");
+  });
+
+  it("should rewrap ClawHub timeout even when it surfaces only on the bare-install fallback after degradation", async () => {
+    // Two unsupported-option errors degrade through the attempts list;
+    // the third (bare install) finally succeeds in reaching the network
+    // and times out. The rewrap must still trigger and the user must NOT
+    // see the raw "ClawHub request timed out" stderr first.
+    const { pluginsInstall } = await loadModule();
+    const stderrSpy = vi.spyOn(process.stderr, "write").mockImplementation(() => true);
+    const stdoutSpy = vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+    try {
+      mockExecFileSync
+        .mockImplementationOnce(() => {
+          const err = new Error("unknown option");
+          (err as any).stderr = Buffer.from(
+            "error: unknown option '--dangerously-force-unsafe-install'",
+          );
+          throw err;
+        })
+        .mockImplementationOnce(() => {
+          const err = new Error("unknown option");
+          (err as any).stderr = Buffer.from("error: unknown option '--force'");
+          throw err;
+        })
+        .mockImplementationOnce(() => {
+          const err = new Error(
+            "Command failed: openclaw plugins install clawhub:octo\nClawHub request timed out after 30000ms",
+          );
+          (err as any).stderr = Buffer.from(
+            "ClawHub request timed out after 30000ms\n",
+          );
+          (err as any).stdout = Buffer.from(
+            "Resolving clawhub:octo…\n",
+          );
+          throw err;
+        });
+
+      let caught: unknown;
+      try {
+        // quiet=false to verify replay-suppression for the timeout path.
+        pluginsInstall("clawhub:octo", false, true);
+      } catch (err) {
+        caught = err;
+      }
+
+      expect(mockExecFileSync).toHaveBeenCalledTimes(3);
+      expect((caught as Error).message).toMatch(/ClawHub install of "clawhub:octo" timed out/);
+      expect((caught as Error).message).toMatch(/openclaw update/);
+      expect((caught as Error).message).toMatch(/2026\.5\.22/);
+
+      // Captured stderr from the timeout (raw "ClawHub request timed out…")
+      // must NOT have been replayed to the user — that's the whole point
+      // of the rewrap.
+      const stderrCalls = stderrSpy.mock.calls.map((c) => String(c[0])).join("");
+      expect(stderrCalls).not.toMatch(/ClawHub request timed out/);
+
+      // Progress stdout from the timed-out attempt is still allowed
+      // through (informative, not the noisy banner we're suppressing).
+      const stdoutCalls = stdoutSpy.mock.calls.map((c) => String(c[0])).join("");
+      expect(stdoutCalls).toMatch(/Resolving clawhub:octo/);
+    } finally {
+      stderrSpy.mockRestore();
+      stdoutSpy.mockRestore();
+    }
+  });
 });
 
 // ---------------------------------------------------------------------------
@@ -797,19 +893,19 @@ describe("detectOpenClawState", () => {
     }
   });
 
-  it("block when version < OPENCLAW_PEER_MIN (2026.4.15)", async () => {
+  it("block when version < OPENCLAW_PEER_MIN (2026.3.22)", async () => {
     const { detectOpenClawState } = await loadModule();
-    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.3.10 (abc1234)\n" as any);
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.3.13 (abc1234)\n" as any);
     const state = detectOpenClawState();
     expect(state.kind).toBe("block");
     if (state.kind === "block") {
-      expect(state.version).toBe("2026.3.10");
-      expect(state.reason).toContain("2026.3.10");
-      expect(state.reason).toContain("2026.4.15");
+      expect(state.version).toBe("2026.3.13");
+      expect(state.reason).toContain("2026.3.13");
+      expect(state.reason).toContain("2026.3.22");
     }
   });
 
-  it("warn when peer-min <= version < recommended (2026.5.18)", async () => {
+  it("warn when peer-min <= version < recommended (2026.5.22)", async () => {
     const { detectOpenClawState } = await loadModule();
     mockExecFileSync.mockImplementation(() => "OpenClaw 2026.4.20 (abc1234)\n" as any);
     const state = detectOpenClawState();
@@ -817,24 +913,24 @@ describe("detectOpenClawState", () => {
     if (state.kind === "warn") {
       expect(state.version).toBe("2026.4.20");
       expect(state.reason).toContain("2026.4.20");
-      expect(state.reason).toContain("2026.5.18");
+      expect(state.reason).toContain("2026.5.22");
     }
   });
 
   it("warn at exact peer-min (boundary)", async () => {
     const { detectOpenClawState } = await loadModule();
-    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.4.15 (abc1234)\n" as any);
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.3.22 (abc1234)\n" as any);
     const state = detectOpenClawState();
     expect(state.kind).toBe("warn");
   });
 
   it("ok at exact recommended (boundary)", async () => {
     const { detectOpenClawState } = await loadModule();
-    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.5.18 (abc1234)\n" as any);
+    mockExecFileSync.mockImplementation(() => "OpenClaw 2026.5.22 (abc1234)\n" as any);
     const state = detectOpenClawState();
     expect(state.kind).toBe("ok");
     if (state.kind === "ok") {
-      expect(state.version).toBe("2026.5.18");
+      expect(state.version).toBe("2026.5.22");
     }
   });
 
--- a/create-openclaw-octo/cli/openclaw-cli.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.ts
@@ -302,6 +302,30 @@ function isUnsupportedOptionError(err: unknown): boolean {
   );
 }
 
+/**
+ * Check if an error indicates a ClawHub fetch timeout.
+ *
+ * Observed on OpenClaw 4.x behind corporate HTTP proxies: the internal
+ * ClawHub HTTP dispatcher does not honor `http_proxy` / `EnvHttpProxyAgent`,
+ * and the hard-coded 30s budget is not enough. Upstream message:
+ *
+ *   "ClawHub request timed out after 30000ms"
+ *
+ * We catch this in `pluginsInstall()` and rewrap with actionable guidance.
+ * 5.x reportedly fixes the underlying dispatcher; we recommend upgrading.
+ */
+function isClawHubTimeoutError(err: unknown): boolean {
+  const sources = [
+    (err as any)?.stderr?.toString?.(),
+    (err as any)?.stdout?.toString?.(),
+    (err as any)?.message,
+    String(err),
+  ];
+  return sources.some(
+    (s) => s && /ClawHub request timed out/i.test(s),
+  );
+}
+
 function isPluginNotInstalledError(err: unknown): boolean {
   const sources = [
     (err as any)?.stderr?.toString?.(),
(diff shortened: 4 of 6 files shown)


Jerry-Xin (dismissed):
The PR is relevant to `octo-adapters` and the implementation matches the stated goal: incompatible default ClawHub installs now fail early with cleaner guidance, and timeout errors are rewrapped before raw stderr is replayed.

💬 Non-blocking

🟡 Warning — [create-openclaw-octo/cli/install.ts](/Users/mlamp/.openclaw/tmp/pr-review-aswpvhnb/repo/create-openclaw-octo/cli/install.ts:124): `requireClawHubProtocol: !opts.from` treats every `--from` value as non-ClawHub, but `--from <spec>` is documented as accepting any OpenClaw install spec, including alternate specs. An explicit `install --from clawhub:octo` on pre-`2026.3.22` would bypass the new hard gate and can still fail later in the install path. Consider deriving the flag from the actual spec, for example requiring the protocol when `!opts.from || opts.from.startsWith("clawhub:")`, or narrowing the documented contract to local paths/tarballs only.

🔵 Suggestion — [create-openclaw-octo/cli/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-aswpvhnb/repo/create-openclaw-octo/cli/index.ts:198): the top-level catch assumes the rejection is object-like. It handles normal `Error` instances well, but `Promise.reject(null)` or similar would throw inside the handler while reading `e.message`. A tiny formatter around `err instanceof Error ? ... : String(err)` would make this fully robust.

✅ Highlights

🟢 The new `ensureOpenClawCompat()` delegates to the shared OpenClaw state detection instead of maintaining a second version comparator.

🟢 Timeout rewrapping in `pluginsInstall()` happens before stderr replay, which preserves the intended no-raw-banner behavior.

🟢 Tests cover the key regression paths: hard block, warning-only paths, timeout on first attempt, timeout after option degradation, and version boundary changes.

Verification note: I attempted `npm test -- --runInBand` and `npm run type-check` in `create-openclaw-octo`, but this checkout has no installed dependencies, so `vitest` and `tsc` were not available.

lml2468 (changes requested):
Blocking issue found:

- `create-openclaw-octo/cli/install.ts:124` sets `requireClawHubProtocol: !opts.from`, so every `--from` value bypasses the hard OpenClaw >= 2026.3.22 gate. But the same file documents `--from` as accepting arbitrary OpenClaw install specs, including `clawhub:...` (`install.ts:102-104`). A supported command such as `create-openclaw-octo install --from clawhub:octo` on OpenClaw < 2026.3.22 will only warn, then continue into the install/migration flow and eventually fail at `pluginsInstall(spec, ...)` with the unsupported `clawhub:` protocol. In migration/deadlock paths, that means config/plugin state is touched before the failure, which is the class of failure this PR is trying to hard-abort before. Please derive the gate from the resolved install spec instead, for example requiring the ClawHub protocol when `spec.startsWith("clawhub:")`, while still allowing local tarball/path `--from` installs on older OpenClaw.

lml2468 (dismissed):
[APPROVE]

Solid fix: consolidates version-gating into `detectOpenClawState()`, adds `requireClawHubProtocol` to avoid over-blocking uninstall/`--from` paths, rewraps ClawHub timeouts with actionable guidance, and plugs `Error.cause` leakage via a top-level `parseAsync` catch.

✅ Highlights

- Single source of truth for version policy — eliminates duplicated `MIN_OPENCLAW_VERSION` + local `compareVersions()` in `utils.ts`.
- `requireClawHubProtocol` option is a clean design: callers declare intent, the gate decides severity.
- Timeout recognition runs *before* stderr replay — users see the friendly message first, not the raw banner.
- `Error.cause` preserved for programmatic/debug callers but suppressed from terminal output — good security hygiene.
- Thorough test coverage: 5 new `ensureOpenClawCompat` cases, 3 timeout-rewrap cases (including the degradation-then-timeout scenario), boundary tests updated.

🔵 Suggestion (non-blocking)

1. **`index.ts:17` — unused import `ensureOpenClawCompat`**: this symbol is imported but never referenced in the file body. Pre-existing, but since the file is touched in this PR it's a good opportunity to clean up:
   ```ts
   import { PLUGIN_ID, renderInstallStatusBanner } from "./utils.js";
   ```

yujiawei (dismissed):
# Code Review — PR #91 (octo-adapters)

Independent review of the install hard-abort restructure on top of `b26e6ef`.

## Summary

The patch is mostly the right shape: delegating to `detectOpenClawState()` is a clean simplification, the `requireClawHubProtocol` opt-in correctly preserves uninstall / local-tarball flows on pre-2026.3.22 hosts, the timeout rewrap is well placed before the captured-stderr replay so the friendly message is what the user sees first, and the top-level `parseAsync().catch(...)` is a real defense against `Error.cause` leaking upstream stderr through Node's default unhandled-rejection printer. Tests are thorough.

There is one correctness gap that recreates the exact failure mode this PR is supposed to prevent.

## P1 — `--from clawhub:<spec>` on old OpenClaw still bypasses the hard gate

`create-openclaw-octo/cli/install.ts:124`:

```ts
ensureOpenClawCompat({ requireClawHubProtocol: !opts.from });
```

The gate is keyed on *whether `--from` was passed*, not on *whether the resolved install spec actually requires the `clawhub:` protocol*. But `--from` is documented (`install.ts:102-115`, and the CLI `--from <spec>` description at `cli/index.ts:78-79`) as accepting "anything OpenClaw's plugins-install accepts: a tarball path, a directory path, or any alternate spec (clawhub:..., openclaw-..., etc.)".

So `create-openclaw-octo install --from clawhub:octo@2.0.0-rc.1` on OpenClaw `< 2026.3.22`:

1. `ensureOpenClawCompat` takes the `block + !requireClawHubProtocol` branch and only prints a warning (`utils.ts:139-146`) that explicitly tells the user "Local-tarball install (--from) … remain available" — which is actively misleading here, since the user's `--from` value *is* a `clawhub:` spec.
2. Control falls through to `detectScenario()`, then to `pluginsInstall(spec, ...)` at `install.ts:237` (`broken`), `install.ts:248` (`fresh`), `install.ts:192/227` (`update`), `install.ts:672` (`deadlock`), or `runMigration` step 5 (`install.ts:445`).
3. `pluginsInstall("clawhub:…", …)` then crashes with `unsupported npm spec: protocol specs are not allowed` — which is the exact step-5 failure issue #90 is closing.

Impact across scenarios:
- `fresh`: just the same old crash, no state damage.
- `broken`: `cleanupBrokenInstall(PLUGIN_ID)` at `install.ts:234` mutates config *before* the doomed `pluginsInstall`. No backup, no rollback — this is the regression class the PR is explicitly trying to eliminate.
- `legacy-to-octo` / `rebrand` / `deadlock`: rollback paths exist (`runMigration` step 0 backup → `rollback()`, `runDeadlockRepair` `.deadlock-backup`), so data is restored, but the user still pays the migration cost plus a misleading warning before hitting the same crash the PR was supposed to hard-abort at entry.

Suggested fix: derive `requireClawHubProtocol` from the resolved spec instead of from `opts.from`'s presence. Pull the `const spec = opts.from ?? CLAWHUB_INSTALL_SPEC;` resolution above the gate and use, e.g.:

```ts
const spec = opts.from ?? CLAWHUB_INSTALL_SPEC;
ensureOpenClawCompat({ requireClawHubProtocol: spec.startsWith("clawhub:") });
```

This keeps local-tarball / directory `--from` paths unblocked on old OpenClaw (the explicit goal of the option) while still hard-aborting any path that ends up dispatching through `clawhub:`. A regression test against `runInstall({ from: "clawhub:octo@x" })` on a mocked old runtime would lock the behaviour in.

## P2 / nits (non-blocking)

- **`ensureOpenClawCompat()` UX inconsistency**: the block-with-clawhub-required path uses plain `console.error([...].join("\n"))` (`utils.ts:120-134`), while `enforceHealthyClawHubInstall()` consistently uses the boxed `printUpgradeNotice(...)` helper a few lines down. Both are user-facing "✗ stop, upgrade" messages on the same surface; using `printUpgradeNotice({ status: "block", ... })` here too would give a single visual language for hard-stops.
- **Exit code asymmetry**: missing-OpenClaw exits `1`, too-old-OpenClaw exits `2` (`utils.ts:118` vs `134`). Not wrong, but unmotivated — and the distinction isn't documented. Either pick one or add a one-line comment on what callers should branch on.
- **Timeout regex breadth**: `isClawHubTimeoutError` only matches `/ClawHub request timed out/i` (`openclaw-cli.ts:325`). If upstream ever changes the wording (e.g. `ClawHub fetch timed out`, the underlying `EnvHttpProxyAgent` path could surface a slightly different banner once the 5.x dispatcher work lands), the rewrap silently falls back to the raw stack trace. Worth either pinning this with a small regression test against the literal upstream message string, or broadening to `/ClawHub.*timed out|ClawHub.*timeout/i` so wording drift is tolerated.
- **`tagLabel` only attached to direct install/update log lines** in `install.ts` (`Installing Octo plugin${tagLabel}…`, etc.), but the `legacy-to-octo` / `rebrand` / `deadlock` paths also internally call `pluginsInstall(spec, …)` and silently drop the label. When debugging a `--from`-driven migration run this means the operator can't tell from the logs which spec was actually used. Trivial: thread `tagLabel` through `runMigration` and `runDeadlockRepair`.
- **Acknowledged manifest mismatch** (`OPENCLAW_PEER_MIN = "2026.3.22"` vs absent `peerDependencies.openclaw`): the PR description already calls this out as out-of-scope; flagging here only so it isn't forgotten when the peer declaration is reintroduced.

## What looks good

- `parseAsync().catch(...)` in `cli/index.ts:198-207` plus `OPENCLAW_OCTO_DEBUG` gate on the cause-chain print is exactly the right shape — `err.message` to terminal, full chain to debug builds, no leakage of upstream stderr by default.
- Timeout-recognition-before-stderr-replay ordering in `pluginsInstall` (`openclaw-cli.ts:382-411`) plus the explicit regression test (`openclaw-cli.test.ts:579-641`) is good defense against re-introducing the raw banner.
- Test coverage of `ensureOpenClawCompat` correctly distinguishes the new `requireClawHubProtocol: true` (hard exit) vs default (warn-and-continue) paths and includes the boundary case at exactly `OPENCLAW_PEER_MIN`.
- `detectInstallState` fail-closed on real `plugins list --json` failures (corruption / EACCES / parse) is the right call — trusting `cfg.entries` alone in that state would let dual-active situations slip past the pre-flight.

Net: structure and intent are right; please tighten the `requireClawHubProtocol` gate so the documented `--from clawhub:<spec>` path doesn't reintroduce the half-mutated-state failure mode this PR is closing.

caster-Q:
Thanks for the review @Jerry-Xin @lml2468 — addressed in `0e236f3`:

**Blocking gate-bypass fix** (both reviewers flagged this)
- Added `requiresClawHubProtocol(installSpec)` helper in `utils.ts` and switched `install.ts` to use it.
- The gate is now derived from the resolved install spec, not just `!opts.from`:
  - default install / `--from clawhub:...` → `requireClawHubProtocol: true` (hard floor enforced)
  - `--from /local.tgz` / `--from file://...` / `--from openclaw-channel-octo` → `false` (warn-only, still works on older OpenClaw)
- 5 new unit tests cover the helper across all these spec shapes.

**Top-level catch robustness** (@Jerry-Xin)
- Guarded with `err instanceof Error` so `Promise.reject(null)` / string rejections fall through to `String(err)` instead of throwing inside the handler.

**Unused import** (@lml2468)
- Dropped `ensureOpenClawCompat` from `cli/index.ts` imports.

Tests: **148 passing**, `tsc --noEmit` clean.

lml2468 (approved):
Re-review of 0e236f388da5 — previous CHANGES_REQUESTED at b26e6ef resolved.

**Previous P0 (resolved):** `--from` bypassed ClawHub protocol gate unconditionally.

**Fix verified:**
- New `requiresClawHubProtocol(spec)` correctly derives gate from actual install spec: `clawhub:*` → hard gate, local tarball/file/npm → soft warn.
- `ensureOpenClawCompat({ requireClawHubProtocol })` now distinguishes "openclaw missing" (always exit 1) from "openclaw too old" (exit 2 if clawhub needed, warn-only otherwise).
- `install.ts:124` passes `requiresClawHubProtocol(opts.from)` — the exact fix requested.
- Uninstall / `--from <tarball>` paths correctly default to `requireClawHubProtocol: false` (no regression).
- Version constants updated: `OPENCLAW_PEER_MIN 2026.3.22` (clawhub: protocol introduction), `OPENCLAW_RECOMMENDED 2026.5.22` (plugin-sdk-native-resolver fix).
- `index.ts` top-level error handler: `parseAsync().catch()` with `instanceof Error` guard and `OPENCLAW_OCTO_DEBUG` env var — prevents raw stack/cause leaking.
- ClawHub timeout rewrap in `pluginsInstall()` runs before stderr replay — clean UX.
- Test coverage: 6 tests for `requiresClawHubProtocol`, 6 tests for `ensureOpenClawCompat` (including regression for the warn-not-block path), 3 tests for timeout rewrap including degradation-then-timeout scenario.

LGTM.

Jerry-Xin (approved):
The PR is in scope for `octo-adapters` and the changes correctly hard-gate ClawHub installs while preserving non-ClawHub flows.

✅ Highlights
- 🔵 `create-openclaw-octo/cli/install.ts:128` correctly gates based on the actual install spec, including `--from clawhub:...`.
- 🔵 `create-openclaw-octo/cli/utils.ts:123` centralizes compatibility behavior through `detectOpenClawState()` and keeps uninstall/local tarball paths usable on older OpenClaw versions.
- 🔵 `create-openclaw-octo/cli/openclaw-cli.ts:382` rewraps ClawHub timeout errors before stderr replay, avoiding the raw timeout banner.
- 🔵 `create-openclaw-octo/cli/index.ts:198` adds a top-level async catch that prevents `Error.cause` from being printed in normal CLI output.

Verification performed:
- `npm run type-check` passed.
- `npm test` passed: 9 files, 148 tests.

caster-Q:
@yujiawei thanks for the thorough independent review — really useful that you traced the gate-bypass through every scenario (`fresh` / `broken` / `legacy-to-octo` / `rebrand` / `deadlock`), the `broken` path in particular was a blind spot.

Your **P1 (gate-bypass on `--from clawhub:…`)** is addressed in `0e236f3` (force-pushed before your review landed):

- New helper `requiresClawHubProtocol(installSpec)` in `cli/utils.ts` keys the gate on the resolved spec, not on `!opts.from`.
- `cli/install.ts` calls `ensureOpenClawCompat({ requireClawHubProtocol: requiresClawHubProtocol(opts.from) })`.
- 5 new unit tests in `cli/utils.test.ts` lock the matrix you described: default install / `clawhub:*` / local tarball / `file://` / bare npm name.

So `create-openclaw-octo install --from clawhub:octo@2.0.0-rc.1` on OpenClaw `< 2026.3.22` now hard-aborts at entry as expected, before `cleanupBrokenInstall(PLUGIN_ID)` in the `broken` path can touch state.

On your **P2 nits**, my read:

- **(1) Use `printUpgradeNotice(...)` for visual consistency** — agreed, makes sense. Happy to fold in this PR.
- **(2) Exit code 1 vs 2 asymmetry** — fair, I'll either unify or add a one-line comment on intent.
- **(3) Broaden timeout regex / add literal-string fixture** — agreed, the wording-drift risk is real; will broaden to `/ClawHub.*time(d)?\s*out/i` style and add a regression fixture against the exact upstream banner.
- **(4) Thread `tagLabel` through `runMigration` / `runDeadlockRepair`** — agreed it's useful for debugging `--from` runs, but it touches a few signatures and feels like its own focused PR. Happy to file a follow-up issue and link it here.
- **(5) Manifest mismatch** — already acknowledged as follow-up in the PR body; will track when we reintroduce a `peerDependencies.openclaw` declaration.

**Question for you**: would you prefer I address P2 (1)/(2)/(3) in this PR (adds ~30 lines + 1-2 tests, no scope creep beyond your review), or merge `0e236f3` as-is for the P1 fix and handle the nits in a focused follow-up? Either works for me — your call.

lml2468 (approved):
Re-review of 5ec6504fb327 — incremental from 0e236f388da5 (previously APPROVED).

**Delta:** Only change beyond the already-approved code is `.github/workflows/octo-pr-review-feed.yml` removing `opened` and `synchronize` triggers (keeps `ready_for_review` + `review_requested`). This reduces redundant workflow runs — no functional impact on the codebase.

All code changes (requiresClawHubProtocol, ensureOpenClawCompat, timeout rewrap, top-level error handler, tests) are identical to the previously approved commit. LGTM.

Jerry-Xin (approved):
No blocking issues found; the PR is relevant to `create-openclaw-octo` and addresses the install compatibility/error-reporting path cleanly.

💬 Non-blocking
- 🔵 Suggestion: Consider adding a focused test for the top-level `parseAsync().catch(...)` path to lock in the “normal mode prints `err.message` only, debug mode prints cause” behavior. This is security-relevant behavior and currently lives in [index.ts](/Users/mlamp/.openclaw/tmp/pr-review-bsf9if0m/repo/create-openclaw-octo/cli/index.ts:198).

✅ Highlights
- The compatibility gate now uses the actual install spec, so default `clawhub:octo` and explicit `--from clawhub:...` hard-block on incompatible OpenClaw versions, while local tarball flows stay available. See [install.ts](/Users/mlamp/.openclaw/tmp/pr-review-bsf9if0m/repo/create-openclaw-octo/cli/install.ts:128) and [utils.ts](/Users/mlamp/.openclaw/tmp/pr-review-bsf9if0m/repo/create-openclaw-octo/cli/utils.ts:94).
- Timeout rewrapping runs before captured stderr replay, preventing the raw ClawHub timeout banner from being emitted first. See [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-bsf9if0m/repo/create-openclaw-octo/cli/openclaw-cli.ts:382).
- Version policy is centralized through `detectOpenClawState()` and the updated `OPENCLAW_PEER_MIN` / `OPENCLAW_RECOMMENDED` constants. See [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-bsf9if0m/repo/create-openclaw-octo/cli/openclaw-cli.ts:1459).
- Added tests cover the compatibility gate, timeout rewrap, degradation fallback timeout case, and version boundaries.

Verification note: I attempted the focused tests and type check, but this checkout does not have dependencies installed, so `vitest` and `tsc` were unavailable.

PR #97, opened 2026-05-29
fix(create-openclaw-octo): prune stale npm package.json residue after npm-legacy uninstall

### LINKED ISSUE #94: fix(create-openclaw-octo): npm root package.json residue after npm-legacy install can resurrect old version weeks later
## Description

After a successful upgrade from the legacy npm-installed package (`openclaw-channel-octo@1.0.0`, Path B in our migration matrix), the user's `~/.openclaw/npm/package.json` still contains a stale `"openclaw-channel-octo": "^1.0.0"` dependency entry — even though the actual `node_modules/openclaw-channel-octo/` directory has been removed by `openclaw plugins uninstall`.

This is silent at upgrade time. Days or weeks later, if anything causes OpenClaw to re-resolve dependencies from that manifest (a stray `npm install` in `~/.openclaw/npm`, a doctor repair pass, a host-side plugin pull, etc.), the **old 1.0.0 plugin reappears in `node_modules`** and gets re-registered. The user sees ghost behaviour from the previous version after they thought the upgrade was complete.

## Steps to Reproduce

1. Install `openclaw-channel-octo@1.0.0` via `openclaw plugins install openclaw-channel-octo` (Path B / npm-legacy install layout — leaves a `~/.openclaw/npm/` tree with `package.json` listing the plugin).
2. Run `npx -y create-openclaw-octo install` to migrate to v2.x ClawHub plugin (`octo`).
3. Verify install reports success and `openclaw plugins list` shows the new `octo` plugin.
4. Inspect `~/.openclaw/npm/package.json`:
   ```
   {
     "dependencies": {
       "openclaw-channel-octo": "^1.0.0"
     }
   }
   ```
   The entry is still there.
5. Trigger any flow that re-runs `npm install` against that manifest. The old plugin's `node_modules/openclaw-channel-octo/` reappears.

## Expected Behavior

After a successful npm-legacy install (the path where we replace `openclaw-channel-octo@1.0.0` with the ClawHub `octo` plugin), `~/.openclaw/npm/package.json` should no longer list `openclaw-channel-octo` in `dependencies`.

## Actual Behavior

Entry persists. `node_modules/openclaw-channel-octo/` was removed, but the manifest still claims the package as a dependency, so any subsequent dep resolution restores it.

## Adapter Version

`create-openclaw-octo@1.0.1-dev.c5a5ad3` (current `@dev` after PR #91 merge). Behaviour pre-dates PR #91.

## Affected Component

Claude Code Adapter (`create-openclaw-octo` CLI) — specifically the npm-legacy uninstall step in `cli/install.ts`.

## Severity

P1 — high. Silent at upgrade time, but resurrection is hard to diagnose later because it looks like the user "downgraded" out of nowhere.

## Reproducibility

Consistently reproducible. Verified on the Windows real-world test host (`openclaw-channel-octo@1.0.0` → `octo@1.0.13` migration).

## Environment

- OS: Windows 10/11 (also reproducible on macOS/Linux Path B installs)
- Node.js: v22.x
- OpenClaw: 2026.5.x
- Adapter layout: npm-only (no `~/.openclaw/extensions/openclaw-channel-octo`)

## Proposed Fix

In `cli/install.ts`, in the npm-legacy uninstall branch (where we currently call `pluginsUninstall(NPM_PACKAGE_NAME)`):

1. Read `~/.openclaw/npm/package.json` (using `getOpenClawNpmDir()` or equivalent).
2. If `dependencies.openclaw-channel-octo` exists, delete that key.
3. Write the modified `package.json` back atomically.
4. Add a unit test covering both "entry present → removed" and "entry absent → no-op".

Optionally also handle `devDependencies` / `peerDependencies` for defence in depth, though we've only observed the residue under `dependencies`.

## Related

- Discovered during three-host real-world testing (May 2026).
- Adjacent to (but separate from) the broader "Bug #2: OpenClaw uninstall does not delete the extensions/ directory" upstream issue, which we intentionally do not fix on our side.

caster-Q:
## Summary

Fixes a latent corruption mode where a Path B (npm-legacy) upgrade succeeds visibly but leaves a stale dependency entry in `~/.openclaw/npm/package.json`. Any later `npm install` against that manifest resurrects the v1.0.0 plugin in `node_modules/`, OpenClaw discovers and re-registers it at the next gateway restart, and the user sees the "ghost" of an old version weeks after they thought the upgrade was complete.

## Related Issue

Fixes #94

## Changes

- **`cli/openclaw-cli.ts`** — add `cleanNpmPackageJsonResidue(packageName)`. Removes the named entry from `dependencies` / `devDependencies` / `peerDependencies`, drops a section that becomes empty, atomic-writes the manifest back (`<path>.tmp` + rename). Failure modes are split:
  - **read-side** (missing file / malformed JSON) → silent no-op. Not actionable for the caller; these aren't the residue case.
  - **write-side** (permission denied / rename failure) → `console.warn` so the user knows the residue is still in place. Main flow never throws — the broader install/uninstall has already succeeded.
- **`cli/install.ts`** — invoke the helper in the npm-legacy uninstall block, immediately after `pluginsUninstall(NPM_PACKAGE_NAME, true)` succeeds and before the saved channels/bindings restore.
- **`cli/openclaw-cli.test.ts`** — 7 unit cases: present in deps, present in dev + peer, now-empty section deletion, package absent (no write), manifest missing (no write), malformed JSON (no write/no throw), write failure (warn but no throw).
- **`cli/install.test.ts`** — end-to-end coverage that `runInstall()` actually invokes the cleanup. Locks the call site so it can't be silently dropped without a failing test (Codex review MINOR #3).

## Testing

- [x] Unit tests added/updated — `cli/openclaw-cli.test.ts` (7 cases for `cleanNpmPackageJsonResidue`), `cli/install.test.ts` (end-to-end assertion in the legacy npm cleanup scenario).
- [x] Manually verified — Path B (Windows) test host: pre-fix the residue persisted after `pluginsUninstall`; with this fix the entry is removed alongside the directory removal.
- Total: **156 tests passing**, `tsc --noEmit` clean.

## Codex Review

The patch was reviewed by an external Codex pass before push. Findings on `84898ce` (the initial commit):

- **MINOR #2** (silent IO failure on write): addressed in the amended commit `f985e61` — write-side failures now `console.warn` so the user knows the residue persists.
- **MINOR #3** (no end-to-end assertion): addressed in the amended commit — `install.test.ts` now asserts the cleanup runs from within `runInstall()`.
- **MINOR #1** (concurrent-write race on the same manifest): not addressed in this PR. Behavior matches the existing `writeConfigAtomic` pattern in the same file — race-hardening would touch all atomic-write call sites and belongs in its own focused PR.

## Checklist

- [x] Updated documentation (no user-facing docs change — internal cleanup helper, no behavioural surface for end users beyond the absent ghost-version reappearance)

### DIFF
--- a/create-openclaw-octo/cli/install.test.ts
+++ b/create-openclaw-octo/cli/install.test.ts
@@ -483,6 +483,109 @@ describe("runInstall — update scenario", () => {
     expect(didCallGatewayRestart(calls)).toBe(true);
   });
 
+  it("legacy npm cleanup also prunes the package from ~/.openclaw/npm/package.json (issue #94)", async () => {
+    // End-to-end coverage that `runInstall()` actually invokes the
+    // manifest-residue cleanup after a successful legacy npm uninstall.
+    // Without this assertion, the helper unit tests would keep passing
+    // even if the call site in install.ts were dropped.
+    const fs = await import("node:fs");
+    const mockReadFileSync = vi.mocked(fs.readFileSync);
+    const mockExistsSync = vi.mocked(fs.existsSync);
+    const mockWriteFileSync = vi.mocked(fs.writeFileSync);
+    const mockRenameSync = vi.mocked(fs.renameSync);
+
+    mockReadFileSync.mockImplementation((path) => {
+      const p = String(path);
+      if (p.endsWith("openclaw.json")) {
+        return JSON.stringify({
+          plugins: {
+            entries: {
+              "openclaw-channel-octo": { enabled: true },
+              octo: { enabled: true },
+            },
+            installs: {
+              "openclaw-channel-octo": { source: "npm", version: "1.0.0" },
+              octo: { source: "clawhub", version: "1.0.7" },
+            },
+            allow: ["openclaw-channel-octo", "octo"],
+          },
+        });
+      }
+      if (p.endsWith("package.json")) {
+        // Mirror the manifest shape OpenClaw leaves behind after a Path B
+        // (npm-legacy) install — `node_modules/openclaw-channel-octo/` has
+        // been removed but the dependency entry persists.
+        return JSON.stringify({
+          name: "openclaw-managed",
+          private: true,
+          dependencies: {
+            "openclaw-channel-octo": "^1.0.0",
+          },
+        });
+      }
+      return "{}";
+    });
+    // Both the ClawHub plugin extension dir AND the npm/package.json need to
+    // appear to exist so isHealthyInstall("octo") returns true and the
+    // residue helper proceeds past its existsSync guard.
+    mockExistsSync.mockImplementation((path) => {
+      return (
+        pathEndsWith(path, "extensions/octo") ||
+        String(path).endsWith("package.json")
+      );
+    });
+
+    const { runInstall } = await loadInstall();
+
+    mockExecFileSync.mockImplementation((_cmd, args) => {
+      const a = args as string[];
+      if (a[0] === "config" && a[1] === "file") return "/home/user/.openclaw/openclaw.json";
+      if (a[0] === "--version") return "OpenClaw 2026.5.7\n";
+      if (a[0] === "plugins" && a[1] === "inspect") {
+        const id = a[2];
+        if (id === "openclaw-channel-octo") {
+          return JSON.stringify({
+            plugin: { id: "openclaw-channel-octo", version: "1.0.0", enabled: true },
+          });
+        }
+        return inspectDispatch(a, {
+          id: "octo",
+          version: "1.0.7",
+          latestVersion: "1.0.7",
+          enabled: true,
+        });
+      }
+      if (a[0] === "plugins" && (a[1] === "uninstall" || a[1] === "disable")) return "";
+      if (a[0] === "gateway" && a[1] === "restart") return "";
+      return "";
+    });
+
+    await runInstall({ force: false, dev: false });
+
+    // The cleaned manifest must be staged to a `.tmp` sibling and then
+    // promoted via rename — this is the atomic-write contract of
+    // cleanNpmPackageJsonResidue.
+    const manifestTmpWrites = mockWriteFileSync.mock.calls.filter((c) =>
+      String(c[0]).endsWith("package.json.tmp"),
+    );
+    expect(manifestTmpWrites.length).toBeGreaterThan(0);
+
+    const writtenManifest = JSON.parse(String(manifestTmpWrites[0][1]));
+    // The stale dep entry must be gone. Other top-level fields (name,
+    // private) must survive; the now-empty dependencies section may either
+    // be dropped or present-but-empty depending on helper internals — we
+    // assert the contract that matters (the entry is gone).
+    expect(writtenManifest.dependencies?.["openclaw-channel-octo"]).toBeUndefined();
+    expect(writtenManifest.name).toBe("openclaw-managed");
+    expect(writtenManifest.private).toBe(true);
+
+    // And the tmp file must have been promoted into place.
+    const manifestRenames = mockRenameSync.mock.calls.filter((c) =>
+      String(c[0]).endsWith("package.json.tmp") && String(c[1]).endsWith("package.json"),
+    );
+    expect(manifestRenames.length).toBeGreaterThan(0);
+  });
+
   it("legacy npm cleanup preserves channels.octo + bindings(channel=octo) (P0-2 regression)", async () => {
     // PR #37 review P0-2: `openclaw plugins uninstall openclaw-channel-octo`
     // ALSO removes channels.octo as a side effect (both plugins own channel
--- a/create-openclaw-octo/cli/install.ts
+++ b/create-openclaw-octo/cli/install.ts
@@ -36,6 +36,7 @@ import {
   pluginsInspect,
   pluginsInstall,
   pluginsUninstall,
+  cleanNpmPackageJsonResidue,
   pluginsUpdate,
   readConfigFromFile,
   removeBindingsFromFile,
@@ -294,6 +295,14 @@ export async function runInstall(opts: InstallOptions): Promise<void> {
       }
       try {
         pluginsUninstall(NPM_PACKAGE_NAME, true);
+
+        // OpenClaw's uninstall removes the node_modules/<pkg>/ directory but
+        // leaves a stale `dependencies."openclaw-channel-octo": "^1.0.0"` entry
+        // in `~/.openclaw/npm/package.json`. Any later `npm install` against
+        // that manifest (doctor repair, manual command, host pull) would
+        // resurrect the v1.0.0 plugin weeks after this upgrade — see #94.
+        // Prune the manifest entry now, while we still own the migration.
+        cleanNpmPackageJsonResidue(NPM_PACKAGE_NAME);
 
         // RESTORE channels.octo + bindings (openclaw uninstall removed them as
         // a side effect; we kept a snapshot above)
--- a/create-openclaw-octo/cli/openclaw-cli.test.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.test.ts
@@ -1587,3 +1587,165 @@ describe("getConfigFilePathSafe (Windows relative path)", () => {
     expect(result).toBe(RESOLVED_CFG_PATH);
   });
 });
+
+// ---------------------------------------------------------------------------
+// cleanNpmPackageJsonResidue — issue #94 regression coverage
+//
+// OpenClaw's `plugins uninstall` removes ~/.openclaw/npm/node_modules/<pkg>/
+// but leaves the entry under ~/.openclaw/npm/package.json `dependencies`,
+// which lets a later `npm install` resurrect the old plugin. This helper
+// prunes the manifest entry; here we lock down the cases that matter.
+// ---------------------------------------------------------------------------
+
+describe("cleanNpmPackageJsonResidue", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  function setupManifest(initial: Record<string, any> | null): {
+    existsSync: any;
+    readFileSync: any;
+    writeFileSync: any;
+    renameSync: any;
+  } {
+    return {
+      existsSync: vi.fn((p: any) => {
+        const path = String(p);
+        // openclaw.json absent → getConfigFilePathSafe falls back to
+        // <homedir>/.openclaw/openclaw.json; the npm/package.json sibling is
+        // what we toggle on `initial !== null`.
+        if (path.endsWith("package.json")) return initial !== null;
+        return false;
+      }),
+      readFileSync: vi.fn(() => JSON.stringify(initial ?? {})),
+      writeFileSync: vi.fn(),
+      renameSync: vi.fn(),
+    };
+  }
+
+  async function loadWithMocks(initial: Record<string, any> | null) {
+    const mocks = setupManifest(initial);
+    vi.resetModules();
+    vi.doMock("node:fs", async (importOriginal) => {
+      const actual = await importOriginal<typeof import("node:fs")>();
+      return {
+        ...actual,
+        existsSync: mocks.existsSync,
+        readFileSync: mocks.readFileSync,
+        writeFileSync: mocks.writeFileSync,
+        copyFileSync: vi.fn(),
+        renameSync: mocks.renameSync,
+      };
+    });
+    const mod = await import("./openclaw-cli.js");
+    return { mod, ...mocks };
+  }
+
+  it("removes the package from dependencies and writes the manifest atomically", async () => {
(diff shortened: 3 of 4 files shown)


lml2468 (approved):
[APPROVE]

Clean fix for a real latent corruption mode — stale npm manifest entry resurrection.

✅ Highlights

- **Two-phase error handling is exactly right**: read-side failures (missing/malformed manifest) are silent no-ops; write-side failures (EACCES, rename) surface a `console.warn` with actionable guidance. The caller's main flow never throws — correct, because the broader install has already succeeded.
- **Atomic write** (`writeFileSync` to `.tmp` + `renameSync`) matches the existing `writeConfigAtomic` pattern in the same file — consistent and crash-safe.
- **Empty-section pruning**: deleting a now-empty `dependencies` object keeps the manifest tidy. Good detail.
- **Call-site coverage**: the end-to-end test in `install.test.ts` locks the call site so the cleanup can't be silently dropped without a failing test. This addresses the exact gap that codex review MINOR #3 identified.
- **Correct scoping**: cleanup lives in `install.ts` (the npm-legacy migration path), not `uninstall.ts` — uninstall deals with `PLUGIN_ID` ("octo"), not the legacy `NPM_PACKAGE_NAME` ("openclaw-channel-octo").
- Deferring the concurrent-write race (codex MINOR #1) to a follow-up is the right call — it's a cross-cutting concern across all atomic-write call sites.

Jerry-Xin (approved):
The PR is relevant to `octo-adapters` and correctly addresses the stale npm manifest residue path with focused install wiring and regression tests.

💬 Non-blocking

🟡 Warning: [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-41j0bjf7/repo/create-openclaw-octo/cli/openclaw-cli.ts:985) parses JSON but does not verify the result is a non-null plain object before [line 992](/Users/mlamp/.openclaw/tmp/pr-review-41j0bjf7/repo/create-openclaw-octo/cli/openclaw-cli.ts:992). A valid-but-invalid manifest like `null` would throw outside the read-side catch and could turn this best-effort cleanup into a migration failure. Consider guarding with `if (!pkg || typeof pkg !== "object" || Array.isArray(pkg)) return;`.

🔵 Suggestion: [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-41j0bjf7/repo/create-openclaw-octo/cli/openclaw-cli.ts:1008) and [openclaw-cli.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-41j0bjf7/repo/create-openclaw-octo/cli/openclaw-cli.test.ts:1726) include review-process provenance in comments. These are not needed for maintainability; referencing the issue or the behavior is enough.

✅ Highlights

The cleanup is called only after the legacy uninstall succeeds, which matches the intended migration ordering.

Write failures are surfaced with `console.warn` while preserving the broader install flow.

The new unit tests cover the important manifest cases, and `install.test.ts` locks the actual `runInstall()` call site.

I could not run the targeted tests locally because `create-openclaw-octo/node_modules` is missing and `vitest` is not installed in this checkout.

yujiawei (approved):
# Code Review — PR #97 (octo-adapters)

## Verdict

**APPROVED** — well-scoped fix for #94 with a clean helper, correct call-site placement, and comprehensive test coverage. No correctness or safety blockers. A couple of P2 documentation/consistency nits that can be cleaned up in a follow-up.

## 1. What was reviewed

- `create-openclaw-octo/cli/openclaw-cli.ts` — new `cleanNpmPackageJsonResidue(packageName)` helper (`openclaw-cli.ts:970-1022`).
- `create-openclaw-octo/cli/install.ts` — invocation in the npm-legacy uninstall branch (`install.ts:296-305`).
- `create-openclaw-octo/cli/openclaw-cli.test.ts` — 7 unit cases for the helper.
- `create-openclaw-octo/cli/install.test.ts` — end-to-end coverage that `runInstall()` invokes the cleanup.

## 2. Verification results

| Item | Status | Evidence |
|---|---|---|
| Bug reproduces against issue #94 description | ✅ | Issue #94 spells out the manifest-residue path; the helper targets exactly that path (`~/.openclaw/npm/package.json` `dependencies`). |
| Call site is the right place | ✅ | `install.ts:297-305` — runs immediately after `pluginsUninstall(NPM_PACKAGE_NAME, true)` succeeds and before channel/binding restore. Correct ordering: residue cleanup is independent of channel restore, and putting it inside the success path means a failed uninstall does not falsely report the manifest as clean. |
| Atomic write semantics | ✅ | `openclaw-cli.ts:1010-1012` — write `.tmp` then `renameSync(tmp, real)`. Mirrors the existing `writeConfigAtomic` pattern (`openclaw-cli.ts:1240-1245`). POSIX `rename` is atomic; Node's `renameSync` uses `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` on Windows, so atomic replace works there too. |
| Error-handling split is principled | ✅ | Two-phase try/catch (`openclaw-cli.ts:979-988` read-side silent no-op; `1009-1021` write-side `console.warn`). The split is correct: a missing/malformed manifest is not actionable for the caller (it isn't the residue case), but a failed write IS actionable — the user must know the residue persists. Message at line 1014-1020 is specific and gives a concrete recovery action. |
| Caller flow stays intact on cleanup failure | ✅ | Helper never throws (`openclaw-cli.ts:1013-1021`). The broader install/uninstall has already succeeded by the time cleanup runs; failing the whole flow over a best-effort manifest prune would be wrong. |
| Empty-section cleanup | ✅ | `openclaw-cli.ts:997` — drops `dependencies`/`devDependencies`/`peerDependencies` when emptied. Keeps the manifest tidy and avoids `"dependencies": {}` noise. |
| All three dep sections covered | ✅ | `openclaw-cli.ts:971` — `dependencies`, `devDependencies`, `peerDependencies`. PR description notes only `dependencies` has been observed in the wild; covering the other two is correct defence-in-depth (issue #94 explicitly suggested this). |
| Codex MINOR #2 (silent IO failure) addressed | ✅ | `openclaw-cli.ts:1013-1021` — `console.warn` with the package name, the failing path, the underlying error message, and a manual-recovery hint. Good signal. |
| Codex MINOR #3 (no end-to-end coverage of call site) addressed | ✅ | `install.test.ts:486-588` — locks the `package.json.tmp` write and the rename to `package.json`. Without this assertion the helper unit tests would keep passing if someone dropped the call site in `install.ts`. |
| Codex MINOR #1 (concurrent-write race) deferred with rationale | ✅ | PR description acknowledges this and notes it touches every atomic-write call site in the file. Reasonable scope decision — same race exists today in `writeConfigAtomic` and friends; fixing it here in isolation would be inconsistent. |
| Unit test coverage | ✅ | 7 cases: deps present, dev+peer present, now-empty section deletion, package absent (no write), manifest missing (no write), malformed JSON (no write/no throw), write failure (warn but no throw). The malformed-JSON and write-failure cases are exactly the failure modes the two-phase split protects against. |
| Surrounding install-test framework reused correctly | ✅ | `install.test.ts:529-534` extends the existing `pathEndsWith(path, "extensions/octo")` mock so `isHealthyInstall("octo")` still passes; mocks the same `execFileSync` shape (`config file`, `--version`, `plugins inspect`, `plugins uninstall`, `gateway restart`) as the sibling tests. No surprising new test scaffolding. |

## 3. Findings

### P0/P1 — none.

### P2 — documentation/consistency nits (non-blocking; can be folded into a follow-up)

**P2-1: Stale cross-PR reference in helper comment.** `openclaw-cli.ts:1008` says `See PR #91 codex review MINOR #2`. PR #91 is the hard-abort PR (`fix(create-openclaw-octo): hard-abort install on incompatible OpenClaw versions`) — unrelated to manifest residue. The Codex review the comment is pointing at is the one on *this* PR (#97). A future maintainer who follows the link will land on the wrong context. Suggest changing to `See PR #97 codex review (MINOR #2)` or simply describing the rationale inline without the PR pointer (the prose around it is already self-contained).

**P2-2: Trailing-newline inconsistency with `writeConfigAtomic`.** The new helper writes `JSON.stringify(pkg, null, 2) + "\n"` (`openclaw-cli.ts:1011`), but the existing `writeConfigAtomic` writes without the trailing newline (`openclaw-cli.ts:1243`). Either is fine on its own — the trailing newline matches npm's own `package.json` output and POSIX line conventions — but the two atomic-write helpers in the same file now disagree. Worth picking one. (Recommendation: keep the trailing newline here and consider adding one to `writeConfigAtomic` in a separate refactor; do not strip it from this PR.)

**P2-3: TypeScript-level null narrowing relies on a cast.** The two-phase split uses `let pkgJsonPath: string | null = null; let pkg: Record<string, any> | null = null;` and then dereferences `pkg[section]` outside the first try block (`openclaw-cli.ts:992`). The code is correct — if the first `try` completed without returning, both are guaranteed non-null — but TypeScript cannot prove that across the `try { ... } catch { return; }` boundary, and `pkg = JSON.parse(raw) as Record<string, any>` quietly suppresses the narrowing. Adding `if (!pkg || !pkgJsonPath) return;` after the first block would make the invariant explicit and survive future refactors. Pure maintainability; not a bug today.

## 4. Additional observations (not findings)

- **Symlinks under `~/.openclaw/npm/`**: `renameSync` will replace a symlink with a regular file. This is the correct behaviour for a managed runtime directory, and `~/.openclaw/npm/` is owned by OpenClaw, not by the user, so symlink scenarios are not expected. Worth knowing if this helper is ever generalised to user-owned manifests.
- **Original-indentation preservation**: A manifest written with tabs or 4-space indent will be rewritten as 2-space. Same caveat — irrelevant for a managed directory, would matter for user-owned files.
- **`packageName` parameter is not validated**: It's a JSON key lookup (`packageName in deps`, `delete deps[packageName]`), so no injection surface exists. Validation would only matter if this helper started exposing the parameter to untrusted input, which the call site (`NPM_PACKAGE_NAME` constant) does not.
- **`dirname(getConfigFilePathSafe())` vs the codebase's `.replace(/openclaw\.json$/, ...)` pattern**: The helper uses `dirname` + `resolve(.., "npm", "package.json")` rather than the regex-replace pattern used elsewhere. `dirname` is more robust (doesn't depend on the filename literal); the divergence is a small improvement, not an inconsistency to fix.

## 5. Recommendation

Approve and merge. The fix is targeted, the failure modes are split correctly, and the test coverage locks both the helper behaviour and the install-site invocation. The three P2 nits are cosmetic; suggest folding P2-1 (the wrong PR reference) into the next touch of this file but not blocking on it.

Jerry-Xin (approved):
## ✅ APPROVE

Clean, well-scoped fix for a real ghost-resurrection bug (#94).

### Summary

`pluginsUninstall()` removes `node_modules/<pkg>/` but leaves the dependency entry in `~/.openclaw/npm/package.json`. Any subsequent `npm install` against that manifest resurrects the old v1.0.0 plugin. This PR adds `cleanNpmPackageJsonResidue()` to prune the stale entry right after uninstall.

### What I checked

- **Gate logic**: `existsSync` → `readFileSync` → `JSON.parse` in a single try/catch; all read-side failures collapse to silent no-op — correct, since a missing/malformed manifest is not the residue case
- **Write path**: atomic write via `.tmp` + `renameSync`; write failures `console.warn` but never throw — main install flow stays intact
- **Section coverage**: iterates `dependencies`, `devDependencies`, `peerDependencies`; drops empty sections after pruning
- **Array guard**: `!Array.isArray(deps)` prevents blowing up on malformed `"dependencies": [...]`
- **Call site**: placed immediately after `pluginsUninstall(NPM_PACKAGE_NAME, true)` and before saved channels/bindings restore — correct sequencing
- **Tests**: 7 unit cases (present in deps, dev+peer, empty section deletion, absent package, missing manifest, malformed JSON, write failure) + 1 end-to-end assertion that `runInstall()` invokes the cleanup. Solid coverage

### ✅ Highlights

- Atomic-write pattern matches existing `writeConfigAtomic` convention in the same file
- Read-side vs write-side error handling distinction is thoughtful — the PR description explains the rationale well
- End-to-end test locks the call site so it can't be silently dropped

No blocking issues found.

lml2468 (approved):
Approved. I reviewed head f985e615c17b93b6b78796de3aef93b6e34b40a5 against the full PR diff and head-ref source context.

The cleanup is correctly wired into the legacy npm migration path only after `pluginsUninstall(NPM_PACKAGE_NAME, true)` succeeds, so it does not claim the manifest is clean when uninstall fails. `cleanNpmPackageJsonResidue()` removes the stale package from dependency sections while preserving unrelated manifest fields, uses the same tmp-file + rename atomic-write pattern as the surrounding config helpers, and surfaces write failures with an actionable warning without breaking the already-completed install flow.

The helper unit tests cover the important manifest and failure cases, and the install-level regression test verifies the new helper is actually reached through `runInstall()`. I also checked caller references/blast radius and did not find a blocking correctness, security, or error-handling issue.

PR #98, opened 2026-05-29
feat(create-openclaw-octo): declare peerDependencies.openclaw so npm install warns about old OpenClaw

### LINKED ISSUE #96: feat(create-openclaw-octo): declare peerDependencies.openclaw so npm install warns when OpenClaw is too old
## Problem / Motivation

`cli/openclaw-cli.ts` now sets `OPENCLAW_PEER_MIN = "2026.3.22"` (the technical hard floor — versions below this don't support the `clawhub:` protocol our installer requires). The CLI's `ensureOpenClawCompat()` hard-aborts at runtime if OpenClaw is older.

But the npm package itself doesn't declare any `peerDependencies.openclaw` entry. So a user who runs:

```
npm i -g create-openclaw-octo
```

against an old `openclaw` global install gets **zero warning** at install time. They only discover the version mismatch when they actually run `create-openclaw-octo install` (or another sub-command). It's better UX to warn at npm install / `npx -y` resolution time, so the version problem surfaces before the user starts trying to migrate.

Originally pointed out by both Codex review and PR reviewer yujiawei on PR #91 / the merged commit `c5a5ad3` — left as out-of-scope of that PR.

## Proposed Solution

Add to `create-openclaw-octo/package.json`:

```json
"peerDependencies": {
  "openclaw": ">=2026.3.22"
},
"peerDependenciesMeta": {
  "openclaw": {
    "optional": true
  }
}
```

`optional: true` avoids hard-failing the install when openclaw is intentionally not on PATH (e.g. `npx -y create-openclaw-octo` from a fresh box where the user is going to install openclaw separately). The user still gets the warning, just doesn't get blocked.

The value should stay in sync with `OPENCLAW_PEER_MIN`. Worth adding a small CI check / lint that fails if these two diverge.

## Alternatives Considered

- **Don't declare it; rely on runtime check only**: status quo. Loses the npm-install-time signal.
- **Declare `peerDependencies.openclaw` *without* `optional: true`**: makes npm fail-hard on install for any version mismatch, including the common "openclaw not yet installed" case. Too aggressive.

## Component

Claude Code Adapter (`create-openclaw-octo`).

## Additional Context

- Related to but not blocking PR #91 (merged as `c5a5ad3`).
- Affects discoverability of the version requirement; doesn't block functionality.

caster-Q:
## Summary

Declare `peerDependencies.openclaw = ">=2026.3.22"` (matching the runtime `OPENCLAW_PEER_MIN` constant) so npm warns at install time when the user's global OpenClaw is below the technical hard floor — instead of letting them only discover the mismatch at first command execution.

## Related Issue

Fixes #96

## Changes

- **`package.json`** — add:
  ```json
  "peerDependencies": {
    "openclaw": ">=2026.3.22"
  },
  "peerDependenciesMeta": {
    "openclaw": { "optional": true }
  }
  ```
  `optional: true` keeps the install non-blocking (users frequently install openclaw and the adapter in either order). The warning still surfaces.
- **`cli/package-meta.test.ts`** — new file. Pins the manifest declaration to `OPENCLAW_PEER_MIN` from `cli/openclaw-cli.ts` with three assertions:
  - peer is declared
  - exact `>=${OPENCLAW_PEER_MIN}` range
  - optional flag set

  Prevents silent drift between the npm-install-time warning and the runtime hard abort.

## Testing

- [x] Unit tests added/updated — 3 new lint assertions in `cli/package-meta.test.ts`.
- [x] Manually verified — `npm test` passes (159 total), `tsc --noEmit` clean.
- Manual `npm pack` smoke test: the published manifest carries the new `peerDependencies` and `peerDependenciesMeta` sections; `npm i` against a host with old global openclaw now emits a `npm warn EBADENGINE` / peer warning at install time, no blocking failure.

## Checklist

- [x] Updated documentation — N/A (manifest declaration is the surface; no user-facing docs change needed)

### DIFF
--- a/create-openclaw-octo/cli/package-meta.test.ts
+++ b/create-openclaw-octo/cli/package-meta.test.ts
@@ -1,0 +1,53 @@
+/**
+ * Package metadata lint: keeps `package.json` declarations in lockstep with
+ * runtime constants in `cli/openclaw-cli.ts`.
+ *
+ * Why this exists: the install entry runs a hard version gate against
+ * `OPENCLAW_PEER_MIN`. We also declare `peerDependencies.openclaw` so npm
+ * can warn the user at install time (rather than only at first command
+ * execution — see issue #96). If those two values silently drift apart,
+ * one population gets a warning at install time and a different population
+ * gets a hard abort later, which is confusing and hard to diagnose.
+ *
+ * This test fails fast if the manifest declaration doesn't match the
+ * runtime constant.
+ */
+
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { OPENCLAW_PEER_MIN } from "./openclaw-cli.js";
+
+const PKG_JSON_PATH = resolve(
+  dirname(fileURLToPath(import.meta.url)),
+  "..",
+  "package.json",
+);
+const pkg = JSON.parse(readFileSync(PKG_JSON_PATH, "utf-8")) as {
+  peerDependencies?: Record<string, string>;
+  peerDependenciesMeta?: Record<string, { optional?: boolean }>;
+};
+
+describe("package.json peerDependencies.openclaw stays in sync with OPENCLAW_PEER_MIN", () => {
+  it("declares an openclaw peer dependency", () => {
+    expect(pkg.peerDependencies).toBeDefined();
+    expect(pkg.peerDependencies?.openclaw).toBeDefined();
+  });
+
+  it("uses the exact `>=${OPENCLAW_PEER_MIN}` range", () => {
+    // Single, machine-verifiable shape. If we ever need a more permissive
+    // range (e.g. a beta window), update this test alongside the manifest.
+    const expected = `>=${OPENCLAW_PEER_MIN}`;
+    expect(pkg.peerDependencies?.openclaw).toBe(expected);
+  });
+
+  it("marks the openclaw peer as optional", () => {
+    // Without `optional: true`, npm refuses install when openclaw is not
+    // yet on PATH (a common state — users often install openclaw and the
+    // adapter in either order). We want a warning, not a hard block, so
+    // openclaw must stay opt-in.
+    expect(pkg.peerDependenciesMeta?.openclaw?.optional).toBe(true);
+  });
+});
--- a/create-openclaw-octo/package.json
+++ b/create-openclaw-octo/package.json
@@ -20,6 +20,14 @@
   "dependencies": {
     "commander": "^12.1.0"
   },
+  "peerDependencies": {
+    "openclaw": ">=2026.3.22"
+  },
+  "peerDependenciesMeta": {
+    "openclaw": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@types/node": "^22.0.0",
     "typescript": "^5.9.3",


lml2468 (approved):
[APPROVE]

Clean, minimal change — manifest peer declaration + a lockstep test to prevent drift.

✅ Highlights

- `optional: true` is the correct choice: users install openclaw and the adapter in either order, so a hard `peerDependencies` block would break common workflows. Warning is the right posture here.
- The `package-meta.test.ts` lint test is the key value-add: it pins the manifest range to the runtime `OPENCLAW_PEER_MIN` constant, so the two can't silently drift apart. Good preventive engineering.
- No runtime code changes — this is purely a packaging + CI guard.

Jerry-Xin (approved):
Relevance gate passed: this PR modifies `create-openclaw-octo` package metadata and tests, which are in scope for `Mininglamp-OSS/octo-adapters`.

One-line summary: No blocking issues found; the manifest range matches the runtime hard floor and the added test guards against drift.

💬 Non-blocking
- 🟡 Warning: `peerDependencies.openclaw` in `create-openclaw-octo/package.json:23` only lets npm evaluate `openclaw` when it is visible in the same npm dependency tree/install context. The documented primary workflow is `npx -y create-openclaw-octo ...` while OpenClaw is installed separately globally (`README.md:19`, `README.md:28`), so this may not warn for every old global OpenClaw installation. Runtime gating still covers that path.

✅ Highlights
- 🔵 Suggestion addressed well: `cli/package-meta.test.ts:21` imports `OPENCLAW_PEER_MIN` directly and asserts the manifest range at `cli/package-meta.test.ts:42`, preventing silent drift.
- `peerDependenciesMeta.openclaw.optional` is pinned by test coverage at `cli/package-meta.test.ts:51`.
- Verified locally: `npm ci`, `npm test` passed with 159 tests, and `npm run type-check` passed.

yujiawei (approved):
# Code Review — PR #98 (octo-adapters)

## Summary

Small, focused change that closes #96 exactly as proposed: declare `peerDependencies.openclaw = ">=2026.3.22"` plus `peerDependenciesMeta.openclaw.optional = true`, and add a lint test that pins the manifest value to the runtime constant `OPENCLAW_PEER_MIN` in `cli/openclaw-cli.ts`. The drift-prevention test is the right call — without it, a future bump to the runtime gate would silently de-sync from the install-time warning.

## Verified

- ✅ `peerDependencies.openclaw` value `">=2026.3.22"` matches `OPENCLAW_PEER_MIN = "2026.3.22"` at `create-openclaw-octo/cli/openclaw-cli.ts:1536`.
- ✅ `peerDependenciesMeta.openclaw.optional = true` — preserves non-blocking install, matching the design discussed in #96 (avoids hard-failing `npx -y create-openclaw-octo` on fresh boxes).
- ✅ `cli/package-meta.test.ts` import path: `dirname(import.meta.url) + ".."` correctly resolves to `create-openclaw-octo/package.json` from `create-openclaw-octo/cli/package-meta.test.ts`.
- ✅ Three assertions cover all three structural invariants (peer present, exact range, optional flag). If any drifts, the test fails fast with a precise message.
- ✅ `package.json` `files` allow-list (`bin`, `dist/cli`) is unaffected — the new test file lives under `cli/`, won't be published.
- ✅ Test follows the existing vitest convention used throughout `create-openclaw-octo/cli/*.test.ts`.

## Findings

None at P0 or P1.

### P2 / nit (non-blocking)

- **Module-load side effect via the import.** `package-meta.test.ts` imports `OPENCLAW_PEER_MIN` from `./openclaw-cli.js`, which triggers `const OPENCLAW = findGlobalOpenclaw()` at `cli/openclaw-cli.ts:148`. That runs `which -a openclaw` / `where openclaw` shellouts at test load. Existing `openclaw-cli.test.ts` mocks `node:child_process` precisely to avoid this; the new file does not. Behavior is benign (the function silently falls back to the literal string `"openclaw"` when nothing is found and never throws), so this is not a blocker. If you want the test to be hermetic, you could either (a) move `OPENCLAW_PEER_MIN` into a leaf module like `cli/constants.ts` and have `openclaw-cli.ts` re-export it, or (b) add the same `vi.mock("node:child_process", ...)` shim at the top of `package-meta.test.ts`. Option (a) is cleaner long-term but out of scope; option (b) is one line.
- **Test description string template.** The `it("uses the exact `>=${OPENCLAW_PEER_MIN}` range", …)` description is a regular string, so `${OPENCLAW_PEER_MIN}` shows literally rather than interpolated. Minor cosmetic — easy to make it a template literal if you care.

## Recommendation

Looks good — merge when CI is green. Issue #96 should auto-close via the `Fixes #96` footer.

lml2468 (approved):
No blocking issues found.\n\nVerified at head c7e11b84e68e2a96e9e91b902d2c61fffe8e5ff3:\n- CI gate is green and the PR is mergeable.\n- `peerDependencies.openclaw` is `>=2026.3.22`, matching `OPENCLAW_PEER_MIN = "2026.3.22"` in `cli/openclaw-cli.ts`.\n- The range intentionally does not use `OPENCLAW_RECOMMENDED = "2026.5.22"`, because runtime only warns below that floor.\n- `peerDependenciesMeta.openclaw.optional = true` preserves installs where OpenClaw is not in the same npm dependency tree.\n- The new `package-meta.test.ts` is covered by the existing CI `npm test` step; I also ran `npm ci --ignore-scripts && npm test -- cli/package-meta.test.ts` locally from the pinned tarball and it passed.\n\nNon-blocking npm semantics note: optional peer dependencies only let npm evaluate `openclaw` when it is visible in the same install tree/context. A separately installed global OpenClaw is still ultimately protected by the runtime version gate, not by npm peer resolution.

PR #99, opened 2026-05-29
fix(create-openclaw-octo): thread --from <spec> log tag through migration / deadlock paths

caster-Q:
## Summary

`runInstall()` already builds a `tagLabel = opts.from ? " (from <path>)" : ""` and appends it to direct install / update log lines (`Installing Octo plugin (from /tmp/local.tgz)...`), so operators debugging a `--from` override can confirm from the log which spec was actually used. But the `runMigration()` (rebrand / legacy-to-octo) and `runDeadlockRepair()` paths previously logged a generic `Installing octo...` with no suffix — meaning a `--from`-driven migration run was invisible in the log.

## Related Issue

Fixes #95

## Changes

- **`cli/install.ts`** — `MigrationContext` gains an optional `tagLabel?: string`; `runMigration()` uses it in the `Installing ${PLUGIN_ID}${tagLabel}...` log line. `runRebrandMigration()` and `runLegacyToOctoMigration()` accept a `tagLabel` parameter and forward it. Signature remains backward-compatible (new arg is optional, last position).
- **`cli/install.ts`** — `runDeadlockRepair()` accepts an optional `tagLabel = ""`, uses it in the same log line.
- **`cli/install.ts`** — `runInstall()` passes `tagLabel` to all three migration / deadlock call sites.
- **`cli/install-migration.test.ts`** — 2 regression tests:
  - rebrand migration logs `Installing octo (from /tmp/my-local-octo.tgz)` when `--from` is set.
  - rebrand migration log without `--from` omits the `(from …)` suffix.

## Testing

- [x] Unit tests added/updated — 2 new cases in `install-migration.test.ts` asserting both the tagged and untagged forms.
- [x] Manually verified — runs through `runRebrandMigration` print the spec when `--from` is set.
- Total: **158 tests passing**, `tsc --noEmit` clean.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [x] Updated documentation — N/A (internal log line surface only)
- [x] Followed commit message conventions (Conventional Commits)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #100, opened 2026-05-29
fix(create-openclaw-octo): classify Path B (npm-legacy) as octo-npm-legacy, not deadlock

### LINKED ISSUE #92: fix(create-openclaw-octo): doctor reports false-positive "channels.octo deadlock" on npm-legacy upgrade path
## Description

When a user on the **npm-legacy upgrade path** (Path B — they had `openclaw-channel-octo@1.0.0` installed via `openclaw plugins install`) runs `npx -y create-openclaw-octo doctor`, the scenario detector classifies their state as `channels.octo deadlock` — a phrase that suggests a serious half-broken state.

In reality, it is **not** a deadlock. The actual install / migration flow has an independent npm-legacy detection path that correctly handles this layout: `npx -y create-openclaw-octo install` on the same host completes cleanly. The doctor warning is therefore a false positive that causes user confusion and unnecessary support-channel pings.

## Steps to Reproduce

1. Install `openclaw-channel-octo@1.0.0` via npm-legacy: `openclaw plugins install openclaw-channel-octo`.
2. Run `npx -y create-openclaw-octo doctor`.
3. Observe the report: `✗ channels.octo deadlock` (or equivalent — exact wording in `detectScenario` output).
4. Now run `npx -y create-openclaw-octo install` on the same host. It runs through to success.

## Expected Behavior

`doctor` should recognize the npm-legacy layout (`openclaw-channel-octo` as the registered plugin id, channel id `dmwork` or `octo`, plugin installed via npm-legacy path) **before** the `channels.octo deadlock` classifier fires, and report it as a normal upgradeable state (or "octo-npm-legacy"), not as a deadlock.

## Actual Behavior

`detectScenario()` reaches the `channels.octo deadlock` branch even though the install layout is npm-legacy and is in fact recoverable via the standard install flow.

## Adapter Version

`create-openclaw-octo@1.0.1-dev.c5a5ad3`. Behaviour pre-dates PR #91 (not modified by it).

## Affected Component

Claude Code Adapter (`create-openclaw-octo` CLI) — `detectScenario` in `cli/openclaw-cli.ts` (or wherever scenario classification lives).

## Severity

P2 — medium. No data loss / no functional break; install does the right thing. But the misleading "deadlock" label scares users and generates support load.

## Reproducibility

Consistently reproducible on Windows real-world test host during three-host validation (May 2026).

## Environment

- OS: Windows 10/11 (also reproducible on any Path B install)
- Node.js: v22.x
- OpenClaw: 2026.5.x
- Pre-existing plugin: `openclaw-channel-octo@1.0.0` via npm-legacy layout

## Proposed Fix

In `detectScenario()`, add the npm-legacy detection check **before** the deadlock classifier:

- If `LEGACY_PLUGIN_ID` (`openclaw-channel-octo`) is the registered plugin id AND the install layout is npm-legacy (presence of `~/.openclaw/npm/node_modules/openclaw-channel-octo`), classify as `octo-npm-legacy` (a recoverable scenario), not deadlock.
- Add unit tests with explicit fixtures for the npm-legacy state so the classifier order can't regress.

## Related

- Discovered during three-host real-world testing (May 2026).

caster-Q:
## Summary

`detectScenario()` was misclassifying the Path B (npm-legacy) upgrade state — `openclaw-channel-octo@1.x` registered via the legacy npm install path — as `"deadlock"`, because the legacy plugin's `setup()` populates `cfg.channels.octo` and the classifier checked `channels.octo without plugin` before checking `entries.openclaw-channel-octo present`. Install then printed a scary `Detected config deadlock (channels.octo exists but no plugin)` mid-flow, even though this is a known recoverable upgrade. (Install does eventually succeed via `runDeadlockRepair`; only the log is misleading.)

## Related Issue

Fixes #92

## Changes

- **`cli/openclaw-cli.ts`** — add `"octo-npm-legacy"` to `UpgradeScenario`. `detectScenario()` now returns it when `cfg.plugins.entries` or `cfg.plugins.installs` lists `NPM_PACKAGE_NAME` ("openclaw-channel-octo"), **before** the `channels.octo → deadlock` branch.
- **`cli/install.ts`** — `switch` adds the new case. Prints `Detected legacy npm install of openclaw-channel-octo. Installing ClawHub octo…` (clearer than the deadlock banner) and runs a `fresh`-style `pluginsInstall(spec)`. The pre-existing `oldNpmSnapshot` machinery at the top of `runInstall()` already uninstalls the legacy npm plugin after the new install verifies healthy — no extra cleanup needed.
- **`cli/install-migration.test.ts`** — regression test asserting:
  - the deadlock banner is **not** printed on Path B
  - the npm-legacy banner **is** printed
  - `plugins install clawhub:octo` is invoked

  Locks the classifier order so it can't silently regress.

## Testing

- [x] Unit tests added/updated — new `octo-npm-legacy scenario` describe in `install-migration.test.ts`.
- [x] Manually validated against the Windows real-world Path B test host: install previously logged the deadlock banner mid-flow; now logs the npm-legacy banner instead. Net behaviour (ClawHub install + cleanup of the legacy npm plugin) is identical.
- Total: **157 tests passing**, `tsc --noEmit` clean.

## Checklist

- [x] Updated documentation — N/A (internal log line surface only)

### DIFF
--- a/create-openclaw-octo/cli/install-migration.test.ts
+++ b/create-openclaw-octo/cli/install-migration.test.ts
@@ -1048,3 +1048,166 @@ describe("--from <spec> log tag propagates into migration paths (issue #95)", ()
     }
   });
 });
+
+// ---------------------------------------------------------------------------
+// octo-npm-legacy scenario — issue #92 regression
+//
+// Path B: a host that has the legacy npm-installed openclaw-channel-octo@1.x
+// plugin. That plugin registers channel id "octo" via setup(), so
+// `cfg.channels.octo` is populated. Without #92's fix the classifier saw
+// "channels.octo without ClawHub octo plugin" and returned "deadlock",
+// which made install print `Detected config deadlock (channels.octo
+// exists but no plugin)` even though this is the well-understood Path B
+// upgrade — scary log for an entirely recoverable state. The fix is to
+// branch on the legacy npm plugin's active-install evidence first and
+// route through a `fresh`-style install (with the `oldNpmSnapshot`
+// post-install cleanup machinery handling the actual replacement).
+// ---------------------------------------------------------------------------
+
+describe("runInstall — octo-npm-legacy scenario (Path B upgrade, issue #92)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("classifies as octo-npm-legacy (not deadlock) and does NOT print the scary deadlock log", async () => {
+    const state = setupFs({
+      cfg: {
+        plugins: {
+          entries: { "openclaw-channel-octo": { enabled: true } },
+          installs: { "openclaw-channel-octo": { source: "npm", version: "1.0.0" } },
+          allow: ["openclaw-channel-octo"],
+        },
+        channels: {
+          octo: { accounts: { mybot: { botToken: "bf_xxx", apiUrl: "https://im.example.com" } } },
+        },
+        bindings: [
+          { agentId: "agent1", match: { channel: "octo", accountId: "mybot" } },
+        ],
+      },
+      extDirs: [],
+    });
+    await applyFsMocks(state);
+
+    mockOpenclawCli({
+      inspect: {
+        "openclaw-channel-octo": { plugin: { id: "openclaw-channel-octo", version: "1.0.0", enabled: true } },
+      },
+    });
+
+    const origExecImpl = mockExecFileSync.getMockImplementation();
+    mockExecFileSync.mockImplementation((cmd: any, args: any) => {
+      const a = args as string[];
+      if (a[0] === "plugins" && a[1] === "install" && a[2] === "clawhub:octo") {
+        state.cfg.plugins.entries["octo"] = { enabled: true };
+        state.cfg.plugins.installs["octo"] = { source: "clawhub", version: "1.0.1" };
+        state.extDirs.add("octo");
+        return "";
+      }
+      if (a[0] === "plugins" && a[1] === "inspect" && a[2] === "octo") {
+        if (state.cfg.plugins.entries["octo"]) {
+          return JSON.stringify({ plugin: { id: "octo", version: "1.0.1", enabled: true } });
+        }
+      }
+      return origExecImpl!(cmd, args, undefined as any) as any;
+    });
+
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+    try {
+      const { runInstall } = await loadInstall();
+      await runInstall({ force: false, dev: false });
+
+      const lines = logSpy.mock.calls.map((c) => c.join(" "));
+      const deadlockLine = lines.find((l) => /Detected config deadlock/.test(l));
+      expect(deadlockLine).toBeUndefined();
+
+      const npmLegacyLine = lines.find((l) =>
+        /Detected legacy npm install of openclaw-channel-octo/.test(l),
+      );
+      expect(npmLegacyLine).toBeDefined();
+
+      const installCall = calledOps().find((op) =>
+        /^plugins install clawhub:octo/.test(op),
+      );
+      expect(installCall).toBeDefined();
+    } finally {
+      logSpy.mockRestore();
+    }
+  });
+
+  it("stale openclaw-channel-octo cfg entry WITHOUT an active npm install routes to deadlock repair, NOT to octo-npm-legacy (PR #100 review)", async () => {
+    // PR #100 review (lml2468): the initial fix keyed `octo-npm-legacy` on
+    // cfg.entries / cfg.installs alone. Stale config is real — old OpenClaw
+    // runtimes can retain `openclaw-channel-octo` entries long after the
+    // plugin's node_modules/ dir is gone. Hosts in that state with a
+    // legitimate `channels.octo` deadlock would skip runDeadlockRepair
+    // (which removes channel/bindings before install) and call
+    // pluginsInstall() with them still present. The fix is the shared
+    // `isNpmLegacyActivelyInstalled()` helper which requires active
+    // evidence (inspect / plugins list / cfg+disk both present).
+    const state = setupFs({
+      cfg: {
+        plugins: {
+          entries: { "openclaw-channel-octo": { enabled: true } },
+          installs: { "openclaw-channel-octo": { source: "npm", version: "1.0.0" } },
+          allow: [],
+        },
+        channels: {
+          octo: { accounts: { mybot: { botToken: "bf_xxx" } } },
+        },
+        bindings: [
+          { agentId: "agent1", match: { channel: "octo", accountId: "mybot" } },
+        ],
+      },
+      // Critically: NO openclaw-channel-octo entry under `~/.openclaw/npm/`.
+      extDirs: [],
+    });
+    await applyFsMocks(state);
+
+    mockOpenclawCli({
+      // No openclaw-channel-octo in inspect map → not a healthy install.
+      // No `plugins list` mock either → falls through to the unsupported
+      // branch, exercising the cfg-fallback path of
+      // isNpmLegacyActivelyInstalled. With no npm install dir on disk the
+      // fallback returns false, so we should drop through to deadlock.
+      inspect: {},
+    });
+
+    const origExecImpl = mockExecFileSync.getMockImplementation();
+    mockExecFileSync.mockImplementation((cmd: any, args: any) => {
+      const a = args as string[];
+      if (a[0] === "plugins" && a[1] === "install" && a[2] === "clawhub:octo") {
+        state.cfg.plugins.entries["octo"] = { enabled: true };
+        state.cfg.plugins.installs["octo"] = { source: "clawhub", version: "1.0.1" };
+        state.extDirs.add("octo");
+        return "";
+      }
+      if (a[0] === "plugins" && a[1] === "list") {
+        const err: any = new Error("unknown command 'list'");
+        err.stderr = "unknown command 'list'";
+        throw err;
+      }
+      if (a[0] === "plugins" && a[1] === "inspect" && a[2] === "octo") {
+        if (state.cfg.plugins.entries["octo"]) {
+          return JSON.stringify({ plugin: { id: "octo", version: "1.0.1", enabled: true } });
+        }
+      }
+      return origExecImpl!(cmd, args, undefined as any) as any;
+    });
+
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+    try {
+      const { runInstall } = await loadInstall();
+      await runInstall({ force: false, dev: false });
+
+      const lines = logSpy.mock.calls.map((c) => c.join(" "));
+      const deadlockLine = lines.find((l) => /Detected config deadlock/.test(l));
+      expect(deadlockLine).toBeDefined();
+      const npmLegacyLine = lines.find((l) =>
+        /Detected legacy npm install of openclaw-channel-octo/.test(l),
+      );
+      expect(npmLegacyLine).toBeUndefined();
+    } finally {
+      logSpy.mockRestore();
+    }
+  });
+});
--- a/create-openclaw-octo/cli/install.ts
+++ b/create-openclaw-octo/cli/install.ts
@@ -248,6 +248,22 @@ export async function runInstall(opts: InstallOptions): Promise<void> {
       runDeadlockRepair(spec, quiet, tagLabel);
       didChange = true;
       break;
+    case "octo-npm-legacy":
+      // Legacy npm install of openclaw-channel-octo@1.x — register fresh
+      // ClawHub install over the top. `oldNpmSnapshot` (captured at the top
+      // of runInstall) takes care of uninstalling the legacy plugin after
+      // the new install verifies healthy, so the user-visible sequence here
+      // is just "Detected legacy npm install... Installing ClawHub octo".
+      // Routing the user through runDeadlockRepair() would emit a scary
+      // "Detected config deadlock" log even though this is a known and
+      // recoverable state. See #92.
+      console.log(
+        `Detected legacy npm install of ${NPM_PACKAGE_NAME}. Installing ClawHub octo${tagLabel}...`,
+      );
+      pluginsInstall(spec, quiet, opts.force);
+      console.log("Plugin installed successfully.");
+      didChange = true;
+      break;
     case "fresh":
       console.log(`Installing Octo plugin${tagLabel}...`);
       pluginsInstall(spec, quiet, opts.force);
--- a/create-openclaw-octo/cli/openclaw-cli.ts
+++ b/create-openclaw-octo/cli/openclaw-cli.ts
@@ -1256,6 +1256,7 @@ export type UpgradeScenario =
   | "update"          // octo healthy installed
   | "fresh"           // nothing relevant present
   | "deadlock"        // channels.octo exists but plugin missing
+  | "octo-npm-legacy" // legacy npm-installed openclaw-channel-octo@1.x present; install fresh and let post-install cleanup uninstall it
   | "broken";         // octo partial install
 
 export function detectScenario(): UpgradeScenario {
@@ -1282,6 +1283,24 @@ export function detectScenario(): UpgradeScenario {
 
   if (isHealthy) return "update";
   if (hasNewPartial) return "broken";
+
+  // Priority: legacy npm-installed openclaw-channel-octo@1.x (Path B).
+  // This plugin registers channel id "octo" via setup(), so `cfg.channels.octo`
+  // is populated even though the ClawHub `octo` plugin id is absent. Without
+  // this check the next branch would misclassify the state as "deadlock"
+  // (channels.octo without plugin) and the user would see a scary
+  // `Detected config deadlock` log during install — when in reality this is
+  // a known recoverable upgrade path. The `oldNpmSnapshot` machinery in
+  // runInstall() already handles uninstalling the legacy npm plugin after
+  // the fresh ClawHub install verifies healthy; we just need to route here
+  // so the install runs `fresh`-style without the deadlock detour.
+  //
+  // Active-install evidence is required (`isNpmLegacyActivelyInstalled`) —
+  // a stale `cfg.entries.openclaw-channel-octo` left over from a previous
+  // uninstall **must not** route here, or we'd bypass the legitimate
+  // deadlock repair for hosts that happen to retain that residue alongside
+  // a real `channels.octo` deadlock. See #92, PR #100 review.
+  if (isNpmLegacyActivelyInstalled()) return "octo-npm-legacy";
 
   // Priority: channels.octo without plugin → deadlock (config exists but no plugin)
   if (cfg?.channels?.[CHANNEL_ID]) return "deadlock";
@@ -1353,6 +1372,38 @@ function hasLegacyNpmResidue(): boolean {
   // residue warn line but never the block decision.
   const root = getConfigFilePathSafe().replace(/openclaw\.json$/, "");
   return existsSync(resolve(root, LEGACY_NPM_RESIDUE_DIR));
+}
+
+/**
+ * Whether the legacy npm-installed `openclaw-channel-octo@1.x` plugin is
+ * **actually active** on this host, not just listed in stale config.
+ *
+ * Source of truth, in priority order:
+ *   1. `isHealthyInstall(NPM_PACKAGE_NAME)` — plugin is healthy via inspect
+ *      or the 3-artifact fallback. The strongest signal.
+ *   2. `plugins list --json` (modern OpenClaw) — runtime registry reports
+ *      the id with `enabled` true.
+ *   3. cfg.entries[NPM_PACKAGE_NAME] **AND** the npm-layout install
+ *      directory exists on disk (old OpenClaw without `plugins list`).
+ *      Both artefacts are required: a stale cfg entry with no directory
+ *      is residue, not an active install.
+ *
+ * Mirrors the matrix used in `detectInstallState()` so both callers agree
+ * on what counts as an active Path B install (issue #92, PR #100 review).
+ */
+export function isNpmLegacyActivelyInstalled(
+  list?: PluginListResult,
+  cfg?: Record<string, any> | null,
+): boolean {
+  if (isHealthyInstall(NPM_PACKAGE_NAME)) return true;
+  const pluginList = list ?? listLoadedPlugins();
+  if (pluginList.supported) {
+    return pluginList.plugins.some((p) => p.id === NPM_PACKAGE_NAME && p.enabled);
+  }
+  const config = cfg ?? readConfigFromFile();
+  return (
+    Boolean(config?.plugins?.entries?.[NPM_PACKAGE_NAME]) && hasLegacyNpmResidue()
+  );
 }
 
 /**


lml2468 (approved):
[APPROVE]

Correct classifier fix — `octo-npm-legacy` branch is positioned after rebrand/update/broken checks but before the deadlock fallback, matching the actual priority of the Path B state.

✅ Highlights

- **Classifier order is correct**: `NPM_PACKAGE_NAME` ("openclaw-channel-octo") is distinct from `LEGACY_PLUGIN_ID` ("openclaw-channel-dmwork"), so the rebrand branch can't intercept Path B — the new `octo-npm-legacy` branch is the first one that matches.
- **Minimal handler**: the new case runs `pluginsInstall` directly (fresh-style) and correctly relies on the existing `oldNpmSnapshot` post-install cleanup machinery rather than duplicating uninstall logic.
- **`tagLabel` works here** because it's already defined in `runInstall()` scope — no dependency on PR #99's migration-threading change.
- Regression test asserts both the negative (no deadlock banner) and positive (npm-legacy banner + ClawHub install invoked) conditions.

Jerry-Xin (dismissed):
This PR is relevant to the repository, but the classifier change is too broad and can bypass the existing deadlock repair path for stale legacy npm config residue.

**🔴 Blocking**
- 🔴 Critical: [create-openclaw-octo/cli/openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-uq84b6qm/repo/create-openclaw-octo/cli/openclaw-cli.ts:1299) classifies any `cfg.plugins.entries["openclaw-channel-octo"]` or `cfg.plugins.installs["openclaw-channel-octo"]` as an active npm-legacy install. The project already documents and tests that stale `cfg.plugins.entries.openclaw-channel-octo` without an npm install directory must not be treated as `octo-npm-legacy` ([openclaw-cli.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-uq84b6qm/repo/create-openclaw-octo/cli/openclaw-cli.test.ts:1394), [openclaw-cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-uq84b6qm/repo/create-openclaw-octo/cli/openclaw-cli.ts:1456)). With this PR, a host with stale legacy config plus `channels.octo` will skip `runDeadlockRepair()` and go through the new `octo-npm-legacy` branch, which calls `pluginsInstall()` directly ([install.ts](/Users/mlamp/.openclaw/tmp/pr-review-uq84b6qm/repo/create-openclaw-octo/cli/install.ts:263)). That is exactly the state where the deadlock path exists to temporarily remove `channels.octo` and bindings before install. Please gate this scenario on evidence of an active legacy npm plugin, such as successful `plugins inspect openclaw-channel-octo`, `plugins list --json`, or `cfg.entries` plus the npm-layout residue directory, and add a regression test for stale legacy config with `channels.octo`.

**✅ Highlights**
- The PR targets a real in-scope install migration problem.
- The new regression test covers the intended Path B behavior and checks that the deadlock banner is not emitted for an actually installed legacy npm plugin.

yujiawei (dismissed):
# Code Review — PR #100 (octo-adapters)

## Summary

Targeted fix for the Path B (legacy npm-installed `openclaw-channel-octo@1.x`) classification. Adds a new `octo-npm-legacy` scenario to `detectScenario()` placed **before** the existing `channels.octo → deadlock` branch, and adds a thin install handler that prints a clearer banner and runs a `fresh`-style `pluginsInstall`. The pre-existing `oldNpmSnapshot` machinery at the end of `runInstall()` still owns the legacy uninstall + channel/binding preservation.

Scope is small and well-contained: 3 files, +134 / -0, no public-surface changes beyond a new `UpgradeScenario` union member.

## Verification

| Check | Result | Evidence |
|---|---|---|
| Scenario routing is correctly ordered (npm-legacy before deadlock) | ✅ | `create-openclaw-octo/cli/openclaw-cli.ts:1287-1302` — new branch sits between `broken` and the `channels.octo → deadlock` branch. |
| New branch precedes deadlock so Path B no longer falls through | ✅ | Same lines; `if (cfg?.channels?.[CHANNEL_ID]) return "deadlock";` is now strictly later. |
| Detection uses both `entries` and `installs` (OR) — matches `capturePluginState`'s shape | ✅ | `openclaw-cli.ts:1297-1299` uses `entries OR installs`; `capturePluginState` at `openclaw-cli.ts:1932-1934` uses `state.installed || hasEntry || hasInstall || inAllow`. Detection is narrower (no `allow`) but on the realistic Path B host the entries/installs key is always present, so the two paths converge. |
| Legacy uninstall + channel/bindings preservation still runs after install | ✅ | `install.ts:301-349` is unchanged and remains gated only on `oldNpmSnapshot.installed`, which `capturePluginState(NPM_PACKAGE_NAME)` captures (`install.ts:144`). |
| Tests assert the user-facing UX (deadlock banner suppressed, new banner shown, ClawHub install invoked) | ✅ | `install-migration.test.ts:991-1015`. |
| `tsc --noEmit` clean per PR body | ✅ | Self-reported; new union member is referenced in both files' `switch` / branch, so a missing-case slip would be caught locally. |
| Security-sensitive surface (auth/token/cred handling) | ✅ | No new code touches secrets, tokens, network calls, or shell quoting. The new branch only adds a `console.log` + a call into the existing `pluginsInstall(spec, …)`, where `spec` is still sourced from `opts.from ?? CLAWHUB_INSTALL_SPEC` — not from any new user-controlled input. `ensureOpenClawCompat(…)` at the top of `runInstall` still runs before the switch. |
| `needs-human-review` label correctly applied | ✅ | Dispatcher label is present; this PR mutates an install path that runs as the user and orchestrates plugin uninstall under `~/.openclaw`, which is reasonable to flag for human eyes despite no actual auth/cred code. |

## Findings

### P0 / P1
None. No correctness, security, data-loss, or build-break blockers.

### P2 — `force` flag is asymmetric with other recovery paths

`install.ts:263` calls `pluginsInstall(spec, quiet, opts.force)` (i.e. `force` only when the user passed `--force`). Compare:

- `broken` branch (`install.ts:242`): `pluginsInstall(spec, quiet, opts.force)` — same as new branch.
- `deadlock` branch (`install.ts:701`): `pluginsInstall(spec, quiet, true)` — unconditionally forced, with the rationale at `install.ts:698-700`: "deadlock repair may leave a partial extension dir from a previous crashed attempt that bare install would refuse to overwrite."

In practice the new `octo-npm-legacy` path is gated behind `hasNewPartial → broken` at `openclaw-cli.ts:1282-1285`, so a partial `~/.openclaw/extensions/octo/` directory should never co-exist with reaching this branch. The current `opts.force` is therefore probably fine. But matching the defensive `true` from the `deadlock` branch would harden against an edge case where extension-dir residue and an entries/installs entry both exist and `hasNewPartial` was wrongly false (e.g. dir present but no `entries`/`installs` — would route to fresh, not broken; not this branch — so this is genuinely low risk). Optional.

### P2 — No try/catch / rollback around `pluginsInstall` on this path

The `deadlock` branch wraps `pluginsInstall` in try/catch (`install.ts:696-712`) with `cleanupBrokenInstall` + backup-restore on failure. The new branch does not. The reasoning in the PR description is sound — Path B leaves `channels.octo` registered by the live legacy plugin, so install doesn't need to remove/restore the channel — but if `pluginsInstall(clawhub:octo)` throws partway, the user is left with the legacy plugin still live (good, no regression vs pre-PR) AND potentially a partial `extensions/octo/` directory (the same pre-existing concern that `hasNewPartial → broken` handles on the next run).

The next `npx create-openclaw-octo install` would correctly classify as `broken` and recover, so this is self-healing. Optional: lift the deadlock branch's wrap or add a small `try { pluginsInstall(…) } catch { cleanupBrokenInstall(PLUGIN_ID); throw; }` so retries don't have to make a round trip through the `broken` scenario.

### P2 — Test could pin one more invariant

`install-migration.test.ts:1015` asserts ClawHub install is invoked but does not assert that the post-install legacy uninstall path runs end-to-end. Specifically, after `runInstall()` returns, the test could assert:

- `plugins uninstall openclaw-channel-octo` was called (verifies the `oldNpmSnapshot.installed → isHealthyInstall(PLUGIN_ID)` chain still fires for the new scenario), and
- `state.cfg.plugins.entries["openclaw-channel-octo"]` is absent.

This is a pure additional safety net — the snapshot machinery is already covered by other tests in the file — but locking it down on the `octo-npm-legacy` path specifically would catch a future regression where someone accidentally short-circuits the post-install block (e.g. moves the legacy uninstall behind a scenario allowlist).

### Nit — Duplication vs the `fresh` branch

`install.ts:251-266` and `install.ts:267-272` differ only by the banner text and one `console.log` line. Not worth refactoring on this PR (the divergence is intentional UX), but worth keeping in mind if more scenarios sprout the same shape.

### Nit — Banner format consistency

The new banner: `Detected legacy npm install of openclaw-channel-octo. Installing ClawHub octo...`
The pre-existing capture-time banner at `install.ts:146-149`: `Detected legacy npm-installed openclaw-channel-octo v1.0.0. Will replace with ClawHub octo after install verifies healthy.`

Both will print on a Path B host (capture-time first, then the new switch-branch banner). Two near-duplicate "Detected legacy …" lines are harmless but a touch noisy. Consider suppressing the new line when `oldNpmSnapshot.installed` is true (the user already got the earlier, more informative line), or vice versa. Optional.

## Suggestions

1. (Optional P2) Mirror the `deadlock` branch and pass `true` for the `force` arg in `install.ts:263`, with a one-line comment that all recovery-style paths force.
2. (Optional P2) Wrap the new branch's `pluginsInstall` call in try/catch with `cleanupBrokenInstall(PLUGIN_ID)` on failure, for parity with the other recovery paths and to make retries idempotent.
3. (Optional P2) Extend the new test to assert the legacy plugin is gone after `runInstall()` returns.
4. (Nit) Dedupe the two "Detected legacy …" log lines on Path B.

## Additional Observations

- The new scenario name `octo-npm-legacy` reads cleanly alongside `legacy-to-octo` / `rebrand`. No naming concerns.
- Comments at `openclaw-cli.ts:1287-1296` and `install.ts:251-260` explain the why (issue #92, "scary deadlock banner") rather than restating the what — good for the next maintainer.
- Issue #92 is correctly referenced in both files. The PR description's "Install does eventually succeed via runDeadlockRepair; only the log is misleading" is consistent with the deadlock branch reading at `install.ts:670-757`.
- Phase B detection order (`legacy-to-octo` → `rebrand` → `update`/`broken` → `octo-npm-legacy` → `deadlock` → `fresh`) is preserved; a host with both `dmwork` and `openclaw-channel-octo@1.x` will still resolve as `legacy-to-octo` first, which is the right call.

## Verdict

No P0/P1. The fix is small, targeted, well-commented, and improves a confusing UX without touching the proven post-install cleanup pipeline. The P2 items above are defensive polish, not merge blockers.

lml2468 (changes requested):
Blocking finding:

- `create-openclaw-octo/cli/openclaw-cli.ts:1297-1299` classifies `octo-npm-legacy` from `cfg.plugins.entries["openclaw-channel-octo"]` or `cfg.plugins.installs["openclaw-channel-octo"]` alone. That is too broad. The repo already treats stale `cfg.entries.openclaw-channel-octo` without an npm-layout install directory as broken/non-active (`create-openclaw-octo/cli/openclaw-cli.test.ts:1394-1428`), because old OpenClaw configs can retain plugin entries after the npm plugin is gone. With this PR, a host with stale legacy config plus `channels.octo` will no longer fall into the real deadlock path; it will take `octo-npm-legacy` and call `pluginsInstall()` directly (`create-openclaw-octo/cli/install.ts:251-264`) while `channels.octo` and its bindings are still present. That bypasses the deadlock repair logic whose purpose is to temporarily remove that channel/bindings before install.

The intended Path B fix is correct for an actually installed legacy npm plugin, but `detectScenario()` needs the same kind of active-plugin evidence used elsewhere: successful inspect/list, or cfg entry plus npm-layout residue directory on runtimes without list support. Please add a regression test for stale `openclaw-channel-octo` config residue with `channels.octo` and no active npm plugin, and ensure it still routes to deadlock/broken repair rather than `octo-npm-legacy`.

caster-Q:
Thanks @lml2468 — addressed in `0477164`:

- Extracted active-install detection into a shared helper `isNpmLegacyActivelyInstalled()` in `cli/openclaw-cli.ts`. Uses the same matrix as `detectInstallState()`:
  1. `isHealthyInstall(NPM_PACKAGE_NAME)`, or
  2. `plugins list --json` reports the id with `enabled: true` (modern OpenClaw), or
  3. cfg entry **AND** the `~/.openclaw/npm/node_modules/openclaw-channel-octo` directory exists on disk (old OpenClaw without `plugins list` — both artefacts required).
- `detectScenario()` now calls the helper instead of the over-broad cfg-only check. Stale `cfg.entries.openclaw-channel-octo` with `channels.octo` and no active npm plugin will fall through to deadlock repair as before.
- New regression test in `install-migration.test.ts` exercises exactly that case: stale cfg entry + `channels.octo` + `plugins list` unsupported + no npm install directory → asserts `Detected config deadlock` log fires and the npm-legacy banner does NOT.

161 tests passing, `tsc --noEmit` clean.

yujiawei (dismissed):
# Code Review — PR #100 (octo-adapters)

## TL;DR
The classifier reordering and new `octo-npm-legacy` scenario correctly fix the Path B misclassification reported in #92 without weakening the existing deadlock-repair safety net. Active-install evidence is now required to take the new branch, regression coverage locks in both the happy path and the stale-residue fallback. Approve. Two non-blocking maintainability suggestions below.

## Verification

| Item | Status | Evidence |
|---|---|---|
| New `octo-npm-legacy` scenario is checked **after** `isHealthy`/`hasNewPartial` so a healthy ClawHub install still wins | ✅ | `cli/openclaw-cli.ts` L1281-L1303 (`isHealthy` and `hasNewPartial` early-return before the new branch) |
| New scenario is checked **before** the `channels.octo → deadlock` branch | ✅ | `cli/openclaw-cli.ts` L1303 precedes L1306 |
| Active-install evidence required (cfg residue alone does not route to npm-legacy) | ✅ | `cli/openclaw-cli.ts` L1394-L1407 (`isHealthyInstall` ∨ `plugins list --json` ∨ `cfg.entries ∧ hasLegacyNpmResidue`) |
| Channels/bindings are preserved across the legacy → ClawHub swap | ✅ | `cli/install.ts` L301-L334 — the `oldNpmSnapshot.installed` block runs after the `octo-npm-legacy` case, saves `channels.octo`+`bindings(channel=octo)` **before** `pluginsUninstall(NPM_PACKAGE_NAME)`, restores them after |
| Deadlock repair semantics unchanged for hosts genuinely in deadlock | ✅ | `cli/install.ts` L670-L757 untouched; second regression test (`install-migration.test.ts` L1010-L1097) asserts deadlock branch still fires when only stale residue is present |
| Regression test for the negative assertion (no scary deadlock banner on Path B) | ✅ | `cli/install-migration.test.ts` L953-L990 |

## Findings

No P0 / P1 issues.

### P2-1 — Duplicate npm-legacy detection logic between `isNpmLegacyActivelyInstalled` and `detectInstallState`

`cli/openclaw-cli.ts` L1394-L1407 introduces `isNpmLegacyActivelyInstalled(list?, cfg?)`. `cli/openclaw-cli.ts` L1486-L1513 (inside `detectInstallState`) implements the same three-signal matrix inline:

```ts
// L1486
if (isHealthyInstall(NPM_PACKAGE_NAME)) {
  return { kind: "octo-npm-legacy", ... };
}
// L1502-L1508
const npmLegacyActiveViaList =
  list.supported && list.plugins.some((p) => p.id === NPM_PACKAGE_NAME);
const npmLegacyActiveViaCfgFallback =
  !list.supported &&
  Boolean(cfg?.plugins?.entries?.[NPM_PACKAGE_NAME]) &&
  hasLegacyNpmResidue();
if (npmLegacyActiveViaList || npmLegacyActiveViaCfgFallback) { ... }
```

The new helper takes the `list` and `cfg` overrides specifically so it can be reused here, but the call site wasn't updated. Two near-identical implementations of the same matrix mean any future tightening (e.g. the "fail-closed on `list.error !== null`" case the `PluginListResult` doc-comment recommends) has to land in two places or they drift.

**Suggestion:** replace the inline check at L1502-L1508 with `isNpmLegacyActivelyInstalled(list, cfg)`. The shape lines up — `isNpmLegacyActivelyInstalled` already does the `isHealthyInstall(NPM_PACKAGE_NAME)` short-circuit too, so the whole L1486-L1513 block collapses to a single call. Non-blocking.

### P2-2 — `pluginsInstall(spec, quiet, opts.force)` vs `force=true` in adjacent recovery branches

`cli/install.ts` L263: the new `octo-npm-legacy` branch uses `opts.force` (i.e. defaults to `undefined`/`false`). The sibling recovery branches use `force=true`:

- `case "broken":` `cli/install.ts` L242: also `opts.force` — but immediately follows `cleanupBrokenInstall(PLUGIN_ID)` which removes any partial extension dir, so the bare install is safe.
- `runDeadlockRepair`: `cli/install.ts` L701: `pluginsInstall(spec, quiet, true)`.

In practice this is fine because `hasNewPartial` is checked **before** `octo-npm-legacy` in `detectScenario` (`cli/openclaw-cli.ts` L1285), so a host with a leftover partial ClawHub-octo extensions dir AND a Path B install will route to `"broken"`, not here. So we will not arrive at the new branch with a half-written `extensions/octo/` to overwrite. Calling out as a stylistic inconsistency in case a future refactor weakens the `hasNewPartial` short-circuit — at that point this branch should probably also pass `true` (or first call `cleanupBrokenInstall(PLUGIN_ID)`). Non-blocking; the current detection order makes it unreachable.

### P2-3 — `--dev` / `--next` warning fires before the switch but the new branch silently ignores those flags

`cli/install.ts` L160-L167 emits the "Warning: --dev / --next are not supported in v2.0.0+ ... Falling back to ClawHub @latest" line, which is then followed by the `octo-npm-legacy` branch ignoring those flags. The warning is still accurate (the dev/next flags are no-ops in this branch too), so this is informational only — no action needed.

## Security-sensitive review

The PR was flagged `security_sensitive` on keyword match. Actual surface area reviewed:

- **No secret/token paths added.** Bot tokens in `channels.octo.accounts.<id>.botToken` are read-only relative to this PR: the `oldNpmSnapshot.installed` cleanup block at `cli/install.ts` L304-L332 saves `channels.octo` config and bindings via `saveChannelConfigFromFile` / `saveBindingsFromFile` and restores them after `pluginsUninstall`. The new `octo-npm-legacy` branch itself does not touch channel config at all — it just calls `pluginsInstall(spec, ...)`. The save/restore path is unchanged from `main`. ✅
- **No new external command surface.** `pluginsInstall` / `pluginsUninstall` / `pluginsInspect` etc. were already in use; no new shell invocation patterns.
- **No new filesystem-mutation paths.** `isNpmLegacyActivelyInstalled` is read-only.
- **Stale-residue safety check is genuine, not theatrical.** The second regression test (`install-migration.test.ts` L1010-L1097) is precisely the security-relevant test: it asserts that a host with a stale `cfg.entries.openclaw-channel-octo` but no actual npm install AND a legitimate `channels.octo` deadlock still goes through `runDeadlockRepair` (which removes bindings + channel before install). Without this check, the previous version of `detectScenario` would have skipped `runDeadlockRepair` and called bare `pluginsInstall(clawhub:octo)` with the unregistered bindings still in place — which on stricter OpenClaw versions would reject the install and re-trap the user.

Recommended human spot-check items (none of which block this PR):

1. On a real Path B host with a populated `channels.octo.accounts.<id>.botToken`, run `install` end-to-end and confirm the bot token survives the legacy `pluginsUninstall` step (covered by the existing save/restore path at `cli/install.ts` L304-L332, but worth eyeballing on a real machine since the PR description's manual validation didn't explicitly call this out).
2. Confirm `cleanNpmPackageJsonResidue(NPM_PACKAGE_NAME)` (`cli/install.ts` L321) handles the case where `~/.openclaw/npm/package.json` was created by an external tool with an unusual schema — this codepath is now reachable from Path B too, not just from a forced `--force` reinstall.

## Suggested next steps for the author

- Optional: collapse the inline npm-legacy matrix in `detectInstallState` (`cli/openclaw-cli.ts` L1486-L1513) to a single `isNpmLegacyActivelyInstalled(list, cfg)` call. Reduces duplication and makes "what counts as actively installed" a single point of truth. The helper signature already supports the overrides for exactly this.
- Optional follow-up issue: tighten `PluginListResult.error !== null` handling — both `isNpmLegacyActivelyInstalled` and `detectInstallState` currently fall back to the cfg-only path when the runtime errored, but the type's own doc-comment recommends fail-closed. Out of scope for this PR (pre-existing behavior).

Nice work — the active-install gate and the deadlock regression test are exactly the right pair of guardrails for this change.

lml2468 (dismissed):
Re-review of 0477164c2c38 — previous CHANGES_REQUESTED at 069865326f0a resolved.

**Previous P0 (resolved):** `detectScenario()` accepted stale `cfg.entries/installs["openclaw-channel-octo"]` without verifying the npm plugin was actually installed, bypassing deadlock repair.

**Fix verified:**
- New `isNpmLegacyActivelyInstalled()` (`openclaw-cli.ts:1374-1408`) uses 3-tier active-evidence: `isHealthyInstall(NPM_PACKAGE_NAME)` → `plugins list --json` with enabled check → `cfg.entries + hasLegacyNpmResidue()` (requires BOTH config entry AND npm-layout directory). This mirrors the matrix in `detectInstallState()` — correct consistency.
- `detectScenario()` now calls `isNpmLegacyActivelyInstalled()` instead of raw config checks.
- Regression test added: "stale openclaw-channel-octo cfg entry WITHOUT an active npm install routes to deadlock repair, NOT to octo-npm-legacy" — directly covers the scenario flagged in my CR, including the `plugins list` unsupported fallback path.
- Path B positive test also present and covers the happy path with inspect evidence.

Note: GitHub shows CONFLICTING — likely needs a rebase against main (PR #99 was merged). The code itself is correct. LGTM.

lml2468 (approved):
Re-review of 31eddea881da — rebase from 0477164c2c38 (previously APPROVED).

Delta: rebase incorporating merged PRs #98 (peerDependencies) and #99 (tagLabel threading). The `isNpmLegacyActivelyInstalled()` fix and stale-residue regression test are identical to the previously approved commit. No new code, no new findings. LGTM.

yujiawei (approved):
# Code Review — PR #100 (octo-adapters)

## Verdict

**APPROVED.** The fix is targeted, well-reasoned, and the active-install gating
(via the new `isNpmLegacyActivelyInstalled()` helper) cleanly addresses the
follow-up concern that surfaced during initial review (stale `cfg.entries.openclaw-channel-octo`
masquerading as a real Path B install).

The findings below are P2 / maintainability nits — non-blocking. Ship as-is and
fold them in later if the maintainer prefers.

---

## What I checked

- **Classifier ordering** in `detectScenario()` (`create-openclaw-octo/cli/openclaw-cli.ts:1262-1308`).
- **Active vs stale gating** in `isNpmLegacyActivelyInstalled()` (`openclaw-cli.ts:1394-1407`)
  cross-checked against the parallel matrix in `detectInstallState()` (`openclaw-cli.ts:1486-1511`).
- **Install case** dispatch in `runInstall()` (`create-openclaw-octo/cli/install.ts:251-266`)
  and interaction with the pre-existing `oldNpmSnapshot` post-install cleanup
  (`install.ts:144-150`, `install.ts:301-349`).
- **Regression tests** in `create-openclaw-octo/cli/install-migration.test.ts:1067-1171`,
  including the FS / openclaw mocks (`setupFs` / `applyFsMocks` at lines 55-115)
  to confirm the stale-cfg test actually exercises the cfg-fallback branch
  rather than passing accidentally.
- **Security review** (the PR was auto-classified `security_sensitive`; see Section 5).

---

## 1. Verification of stated behavior

| Claim | Status | Evidence |
|---|---|---|
| `detectScenario()` now returns `octo-npm-legacy` before the deadlock branch on Path B | ✅ | `openclaw-cli.ts:1303` runs before `openclaw-cli.ts:1306` |
| New branch only fires when active-install evidence is present (not just stale cfg) | ✅ | `isNpmLegacyActivelyInstalled()` at `openclaw-cli.ts:1394-1407`: inspect-healthy → `plugins list --json` enabled → cfg.entries **AND** `~/.openclaw/npm/node_modules/openclaw-channel-octo` on disk |
| `octo-npm-legacy` install path runs a `fresh`-style install, no deadlock banner | ✅ | `install.ts:251-266` |
| Legacy npm plugin is still uninstalled after the new install verifies healthy | ✅ | Pre-existing `oldNpmSnapshot` machinery at `install.ts:144-150` + `install.ts:301-349` runs unconditionally, including channels.octo + bindings save/restore around `pluginsUninstall` |
| `update` / `broken` still take precedence when octo is already (partially) installed | ✅ | `openclaw-cli.ts:1284-1285` returns before the new branch |
| Stale `cfg.entries.openclaw-channel-octo` without an active install routes to deadlock, not octo-npm-legacy | ✅ | Regression test at `install-migration.test.ts:1137-1170` |

The test infrastructure makes the second test genuinely meaningful:
`setupFs` (`install-migration.test.ts:55-67`) only marks `/extensions/<id>$`
paths as existing (line 94-98), so `hasLegacyNpmResidue()`'s
`existsSync('.../npm/node_modules/openclaw-channel-octo')` correctly resolves
to `false` and exercises the "stale cfg without disk" rejection path —
not a false positive from over-mocked `existsSync`.

## 2. Findings

### P2 — nit — Two near-identical "detected legacy" log lines

In the Path B happy case the user sees, in order:

1. `runInstall()` top-of-function banner (`install.ts:146-149`):
   `Detected legacy npm-installed openclaw-channel-octo v<x>. Will replace with ClawHub octo after install verifies healthy.`
2. The new `octo-npm-legacy` case banner (`install.ts:260-262`):
   `Detected legacy npm install of openclaw-channel-octo. Installing ClawHub octo...`

Both fire on the same condition. Not a bug — the second line marks the
actual install kick-off — but reads slightly repetitive. Consider either
dropping the second banner (the first already names the intent) or making
the first banner conditional on `scenario !== "octo-npm-legacy"` so only
one announcement is printed. Non-blocking.

### P2 — maintainability — Docstring claim vs implementation mismatch in `isNpmLegacyActivelyInstalled`

`openclaw-cli.ts:1391-1393` says:
> Mirrors the matrix used in `detectInstallState()` so both callers agree on what counts as an active Path B install.

The matrices don't quite mirror:

- `isNpmLegacyActivelyInstalled` (`openclaw-cli.ts:1401`) requires
  `p.id === NPM_PACKAGE_NAME && p.enabled` from `plugins list --json`.
- `detectInstallState` (`openclaw-cli.ts:1503`) accepts any entry where
  `p.id === NPM_PACKAGE_NAME` — enabled or disabled.

In practice this only matters when the legacy plugin is installed-but-disabled
on a modern OpenClaw (one that supports `plugins list --json`) AND the
inspect short-circuit at line 1398 returns false. Routing a disabled
plugin to `deadlock` instead of `octo-npm-legacy` is arguably the
right call (setup() won't run, so the legacy plugin isn't actively
holding `channels.octo`), but if so, please update the docstring or
align the two checks so a future reader doesn't get tripped by the
"mirrors the matrix" wording. Non-blocking.

### P2 — consistency — `PluginListResult.error !== null` is silently treated as "old runtime"

`PluginListResult` (`openclaw-cli.ts:587-607`) documents the contract:

> `error` non-null when supported=false because of a real error (not "subcommand unknown");
> callers should fail-closed rather than trust cfg.entries.

`isNpmLegacyActivelyInstalled` (`openclaw-cli.ts:1399-1406`) treats both
flavours of `supported=false` the same way — falling through to the
`cfg.entries + hasLegacyNpmResidue()` fallback even when `error` is set.

The blast radius is small because the fallback still requires the npm
install directory to be physically present on disk, so a corrupt-cfg-only
host won't be misclassified. Still, for hosts where the runtime is in a
genuinely unhappy state (config corruption, plugin load crash) the
documented "fail-closed" intent isn't honored here. Worth a small follow-up
to either branch on `list.error !== null` or relax the contract on the
`PluginListResult` docstring. Non-blocking.

## 3. Suggestions

- Pick one of the two log lines in §2.1 (or gate the first on scenario) to
  reduce duplication on the Path B happy path.
- Either update the docstring on `isNpmLegacyActivelyInstalled` to call
  out the "requires `enabled === true`" delta vs. `detectInstallState`,
  or align the two by adding `&& p.enabled` at `openclaw-cli.ts:1503`
  (whichever the maintainer thinks better reflects intent).
- Optionally add a third regression test asserting that an
  installed-but-disabled npm legacy plugin (enabled=false in
  `plugins list --json`) does NOT route to `octo-npm-legacy` — locks in
  whichever semantics you choose for §2.2.

## 4. Additional observations (not part of the diff but worth knowing)

- The `octo-npm-legacy` branch is reachable only when neither `isHealthy`
  nor `hasNewPartial` is true (`openclaw-cli.ts:1284-1285`), so the
  combined-state "legacy npm + healthy ClawHub octo" continues to be
  handled by the `update` branch + the unconditional `oldNpmSnapshot`
  cleanup — the new branch does not introduce a regression for that
  combo.
- The `oldNpmSnapshot` cleanup at `install.ts:301-349` already preserves
  `channels.octo` + bindings across the legacy uninstall (the side
  effect that `openclaw plugins uninstall` has on cfg), so the new
  `fresh`-style path doesn't need its own save/restore. Good reuse.
- `cleanNpmPackageJsonResidue` at `install.ts:321` continues to be
  called as part of cleanup, so #94's manifest-residue fix is still
  exercised on this path.

## 5. Security review (PR was auto-tagged `security_sensitive`)

The classification fired on `install` / `migration` keywords. After
reading the diff:

- No new user-controlled input is introduced.
- The new helper reads cfg (already trusted), runs `openclaw plugins
  list --json` (already used elsewhere), and checks `existsSync` on a
  path derived from `getConfigFilePathSafe()` (already trusted).
- `pluginsInstall(spec, ...)` is unchanged — `spec` still comes from
  `CLAWHUB_INSTALL_SPEC` or `opts.from`, same as the existing flow.
- No credential, token, or middleware change.

Reads as a false positive on the auto-classifier; no new security
surface from this PR. The `needs-human-review` label still gates the
merge as designed, so a human will see this section.

---

Jerry-Xin (approved):
## ✅ APPROVE (re-review)

Previous round blocking issue is fixed:

### Classifier over-breadth ✅

`isNpmLegacyActivelyInstalled()` replaces the bare `cfg.plugins.entries` check with 3-layer active-evidence:
1. `plugins inspect openclaw-channel-octo` → installed + enabled
2. `plugins list --json` fallback → package in list
3. cfg entries + npm-layout directory existence → filesystem evidence

Stale config residue (entries present but no npm install directory) + `channels.octo` now correctly routes to `deadlock` repair, not `octo-npm-legacy`. Regression test added.

### Rebase clean

31eddea881da is a rebase onto latest main (absorbing #97/#98/#99 merges). The `octo-npm-legacy` classification logic is unchanged from the previously-approved 0477164c2c38.

CI green, tsc clean. No remaining blocking issues.

Jerry-Xin (approved):
Scope check passed: this PR modifies the `create-openclaw-octo` installer/migration flow, which is directly part of this repository.

One-line summary: The change correctly prioritizes active legacy npm Path B detection before deadlock handling, with useful regression coverage and no blocking issues found.

💬 Non-blocking
- 🟡 Warning: I could not run the focused tests or type-check locally because dev dependencies are not installed in this checkout: `vitest: command not found` and `tsc: command not found`.
- 🔵 Suggestion: `create-openclaw-octo/cli/install.ts:1` still documents the scenario order without the new `octo-npm-legacy` case. Updating that header would keep the operational docs aligned with the switch.
- 🔵 Suggestion: `create-openclaw-octo/cli/openclaw-cli.ts:1391` says the new helper mirrors `detectInstallState()`, but the list-based check filters `enabled` at `openclaw-cli.ts:1401` while `detectInstallState()` checks only the id at `openclaw-cli.ts:1502`. That may be intentional, but a short comment explaining the distinction would reduce future regression risk.

✅ Highlights
- 🔵 Good: `detectScenario()` now checks active `openclaw-channel-octo` evidence before `channels.octo` deadlock classification at `openclaw-cli.ts:1303`, which addresses the misleading Path B log.
- 🔵 Good: The active-install guard avoids treating stale config residue as npm legacy, preserving legitimate deadlock repair behavior at `openclaw-cli.ts:1398`.
- 🔵 Good: The regression tests cover both the intended Path B route and the stale-config deadlock route in `install-migration.test.ts:1072` and `install-migration.test.ts:1137`.

PR #102, opened 2026-05-29
fix(ci): grant actions:read permission to Octo CI Status workflow

lml2468:
## Problem

`Octo CI Status` workflow has been failing with `startup_failure` on **every run** since at least May 21st.

## Root Cause

The caller workflow sets `permissions: {}` (empty), but the reusable workflow (`Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml`) needs `actions: read` at the job level to query the GitHub API for workflow run history (state-change detection).

GitHub Actions enforces that reusable workflow job permissions cannot exceed caller permissions. Since the caller grants nothing, the job's `actions: read` request fails at startup.

## Fix

Change `permissions: {}` to `permissions: actions: read`, matching the working configuration in octo-server, octo-web, octo-matter, octo-lib, octo-smart-summary, octo-speech, and octo-daemon-cli.

## Verification

All 7 other repos with `permissions: actions: read` have 100% success rate on this workflow.

### DIFF
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -6,7 +6,8 @@ on:
     types: [completed]
     branches: [main]
 
-permissions: {}
+permissions:
+  actions: read
 
 jobs:
   notify:


lml2468 (commented):
[APPROVE]

Minimal, correct fix — `actions: read` is the exact permission the reusable `octo-ci-status.yml` needs to query workflow run history for state-change detection.

✅ Highlights

- Read-only permission; no write access granted.
- Matches the working configuration across 7 other repos in the org.
- `actionlint` and Workflow Sanity checks both pass — the YAML is well-formed.

Note: `check-sprint` and `welcome` failures are unrelated administrative workflows, not code CI.

yujiawei (approved):
# Code Review — PR #102 (octo-adapters)

## Summary

Minimal, correct one-line fix for a real CI breakage. The root cause analysis in the PR body is accurate and the chosen remediation matches the established pattern across the sibling repos. **Approving.**

## Verification

| Claim | Verified | Evidence |
|---|---|---|
| Reusable workflow needs `actions: read` at job level | ✅ | `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@main` declares `jobs.notify.permissions: { actions: read }` |
| Reusable workflow itself has `permissions: {}` at top level | ✅ | Same file, line above `jobs:` |
| GitHub enforces caller ≥ callee job permissions | ✅ | Documented behavior; caller's `permissions: {}` cannot satisfy `actions: read` request → `startup_failure` |
| 7 peer repos use the same pattern | ✅ | Confirmed via `gh api` against octo-server, octo-web, octo-matter, octo-lib, octo-smart-summary, octo-speech, octo-daemon-cli — all carry `permissions: actions: read` |
| Diff is minimal (1 file, +2/-1) | ✅ | Only `.github/workflows/octo-ci-status.yml` changed |

## Findings

No P0 / P1 issues.

### Nit (non-blocking)

- Scope could technically be tightened to job-level (`jobs.notify.permissions: { actions: read }`) instead of workflow-level, matching how the reusable workflow itself scopes the grant. Functionally identical here because the workflow has only the single `notify` job that immediately delegates to the reusable, so this is purely stylistic — not worth a follow-up.

## Additional Observations (out of scope)

- The reusable workflow uses an unpinned `@main` ref. If the org wants reproducibility / supply-chain hardening, pinning to a SHA is the standard play. Not introduced by this PR, just noting it surfaced while reviewing.

## Verdict

`actions: read` is the minimum read-only scope required, does not expose any write capabilities, and is the exact configuration already in production on the 7 peer repos. Ship it.

Jerry-Xin (approved):
This PR is relevant to `Mininglamp-OSS/octo-adapters` because it fixes a repository-owned CI notification workflow.

✅ Highlights

🔵 Suggestion / Positive: `.github/workflows/octo-ci-status.yml:9` now grants the caller workflow the minimal `actions: read` permission needed by the reusable workflow, instead of broadening permissions unnecessarily.

🔵 Suggestion / Positive: `.github/workflows/octo-ci-status.yml:13` continues to delegate behavior to the shared `Mininglamp-OSS/.github` reusable workflow, so the fix stays aligned with the repository’s existing CI architecture.

No blocking issues found. Static review only; no runtime workflow execution was available locally.

PR #114, opened 2026-06-04
ci: set fail-on-vuln false on scheduled OSV scan

lml2468:
## What

Set `fail-on-vuln: false` on the `scan-scheduled` job in `osv-scanner.yml`.

## Why

The full scan (push/schedule/dispatch) flags pre-existing vulnerabilities
and fails the workflow. This creates permanent red CI on main — every push
and every weekly scheduled run fails on issues unrelated to the change.

With this fix:
- **scan-scheduled**: reports to Code Scanning tab but does NOT fail the workflow.
  Pre-existing vulns are tracked and remediated via the Security tab.
- **scan-pr**: unchanged (default `fail-on-vuln: true`). New vulns introduced
  by a PR are still blocked.

This matches the hermes-agent approach and the review feedback on octo-lib#61.

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -6,7 +6,7 @@ on:
   push:
     branches: [main]
   schedule:
-    - cron: "30 4 * * 1" # Weekly Monday 04:30 UTC (12:30 CST)
+    - cron: "30 4 * * 1"  # Weekly Monday 04:30 UTC (12:30 CST)
   workflow_dispatch:
 
 permissions:
@@ -17,20 +17,23 @@ permissions:
 jobs:
   scan-scheduled:
     if: github.event_name == 'push' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
-    uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@9a498708959aeaef5ef730655706c5a1df1edbc2 # v2.3.8
+    uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@9a498708959aeaef5ef730655706c5a1df1edbc2  # v2.3.8
     with:
       scan-args: |-
         -r
         ./
+      # Full scan reports pre-existing vulns to Code Scanning tab but does
+      # not fail the workflow. Pre-existing vulnerabilities should be tracked
+      # and remediated via the Security tab, not by blocking every push to main.
+      # New vulnerabilities are still blocked at PR time (scan-pr defaults to
+      # fail-on-vuln: true).
+      fail-on-vuln: false
 
   scan-pr:
     if: github.event_name == 'pull_request'
-    uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable-pr.yml@9a498708959aeaef5ef730655706c5a1df1edbc2 # v2.3.8
+    uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable-pr.yml@9a498708959aeaef5ef730655706c5a1df1edbc2  # v2.3.8
     with:
       scan-args: |-
         -r
         ./
-      # Fork PRs have read-only GITHUB_TOKEN; skip SARIF upload to avoid
-      # security-events:write permission failure. Scan still runs and
-      # results are visible in the workflow output.
-      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}


Jerry-Xin (approved):
The PR is in scope: it adjusts repository CI security scanning behavior for `Mininglamp-OSS/octo-adapters`.

💬 Non-blocking

- 🔵 Suggestion: [.github/workflows/osv-scanner.yml:39](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-918icfbi/repo/.github/workflows/osv-scanner.yml:39) also changes `upload-sarif` behavior for Dependabot PRs, but the PR description focuses on `fail-on-vuln: false` for full scans. This looks reasonable, but the description should mention it so reviewers understand the full behavior change.

✅ Highlights

- [.github/workflows/osv-scanner.yml:30](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-918icfbi/repo/.github/workflows/osv-scanner.yml:30) correctly uses the pinned reusable OSV workflow’s boolean `fail-on-vuln` input.
- PR scanning remains fail-closed by default, preserving the intended protection against newly introduced vulnerabilities.
- The change is limited to CI workflow configuration and aligns with the stated goal of avoiding permanent red CI on pre-existing findings.

No blocking issues found. I also checked the pinned upstream reusable workflow contract; the added inputs are supported.

yujiawei (approved):
# Code Review — PR #114 (octo-adapters)

## Summary

This PR sets `fail-on-vuln: false` on the `scan-scheduled` job of `osv-scanner.yml` so that the full (push / schedule / dispatch) OSV scan reports pre-existing vulnerabilities to the Code Scanning tab without turning main's CI permanently red. The intent is sound and follows an established pattern (osv-scanner-action's `fail-on-vuln` input defaults to `true`, so an explicit `false` is the correct lever). New vulnerabilities introduced by a PR remain blocked because `scan-pr` keeps the default `fail-on-vuln: true`.

**Verdict: APPROVED.** No P0/P1 blockers. A few non-blocking observations below — please consider folding the description fix in, but none of these need to gate the merge.

## Verification

- ✅ `fail-on-vuln` is a valid input on `osv-scanner-reusable.yml` and defaults to `true`; setting it `false` only on `scan-scheduled` is the right scope. Pinned action SHA `9a4987…` (v2.3.8) is unchanged.
- ✅ `scan-scheduled` does not override `upload-sarif`, so it still defaults to `true` — pre-existing vulns are still uploaded to Code Scanning. The "tracked via Security tab" claim in the description holds.
- ✅ `scan-pr` still inherits `fail-on-vuln: true` (no override), so PR-time blocking of newly introduced vulns is preserved.
- ✅ Pinned action digest, `permissions` block, and trigger matrix are untouched. `actionlint` and the "No tabs in workflow files" sanity checks pass.

## Observations (non-blocking)

### 1. Description understates the change — `scan-pr` was NOT left unchanged (maintainability)

The PR description states *"scan-pr: unchanged"*, but the diff modifies `scan-pr`'s `upload-sarif` condition:

```yaml
- upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+ upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}
```

The change itself is **correct and worth keeping**: Dependabot `pull_request` events run with a read-only `GITHUB_TOKEN`, so a SARIF upload (which needs `security-events: write`) would fail. Excluding `dependabot[bot]` avoids that failure while the scan itself still runs (and still blocks on new vulns via the default `fail-on-vuln: true`). The only consequence is that Dependabot PRs no longer surface SARIF in the Code Scanning tab — an acceptable trade-off given the token constraint.

The issue is purely that this fix is bundled silently under a PR titled/described as a `fail-on-vuln` change. Suggest updating the PR description to mention the Dependabot SARIF-skip so the diff and the narrative match (helps future `git blame` / archaeology).

### 2. Removed explanatory comment on `scan-pr` (minor)

The same hunk deletes the comment that explained *why* the `upload-sarif` gate exists (fork PRs → read-only token → skip SARIF). The replacement condition is longer and now also encodes the Dependabot case, so a one-line comment covering both fork PRs and Dependabot would preserve the rationale rather than dropping it entirely.

### 3. Cosmetic whitespace churn (nit)

Several lines gain a second space before inline `#` comments (the `cron` line and both `uses:` lines). Harmless, but it's unrelated churn in a change otherwise focused on `fail-on-vuln`. No action required.

## Security note

Classified `security_sensitive`. The one item a human should consciously sign off on: after this lands, the **only blocking** OSV gate is the PR-diff scan (`scan-pr`, which compares old vs. new and fails on *newly introduced* vulns). The full-tree scheduled scan no longer fails the run — it becomes report-only. That is the documented and intended trade-off here, and pre-existing vulns remain visible in the Code Scanning / Security tab, but it does mean a vulnerability that lands outside the normal PR flow (e.g. a direct push to main, or a transitive vuln that only the full scan catches) will be reported but not block. This is a reasonable posture for reducing permanent-red CI; flagging it so it's an explicit decision rather than an implicit one.

Jerry-Xin (approved):
LGTM — clean CI-only change.

- `fail-on-vuln: false` on `scan-scheduled` correctly decouples pre-existing vuln noise from push/schedule CI status while keeping PR-time blocking intact.
- dependabot exclusion on `upload-sarif` closes the timing window (matches the pattern from octo-ios #22 and the broader fork-PR SARIF guard rollout).
- Whitespace normalization is harmless.

Consistent with the approach already landed across octo-lib, hermes-agent, and other repos. Ship it.

PR #115, opened 2026-06-04
ci: set fail-on-vuln false on scheduled OSV scan

lml2468:
## What

Add `fail-on-vuln: false` to the `scan-scheduled` OSV job.

## Why

Without this, every push to main and every weekly scheduled run fails
on pre-existing vulnerabilities — creating permanent red CI unrelated
to the change being pushed.

With `fail-on-vuln: false`, results are still uploaded to the Security
tab (SARIF) for tracking, but don't block the workflow. PR scans still
block new vulnerabilities (default `fail-on-vuln: true`).

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -22,6 +22,7 @@ jobs:
       scan-args: |-
         -r
         ./
+      fail-on-vuln: false
 
   scan-pr:
     if: github.event_name == 'pull_request'


Jerry-Xin (approved):
This PR is relevant to `Mininglamp-OSS/octo-adapters`: it updates the repository’s OSV security scanning workflow, which is part of project CI.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
🔵 Suggestion-level context only: `.github/workflows/osv-scanner.yml:18-25` now disables vulnerability-found failure only for the `scan-scheduled` job, which covers `push`, `schedule`, and `workflow_dispatch`. The PR workflow remains separate at `.github/workflows/osv-scanner.yml:27-38`, so this change does not weaken the PR-specific scan configuration.

No tests are needed for this YAML-only CI behavior change beyond workflow validation.

yujiawei (approved):
# Code Review — PR #115 (octo-adapters)

**Verdict: APPROVED.** This is a one-line, well-scoped CI change that adopts a defensible pattern: a blocking PR scan plus a non-blocking informational scheduled/push scan. I verified the diff and the surrounding workflow against the pinned reusable-workflow SHA. No P0/P1 issues.

## What the change does

```diff
   scan-scheduled:
     uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@9a49870… # v2.3.8
     with:
       scan-args: |-
         -r
         ./
+      fail-on-vuln: false
```

- `fail-on-vuln` is a real declared boolean input on `osv-scanner-reusable.yml` (default `true`), wired to `--fail-on-vuln=${{ inputs.fail-on-vuln }}`. No typo, no schema error — confirmed against the pinned SHA.
- `fail-on-vuln: false` only changes the **final job exit code**. SARIF is still uploaded to the Security tab on push/schedule runs, so findings remain tracked and queryable.
- The `scan-pr` job is untouched and keeps `fail-on-vuln` at its default `true`. **PRs still block on newly introduced vulnerabilities** — the gate that actually matters for preventing regressions from entering `main` is intact.

## Assessment of the approach

The stated problem is legitimate: pre-existing transitive vulnerabilities with no available fix turn every push-to-`main` and every weekly scheduled run permanently red, regardless of whether the pushed change introduced anything. A permanently-red signal is worse than no signal — it trains maintainers to ignore the workflow, which then masks a *real* PR-scan failure when one eventually lands.

The chosen split — **blocking on PRs, informational on push/schedule** — is a common and reasonable CI pattern. The PR scan is where blocking has leverage (it stops bad code before merge); the scheduled scan's value is detection/visibility of drift in already-merged dependencies, which SARIF upload preserves. So the change keeps the useful property (tracking) and drops the harmful one (a stuck-red gate). I'd approve this as a pragmatic stopgap.

## Considerations worth noting (non-blocking)

1. **Visibility risk: nobody watches a scan that never goes red.** (severity: P2 / maintainability)
   Once push/schedule runs are green-by-construction, the only place a new regression surfaces is the **Security tab code-scanning alerts**, not the Actions tab. This works *only if* someone actually watches that tab. GitHub does generate alert notifications for new code-scanning results, but those are easy to tune out. Recommend (follow-up, not required for this PR): confirm code-scanning alert notifications are routed to a channel a human reads, and/or add a lightweight periodic check (even a manual cadence) so "scan is green" doesn't silently become "scan is ignored." This is the main thing I'd want a human to sanity-check given the `security_sensitive` classification.

2. **Alternatives considered — and why this is a fine first move.** (severity: nit / informational)
   - *Allowlist/ignore specific CVEs* (e.g. an `osv-scanner.toml` with per-CVE `[[IgnoredVulns]]` entries): this is the more surgical long-term answer — it keeps push/schedule **blocking** while suppressing only the known, unfixable, triaged vulns, so a *new* vuln would still turn the scheduled run red. The downside is it requires per-CVE triage and ongoing maintenance, and entries can rot. The blanket `fail-on-vuln: false` is a coarser instrument: it suppresses *all* current and *future* gating on push/schedule, not just the known set.
   - *Fixing the vulns*: ideal but evidently blocked (transitive / no upstream fix), hence not actionable now.
   - *Separating informational from gating*: that is effectively what this PR does, just at the `fail-on-vuln` flag level rather than as separate jobs.

   Net: blanket disable is acceptable today. If/when the team wants the scheduled scan to catch *new* regressions automatically, migrating to an allowlist config is the natural next step.

3. **Self-documenting commit / no inline comment.** (severity: nit)
   The PR rationale (pre-existing unfixable vulns) lives in the PR description, not in the workflow. A one-line YAML comment next to `fail-on-vuln: false` (e.g. "informational only; PR scan still blocks; see #115") would help a future maintainer who finds this and wonders whether the relaxed gate was intentional. Optional.

## Security sign-off

No new permissions, no secret handling, no token scope change, no third-party action added (the pinned reusable workflow SHA is unchanged). The blocking pre-merge guard (`scan-pr`, `fail-on-vuln` default true) is preserved. The only delta is the exit-code behavior of the post-merge informational scan. Safe to merge; the one human-verifiable item is consideration #1 (someone owns the Security-tab alerts).

lml2468 (commented):
**Security review: LGTM** ✅

_(Same GitHub account, comment-based sign-off.)_

1-line change: `fail-on-vuln: false` on scan-scheduled only. Correct — prevents main branch permanent red from pre-existing advisories in transitive deps. PR scan keeps default `fail-on-vuln: true`, so new vulns still block.

— 王大锤 🔨

yujiawei (approved):
# Code Review — PR #115 (octo-adapters)

**Verdict: APPROVED** — pragmatic, correctly-scoped CI change. No blocking issues.

## Summary

This PR adds a single line — `fail-on-vuln: false` — to the `scan-scheduled` job in `.github/workflows/osv-scanner.yml`. The push / schedule / `workflow_dispatch` OSV scan becomes informational-only instead of gating, while the pull-request scan stays blocking.

## Verification

- ✅ **Diff is exactly one line.** Only `.github/workflows/osv-scanner.yml` changes (+1/-0); no permissions, secrets, or third-party actions touched.
- ✅ **`fail-on-vuln` is a valid input.** Confirmed against the pinned reusable workflow SHA (`osv-scanner-reusable.yml@9a49870…` / v2.3.8): `fail-on-vuln` is a declared boolean input (default `true`) wired to `--fail-on-vuln=${{ inputs.fail-on-vuln }}`. No typo, no schema error.
- ✅ **YAML is correct.** `fail-on-vuln:` sits at the same indentation as `scan-args:` under `with:`, dedented below the `|-` block scalar, so it parses as a distinct input (native boolean `false`, not a string) rather than being swallowed into `scan-args`.
- ✅ **PR gate is intact.** `scan-pr` does not set `fail-on-vuln`, so it keeps the default `true` — newly introduced vulnerabilities still block merge on the PR path.
- ✅ **Tracking is preserved.** SARIF is still uploaded to the Security tab on push/schedule regardless of `fail-on-vuln`; only the job exit code changes.
- ✅ **Supply-chain hygiene maintained.** The reusable workflow remains pinned to a full commit SHA, not a mutable tag.

The rationale (pre-existing/unfixable transitive vulns producing permanently-red push and weekly runs) is legitimate. A stuck-red gate is arguably worse than an informational one, because it trains maintainers to ignore the workflow.

## Non-blocking suggestions

These are follow-ups, not merge blockers:

1. **Confirm branch protection on `main`.** With the gate off for `push`, a vulnerability introduced via a *direct push* to `main` (bypassing the PR flow) would now produce green CI and only a passive SARIF entry. This is moot if `main` requires PRs before merging (admins included). Worth a quick confirmation given the security-sensitive classification; otherwise this path's posture is weakened.
2. **Make sure Security-tab alerts reach a human.** The scheduled scan's real value is catching CVEs disclosed *after* merge — something the PR gate can never do. With `fail-on-vuln: false` that signal moves from a red build to a code-scanning alert. Ensure those alerts are routed somewhere people actually watch, so "scan is green" doesn't silently become "scan is ignored."
3. **Consider an allowlist as the longer-term fix.** A per-CVE `osv-scanner.toml` ignore list for the known pre-existing findings would keep push/schedule runs green on the known set while still failing on *genuinely new* vulns — a stronger posture than fully disabling the gate. The current blanket approach is simpler and defensible as a stopgap.

## For the human reviewer (security-sensitive)

The one item most worth a manual check is **suggestion #1** — whether `main` actually forbids direct pushes. If it does, the residual exposure from this change is negligible.

PR #116, opened 2026-06-05
ci: add history check (reject orphan-branch PRs)

lml2468:
## Summary

Wire up the org-wide reusable history check (Mininglamp-OSS/.github#62) to reject PRs with no common ancestor on the target branch.

## Why

GitHub's merge UI does not refuse merges of unrelated histories. Orphan-branch PRs merge cleanly but collapse `git blame` for every file onto a parent-less root commit. Precedent: hermes-agent PR #25045 (May 2026) re-rooted blame on ~1500 files.

## What

Add `.github/workflows/history-check.yml` that calls the reusable workflow at `Mininglamp-OSS/.github/.github/workflows/reusable-history-check.yml@main`.

Caller grants `permissions: contents: read` (required for the reusable workflow's checkout step; `{}` causes startup_failure — validated in pilot).

## Cost & verification

Per-PR: ~5-15s. Pilot Mininglamp-OSS/octo-lib#64 ran in 9s, status `success`.

Part of **T14 — reusable-history-check consumer rollout**.

cc @lml2468

### DIFF
--- a/.github/workflows/history-check.yml
+++ b/.github/workflows/history-check.yml
@@ -1,0 +1,26 @@
+# History check — reject PRs with no common ancestor on the target branch.
+#
+# Calls the org-wide reusable workflow:
+#   Mininglamp-OSS/.github/.github/workflows/reusable-history-check.yml
+#
+# Why: GitHub's merge UI does NOT refuse merges of unrelated histories.
+# An orphan-branch PR (e.g. `git checkout --orphan` or a re-initialized
+# `.git/`) merges cleanly but grafts a parent-less root commit into the
+# target branch, collapsing `git blame` for every file in that snapshot
+# onto a single commit. Precedent: hermes-agent PR #25045 (May 2026).
+
+name: History Check
+
+on:
+  pull_request:
+    branches: [main]
+
+# Caller permissions act as an upper bound for the reusable workflow.
+# The reusable workflow needs `contents: read` to run `actions/checkout`,
+# so we must grant it here too (a top-level `{}` would block startup).
+permissions:
+  contents: read
+
+jobs:
+  history:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-history-check.yml@main


Jerry-Xin (approved):
LGTM — straightforward reusable-workflow consumer.

- Trigger scoped to `pull_request → [main]` ✅
- `permissions: contents: read` is the correct minimal grant for `actions/checkout` in the reusable workflow; `{}` would block startup as noted in the comment ✅
- `@main` ref for an org-internal reusable workflow is acceptable (same trust boundary)
- Comments clearly document the rationale and precedent (hermes-agent #25045)
- CI confirms the new `history / Common ancestor check` job itself passes in 6s — consistent with the 5-15s estimate from the pilot (octo-lib#64)

No blockers. Ship it.

yujiawei (approved):
# Code Review — PR #116 (octo-adapters)

**Verdict: APPROVED** — clean, minimal, well-documented. No blocking issues.

## Summary

Adds a single 26-line caller workflow (`.github/workflows/history-check.yml`) that invokes the org-wide reusable `reusable-history-check.yml@main` to reject PRs whose history has no common ancestor with `main` (orphan branches). The motivation (preventing a `git blame` re-root, per the hermes-agent #25045 precedent) is sound.

## Verification

- ✅ **Reusable workflow path resolves.** `Mininglamp-OSS/.github/.github/workflows/reusable-history-check.yml` exists and declares `on: workflow_call`, so the `uses:` reference is valid.
- ✅ **Permissions claim is correct.** The reusable workflow's job requests `contents: read`. A called workflow can only *downgrade* the caller's `GITHUB_TOKEN` scope, never escalate it — so a top-level `permissions: {}` in the caller would prevent the job from acquiring `contents: read` and fail at startup. Granting `contents: read` at the caller's top level is the right minimal scope.
- ✅ **Trigger scoping.** `pull_request: branches: [main]` matches the default branch and the intended gate.
- ✅ **No untrusted-code execution.** The reusable workflow checks out `head.sha` but only runs `git merge-base` — it never executes PR code, and `pull_request` (not `pull_request_target`) keeps the fork token read-only. No injection surface.
- ✅ **Reusable workflow internals are sound** — pins `actions/checkout` to a SHA, uses `fetch-depth: 0`, `persist-credentials: false`, checks out `head.sha` (not the merge ref, which would mask orphan branches), and handles both the non-zero-exit and empty-output signals from `merge-base`.

## Suggestions (non-blocking)

1. **Pin the reusable workflow to a commit SHA** instead of `@main`. Referencing `@main` means any change to the org `.github` repo immediately affects every consumer, including this one. For an org-internal repo the risk is low, but a pinned SHA (with Dependabot/Renovate to bump it) gives reproducible, auditable behavior and matches how the reusable workflow itself pins `actions/checkout`. This is the only thing I'd consider tightening.
2. **Branch protection follow-up.** This check only becomes a real merge gate once `History Check` (the `check-common-ancestor` job) is added to the `main` branch protection as a required status check. That's a repo-settings change outside this diff — worth confirming as part of the T14 rollout so the workflow isn't silently advisory.

Neither point blocks merge.

PR #117, opened 2026-06-08
ci: pin .github reusable workflows to @v1

lml2468:
Migrate caller refs `@main` -> rolling `@v1` now that Mininglamp-OSS/.github has governed versioning (v1.0.0 + rolling v1). Issue notification caller repointed to `octo-issue-notify.yml@v1` (renamed from octo-issue-feed.yml). Files changed: 15.

PR #118, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
This PR is in scope for `octo-adapters`: it updates repository CI/automation for issue notification and triage.

💬 Non-blocking
- 🔵 Suggestion: [.github/workflows/octo-issue-feed.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-bmntdaqa/repo/.github/workflows/octo-issue-feed.yml:5) still triggers on `reopened`, while the called reusable workflow only forwards `opened` events. This is harmless but creates a no-op run for reopened issues.

✅ Highlights
- 🔵 The added secret forwarding at [.github/workflows/octo-issue-feed.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-bmntdaqa/repo/.github/workflows/octo-issue-feed.yml:21) matches the reusable workflow’s optional `TRIAGE_WEBHOOK_URL` contract.
- 🔵 No sensitive value is hardcoded; the webhook URL remains sourced from GitHub Secrets.
- 🔵 `permissions: {}` remains appropriate because this caller only passes event metadata and secrets to the reusable workflow.

yujiawei (approved):
# Code Review — PR #118 (octo-adapters)

## Summary

Single-line CI change: forward the org-level `TRIAGE_WEBHOOK_URL` secret from the
`octo-issue-feed.yml` caller into the reusable `octo-issue-notify.yml@v1` workflow.
This re-enables the triage webhook path, which was silently disabled because only
`OCTO_BOT_TOKEN` was being forwarded.

```diff
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}
```

## 1. Verification

- ✅ **Cross-file contract holds.** The pinned reusable workflow
  `Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1` explicitly
  declares `TRIAGE_WEBHOOK_URL` under `secrets:` (`required: false`). Passing this
  secret therefore does **not** trigger GitHub's "secret is not defined for the
  reusable workflow" hard failure. Verified against ref `@v1`.
- ✅ **Secret is actually consumed.** The reusable forwards it to the
  `octo-notify@v1` composite action as `webhook-url:
  ${{ secrets.TRIAGE_WEBHOOK_URL }}`. The wiring is end-to-end, not a dead pass.
- ✅ **Graceful degradation preserved.** The secret is declared `required: false`.
  If the org/repo secret is unset, `${{ secrets.TRIAGE_WEBHOOK_URL }}` resolves to
  an empty string and the triage webhook is simply skipped (the reusable's "Omit
  to disable" contract). No build break in the unconfigured case.
- ✅ **Secret hygiene.** The capability URL is passed via the `secrets:` context,
  never hardcoded. Consistent with the reusable's documented "never hardcode it;
  pass from a secret" guidance.
- ✅ **No data egress regression.** The webhook payload built by the reusable only
  contains identifying fields (repo, number, title, url, author, action); the
  issue body is fetched by the agent via the GitHub API, not egressed through
  Actions. This PR does not alter that.
- ✅ **Pinning intact.** The reusable remains pinned to `@v1` (from #117); this PR
  does not loosen the ref.

## 2. Findings (P0 / P1 / P2)

None. No correctness, security, or build-break issues.

## 3. Suggestions (non-blocking)

- **(nit / ops)** This change only takes effect if the `TRIAGE_WEBHOOK_URL`
  org/repo secret is configured for this repository. The code is correct either
  way, but worth a one-line confirmation that the secret exists at the
  org/Actions level so triage actually fires after merge — otherwise the path
  stays silently disabled (the very symptom this PR fixes).

## 4. Additional observations

- The reusable centrally enforces the event policy (only `opened` is forwarded),
  so the caller's `on: issues: [opened, reopened]` subscribing to `reopened` is
  harmless — `reopened` is filtered out server-side in the reusable. No action
  needed.

## Verdict

**APPROVED.** Correct, minimal, well-scoped fix; the cross-repo secret contract
is verified against the pinned `@v1` reusable. The only follow-up is operational
(ensure the secret is provisioned), not a code change.

